PR #22589, opened 2018-10-03
Add removed constructor to restore backward compatibility

fitmanyak:
Add removed constructor to `com.sun.enterprise.admin.util.AuthenticationInfo` to restore backward compatibility. This fixes https://youtrack.jetbrains.com/issue/IDEA-194181 .

### DIFF
--- a/nucleus/admin/util/src/main/java/com/sun/enterprise/admin/util/AuthenticationInfo.java
+++ b/nucleus/admin/util/src/main/java/com/sun/enterprise/admin/util/AuthenticationInfo.java
@@ -28,7 +28,14 @@ public final class AuthenticationInfo {
     private final char[] password;
 
     /**
-     * The only way to construct the instances of this class.
+     * @param user      the user name for the connection
+     * @param password  the clear text password for the connection
+     */
+    public AuthenticationInfo(String user, String password) {
+        this(user, password.toCharArray());
+    }
+
+    /**
      * @param user      the user name for the connection
      * @param password  the clear text password for the connection
      */


nucleus/admin/util/src/main/java/com/sun/enterprise/admin/util/AuthenticationInfo.java
```diff
@@ -28,7 +28,14 @@
     private final char[] password;
 
     /**
-     * The only way to construct the instances of this class.
+     * @param user      the user name for the connection
+     * @param password  the clear text password for the connection
+     */
+    public AuthenticationInfo(String user, String password) {
+        this(user, password.toCharArray());
```
pzygielo: :-1: I've seen many recommendations to avoid using `String` for password passing, none in favour of it.

smillidge: this is adding in a previously removed constructor which IntelliJ uses changing to char[] would again break IntelliJ. So the question is should we fix IntelliJ support or should IntelliJ fix the plugin?

pzygielo: [Marc reported](https://youtrack.jetbrains.com/issue/IDEA-194181#focus=streamItem-27-3084070-0-0)
> From IntelliJ IDEA 2018.2.4 build 18 Sept. 2018, did NOT encounter the problem anymore.

fitmanyak: Problem is still reproducing with any version of IntelliJ IDEA. But backward compatibility was restored in Payara (https://github.com/payara/Payara/issues/2328).

pzygielo: > Problem is still reproducing with any version of IntelliJ IDEA.

Yet [Marc' comment for IDEA-194181](https://youtrack.jetbrains.com/issue/IDEA-194181#focus=streamItem-27-3084070-0-0) remains uncorrected.

> But backward compatibility was restored in Payara ([payara/Payara#2328](https://github.com/payara/Payara/issues/2328)).

IMO - it's bad idea to have this Ctor back just for testability. Should be fixed in Intelli-J or other clients.
It is class in `com.sun` package, and some, [first example from](https://stackoverflow.com/a/8565783) @BalusC recommend:
> It's preferable to not use those classes directly in your code as it would make your code tight coupled to the implementation.

Thus - developers shall follow changes in API used.

PR #22602, opened 2018-10-25
Upgrade to Jakarta Transaction API 1.3.1

tomjenkinson:
Signed-off-by: Tom Jenkinson <tom.jenkinson@redhat.com>


smillidge (approved):

smillidge:
Glassfish doesn't build now 
` Could not resolve dependencies for project org.glassfish.main.deployment:dol:glassfish-jar:5.0.1-SNAPSHOT: Could not find artifact jakarta.transaction:jakarta.transaction-api:jar:1.3.1 in central (https://repo.maven.apache.org/maven2)`

bshannon:
You're definitely going to need to use the staging profile to build.
Can we just activate that profile by default during development?
(Of course, no one will remember to deactivate it for the final release.)

romain-grecourt:
Yes, IMO we should activate it during dev cycles *only*.

Not having it enabled by default would mean that the workspace is not build-able with just a "mvn install :(

The other way would be to use -Pstaging in the CI jobs.
This would actually be worse: it could potentially allow pushing GlassFish release artifacts to Maven Central with unresolvable dependencies that are still in staging and not released to Maven Central.

Activating/Deactivating the staging repository in the pom should be a part of the release process.
Before a release we deactivate the repository to enforce that all dependencies are available strictly from Maven Central.

That will require coordinating with all projects to ensure they release before we deactivate the staging repository.

smillidge:
Perhaps just update the readme? I can raise a PR.

romain-grecourt:
OK, let's update the readme for now.

PR #22624, opened 2018-11-26
Yasson 1.0.3, jsonb api 1.0.1

bravehorsie:
Jakarta dependencies uptake.


romain-grecourt:
@bravehorsie there is some issue with org.glassfish:json:1.1.5 not being found. See my comment in pr #22625 ; I'm unsure if this is the underlying error ; however the pipeline should have failed the same way as for #22625.

romain-grecourt (approved):

PR #22646, opened 2018-11-30
Ci pipeline updates ee4j8

romain-grecourt:
Minor cleanups for the CI pipeline.
Fix for the empty trustAnchors issue.


bshannon (approved):

romain-grecourt:
@bshannon I agree this is still a work-around. The test loads a glassfish config document with java options, one of them is `<jvm-options>-Djavax.net.ssl.trustStore=${com.sun.aas.instanceRoot}/config/cacerts.jks</jvm-options>`. I tried removing this line but the issue was still occurring. Overall this test feels wrong, it's testing upgrade from v2 to v3 config with one exception that does not fail the test. The config environment for the test is incomplete and when I fix the original exception I can see a lot of other hidden issues.

Maven 3.6.0 is not required, I ended-up installing the latest version when doing my troubleshooting.

bshannon:
Are the jvm-options in the config file loaded when the config document is loaded?
I thought they were only loaded/applied by the launcher when it launched a new
jvm to run the server.

romain-grecourt:
I think so too, which explains why it still failed when I removed that line.
Instead it could be related to the one of the config upgrade services.

bshannon:
Yes, that looks like the source of the problem.

Please file a bug against GlassFish, noting the need for the workaround.
If/when someone fixes SecuritySupportImpl, the workaround can be removed.

PR #22653, opened 2018-12-04
Add utility job to prune and re-resolve artifacts in the maven cache

romain-grecourt:
Adding a script for a job that can be used to prune and re-resolve artifacts from the maven cache used by the GF pipelines.


bshannon (approved):
Thre's a lot here I don't understand, so I'm probably not the best reviewer of this...

What calls this Groovy code?  When?

Who sets INCLUDES and RE_RESOLVE?

Why is :get used instead of :resolve to reload the artifacts?

Why not let :purge-local-repository resolve the artifacts?

Assuming I just don't understand, and there are reasonable answers to these questions, I'll approve this.

PR #22654, opened 2018-12-04
Integrate jakarta dependencies - Metro, Soteria, Interceptors-api

lukasj:
In first commit, Metro 2.4.3 and related APIs and subprojects are integrated
In second commit, Soteria 1.0.1 and related APIs are integrated + fixed forgotten updates to jakarta.annotation-api and jakarta.servlet.jsp-api
And the last commit integrates jakarta.interceptor-api

Let me know if you prefer extra PR for each update.
Thanks!

BTW: QuickLooks passed on my end for all these changes


romain-grecourt:
@lukasj fingers crossed then ;)

lukasj:
btw: I've noticed a mismatch between javax.validation-api.version in appserver/pom.xml and  javax.validation.version in nucleus/parent/pom.xml - not sure which one is correct or the comment in sources is misleading - @romain-grecourt can you check this, please?

romain-grecourt:
The property in appserver/pom.xml is only used by the javax.javaee* aggregated API artifacts. I'd say you can discard it since these artifacts should probably be re-done somewhere else.

lukasj:
@romain-grecourt do you have more tests? :-)

romain-grecourt (approved):

romain-grecourt:
@lukasj we do have more glassfish tests but they were disabled due to some earlier issues. After that we would need to run the cts tests. IMO this is good to go as-is, let me know if I should hold-off the merge.

lukasj:
I'd merge this. We may rebuild metro at some point but I believe it is already in the good enough shape and no possible changes I have in my head will affect GF

romain-grecourt:
roger that.

PR #22672, opened 2018-12-12
Add missing versions for embedded modules

vinayvishal:
If embedded profile is activated, version number to these dependencies needs to be provided.


romain-grecourt:
@vinayvishal Can you instead add these dependencies in the dependencyManagement section inside appserver/pom.xml ? Thanks.

senivam:
actually I've fixed that in PR #22673 but using dependencies management. The issue is that shoal and javax.inject dependencies changed theirs maven coordinates (and not versions), so instead of **org.shoal** the **org.glassfish.shoal** (groupId) is used and  instead of **javax.inject** the **jakarta.inject** (artifactId) is used.

PR #22684, opened 2018-12-20
HK2 groupId change.

ankathur:

PR #22685, opened 2018-12-20
Refer to project version of hk2-config

jansupol:


jansupol:
All of the modules do contain an import to hk2.config classes.
Prevents from transitively grabbing incorrect hk2-config

romain-grecourt:
this will conflict with pr #22684

romain-grecourt (approved):

PR #22686, opened 2018-12-20
Adopt Jersey 2.28-RC3

jansupol:


jansupol:
It is not possible to exclude the dependencies from whole import scope of maven. The GF bits coming from Jersey to Glassfish break the build. The cleanest way of preventing this is not to use the import scope and add the dependencies one-by-one excluding the bits causing issues in GF

jansupol:
Also breaks #22684

romain-grecourt:
@jansupol why do you need to exclude dependency management, does the jersey-bom define dependency management for GF artifacts ?

How does it build break, can you share some console output snippet ?

romain-grecourt:
Should we close the other pr (#22668) ?

jansupol:
It does not define dependency management for GF artifacts. But it transitively brings the GF dependencies. to maven upon build time. Jersey RC2 currently uses GF 5.0 bits, and it brings the GF 5.0 dependency to the build, including servlet 3.0 dependency.

To reproduce: Update version of Jersey to 2.28-RC2. Go to appserver/web/web-ha and run `mvn dependency:tree` -Pstaging

romain-grecourt:
This is very unlikely related to the import of the jersey-bom. I'll try it.

romain-grecourt (approved):
Looks good, can be merged when the pipeline passes.

PR #22688, opened 2018-12-20
move to jakarta.* in default-web.xml

lukasj:


bshannon (approved):
Who would've guessed the jar file names are here too!  :-)

Thanks for fixing!

lukasj:
@bshannon I'm only thinking whether ie jakarta.json.jar or jakarta.json.bind-api.jar (maybe few others) shouldn't be added here as well. Question for someone responsible for JSP perhaps? Do you know who to ask? In any case, I think that additions can wait to post RC2...

bshannon:
I have no idea how this list is used so I don't know what should or shouldn't be added to it.
Ask Shing Wai Chan.

lukasj:
Only limited set of jars is supposed to be used here for performance reasons, so updates are OK, additions are not

PR #22694, opened 2019-01-04
Update version to 5.1.0

jayasheelankumar:
Update version to 5.1.0


romain-grecourt:
@jayasheelankumar You'll need to update the copyright year to 2019 in the header of all these files.
@bshannon Maybe we can skip this for now and use the repair goal of the copyright plugin ?

bshannon:
If you want to use the repair goal of the copyright plugin to fix the copyright year in just the files
that are being changed here, that's fine.  Just don't use it to change *every* file in the repo.

bshannon (approved):
Everything that was changed looks correct, but I can't tell whether there are changes that were missed.

In appserver/distributions/glassfish-common/src/main/resources/README.txt, is the required minimum
JDK version still "Oracle JDK 8 Update 144"?

romain-grecourt:
The correct version for the required JDK is 8_191

jayasheelankumar:
Incorporating review comments.

PR #22706, opened 2019-01-18
Fix jakarta-annotation-api.jar location

vinayvishal:
In web profile, `jakarta-annotation-api.jar` is expected to be available
under `glassfish/modules/endorsed`, however, at present, it appears under
`glassfish/modules`. This PR fixes the location of `jakarta-annotation-api.jar`.


jigneshbhodia (approved):
LGTM

PR #22710, opened 2019-01-21
Integrate EJB 3.2.5 and javax.servlet.jsp.jstl (RI) 1.2.6

senivam:
Signed-off-by: Maxim Nesen <maxim.nesen@oracle.com>

PR #22713, opened 2019-01-22
Revert "Integrate javax.servlet.jsp.jstl (RI) 1.2.6 (#22710)"

Tomas-Kraus:
Fixed EJB API dependency.


Tomas-Kraus:
@romain-grecourt Looks like there is still some issue with nucleus tests:
No artifacts found that match the file pattern "ql_gf_nucleus_all-results.tar.gz". Configuration error?
Can you run it again?

Tomas-Kraus:
This change makes no sense now because JSP/JSTL is fixed.

PR #22714, opened 2019-01-23
add legal files to built artifacts

lukasj:
Signed-off-by: Lukas Jungmann <lukas.jungmann@oracle.com>

### DIFF
--- a/nucleus/parent/pom.xml
+++ b/nucleus/parent/pom.xml
@@ -70,6 +70,7 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <legal.doc.source>${maven.multiModuleProjectDirectory}</legal.doc.source>
         <!-- on Hudson and RE, this gets replaced by real build ID -->
         <build.id>${user.name}-private</build.id>
         <findbugs.skip>false</findbugs.skip>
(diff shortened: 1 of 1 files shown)


m0mus (approved):
LGTM

nucleus/parent/pom.xml
```diff
@@ -70,6 +70,7 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <legal.doc.source>${maven.multiModuleProjectDirectory}</legal.doc.source>
```
romain-grecourt: Are there any requirement for this property to work ?

lukasj: correct env setup... given that there exists https://issues.apache.org/jira/browse/MNG-5830.
I think using rootlocation looks better as it is not tight to from where or where the build started

lukasj: the only drawback is that it runs for every module. Or I can move it to the root pom if you prefer that

romain-grecourt: This needs to run for every module, correct ? The updated version looks fine IMO.

romain-grecourt (approved):

nucleus/parent/pom.xml
lukasj: yes, every produced source and binary jar should contain these file

PR #22722, opened 2019-01-28
Contribute Glassfish Docs

vinayvishal:
Initial contribution of Glassfish docs source

Signed-off-by: Vinay Vishal <vinay.vishal@oracle.com>


Cousjava:
Would this be better going to a gh-pages branch?

edbratt:
The adoc files are built into html, which will be pushed to the gh-pages 
branch. So, we can edit, produce private editions, then when it's all 
ready, the site can be updated. We plan to push an initial collection soon.

edbratt (approved):
LGTM

Issue #22747, opened 2019-04-11
JCA - config-property defined in descriptor is overridden with default from annotation


pzygielo:
## Environment Details
* GlassFish Version (and build number): GlassFish Server Open Source Edition  5.1.0
* JDK version: openjdk version "1.8.0_201"
* OS: GNU/Linux
* Database: none

----------

## Problem Description
[//]: # "Describe the bug in detail highlighting current behavior vs expected behavior"
_Current_: property annotated as `javax.resource.spi.ConfigProperty` with `default` provided has this default applied AFTER value from descriptor is set.
_Expected_: [If at all,] default is set first, and then value from descriptor.

IMO section `18.3.2` of [JSR-322](https://jcp.org/en/jsr/detail?id=322) (v1.7) is not preserved.

[//]: # "State if the problem is easily reproducible or happens intermittently"
Problem is easily reproducible.

[//]: # "Include stack traces or command outputs"
[Console log](https://travis-ci.org/pzrep/jca-config-property/jobs/518642574#L602-L617)

## Steps to reproduce
[//]: # "Step by step instructions to reproduce the problem"
1. Have adapter with annotated property with default set
2. Have RAR with above adapter with different config-property set in descriptor
3. After deployment property has value of default from annotation

[//]: # "Provide sample code/application if relevant"
Sample code: [jca-config-property](https://github.com/pzrep/jca-config-property/tree/a6ec89cbd0d57b5dcee18e51bd728b02d5791697)

## Impact of Issue
[//]: # "How has this issue affected you? What are you trying to accomplish?"
Default defined by Developer of Adapter in code can't be overridden by Assembler in DD of RAR (unless `metadata-complete="true"`, which defeats the idea behind defaultValue I think).

pzygielo:
Apparently, when in DD property is Capitalized, then https://github.com/eclipse-ee4j/glassfish/blob/a2b431614da72e9ecd4e3372a871353d4c51f581/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/connectors/deployment/annotation/handlers/ConfigPropertyHandler.java#L559-L569 returns true.

But - for property `foo` as:

```java
String getFoo() {...}

@ConfigProperty(defaultValue="a")
void setFoo(String v) {...}
```
value from DD (all lower case)
```xml
<config-property>
  <config-property-name>foo</config-property-name>
  <config-property-type>java.lang.String</config-property-type>
  <config-property-value>bar</config-property-value>
[...]

PR #22748, opened 2019-04-14
Remove unused constructors

pzygielo:
Signed-off-by: Piotrek Zygielo <piotr@zygielo.pl>

### DIFF
--- a/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/ConnectorConfigProperty.java
+++ b/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/ConnectorConfigProperty.java
@@ -35,17 +35,6 @@ public class ConnectorConfigProperty extends EnvironmentProperty {
     private final static Logger _logger = DOLUtils.getDefaultLogger();
 
     /**
-    ** copy constructor.
-    */
-    public ConnectorConfigProperty(ConnectorConfigProperty other) {
-        super(other);
-    }
-
-    public ConnectorConfigProperty(EnvironmentProperty other){
-        super(other);
-    }
-
-    /**
     ** Construct an connector  config-property if type String and empty string value and no description.
     */
 
@@ -68,19 +57,6 @@ public class ConnectorConfigProperty extends EnvironmentProperty {
 
     public ConnectorConfigProperty(String name, String value, String description, String type) {
     	super(name, value, description, type);
-    }
-
-    /**
-    ** Construct an connector config-property of given name value and description and type.
-    ** Throws an IllegalArgumentException if bounds checking is true and the value cannot be
-    ** reconciled with the given type.
-    */
-    public ConnectorConfigProperty(String name, String value, String description, String type,
-                                   boolean ignore, boolean supportsDynamicUpdates, boolean confidential) {
-    	super(name, value, description, type);
-        this.ignore = ignore;
-        this.supportsDynamicUpdates = supportsDynamicUpdates;
-        this.confidential = confidential;
     }
 
     public boolean isIgnore() {


smillidge:
Can you detail what testing was carried out against the PR?

pzygielo:
> Can you detail what testing was carried out against the PR?

Apart from checking GF being buildable (ie. no references to removed Ctors in project) I performed following manual steps:

1. I built `appserver/extras/embedded/all/pom.xml`
1. I ran [test project](https://github.com/pzrep/jca-config-property/tree/db70ea932b86dd5ce48b74ab208aea9dcf8b84bd) against local SNAPSHOT built above
which in my eyes proved that both - config-parameters from deployment descriptor and `@`ConfigParameters from code were still recognized.
Change [obviously] does not solve #22747 [yet], so my test was still failing, but RAR was successfully deployed.

bshannon:
I'm not clear on what problem is being solved by removing these constructors.

ConnectorConfigProperty extends EnvironmentProperty, and EnvironmentProperty
has these kinds of constructors.  Why remove them here and not there?  Why were
they added to begin with?

pzygielo:
> ConnectorConfigProperty extends EnvironmentProperty, and EnvironmentProperty
> has these kinds of constructors. 

Constructors are not inherited so having in super-class handful of them does not mean sub-class has to provide similar set nor is excused for doing so.

> Why remove them here and not there?

Good idea. Will check if possible to do so.

> I'm not clear on what problem is being solved by removing these constructors.

It's the code that never gets executed (not even in UTs), but has to be maintained, compiled, kept in jar, etc.. and discussed about like here.

I focused on ConnectorConfigProperty only, as for me it seems there is the place to fix #22747.
Without those 3 Ctors fix is a bit simpler. (Should they be missing already I'd save some time on my investigation).

> Why were they added to begin with?

Looks like they [were always with us](https://github.com/javaee/glassfish/commit/962f22930c9#diff-27e601c417bbd77a361e2a23ba5a220f) (brand new file introduced in Jul 2009).

pzygielo:
>  Why remove them here and not there?

Cannot do, as 

this

https://github.com/eclipse-ee4j/glassfish/blob/eac66d65e6a8eccedaaf40fac0e514d12c71e314/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/EnvironmentProperty.java#L77

is used (at least) here

https://github.com/eclipse-ee4j/glassfish/blob/eac66d65e6a8eccedaaf40fac0e514d12c71e314/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/MessageDestinationReferenceDescriptor.java#L61



this

https://github.com/eclipse-ee4j/glassfish/blob/eac66d65e6a8eccedaaf40fac0e514d12c71e314/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/EnvironmentProperty.java#L95

is used (at least) here

https://github.com/eclipse-ee4j/glassfish/blob/eac66d65e6a8eccedaaf40fac0e514d12c71e314/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/ResourceReferenceDescriptor.java#L105




this 

https://github.com/eclipse-ee4j/glassfish/blob/eac66d65e6a8eccedaaf40fac0e514d12c71e314/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/EnvironmentProperty.java#L105

is used (at least) here

https://github.com/eclipse-ee4j/glassfish/blob/eac66d65e6a8eccedaaf40fac0e514d12c71e314/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/ConnectorConfigProperty.java#L69

bshannon:
Since we no longer have people who understand this code intimately, I'm a bit nervous
about making changes that are not necessary.  And with other subclasses of
EnvironmentProperty also providing similar constructors, I don't really see the
motivation to remove them here.  Maybe I would change my mind if the fix for #22747 were significantly more complicated because of these constructors.

@smillidge what do you think?

bshannon:
Still waiting for comments from @smillidge before approving all your other PRs that
remove all the "unused" constructors.

And I'd like to see the fix for #22747 that justifies these changes.

pzygielo:
> And I'd like to see the fix for #22747 that justifies these changes.

[My fix](https://github.com/pzygielo/glassfish/commit/23615c7572e49ec489c6d8cc6587e0e3da7d40ec) doesn't do that. If taken - 22748 would save only one call.

These removals, IMO, are due to their own bad smell alone.

(It's clear, that it is not pull _demand_ :grin:)

smillidge:
I'm happy to remove dead code

bshannon:
@smillidge if you're sure all of this is dead code, please approve and merge these PRs.  Thanks.

bshannon:
@smillidge poke

smillidge (approved):

smillidge:
Jenkins is green so I will merge

PR #22750, opened 2019-04-17
Add UT for EnvironmentProperty.getValue

pzygielo:


bshannon (changes requested):
The file needs a copyright/license header.

It looks like you're using 2 space indentation instead of the standard 4 spaces,
probably because EjbInterceptorTest.java does.  Don't copy their mistake!  :-)
Thanks.

pzygielo:
Thanks for checking. But it's only draft for now. (Sign-off is also missing, deliberately).
I just want the test from the second commit to fail :grin:

PR #22795, opened 2019-07-27
Upgrade to Jakarta Concurrency 1.1.2

smillidge:
Signed-off-by: Steve Millidge <steve.millidge@payara.fish>


Cousjava (approved):

bshannon:
The CI job failed because org.glassfish:jakarta.enterprise.concurrent:jar:1.1.2 is not in the staging repository.  Please update the staging repository so we can try again.

smillidge:
OK I see I need to put the RI there as well with the same version number. May take some time given the Jakarta EE spec work

smillidge:
Removed the upgraded RI dependency as that isn't released yet and there's no need at this point.

PR #22843, opened 2019-09-27
Fix for ear redeploy fails(eclipse-ee4j#22383)

endtak:
Enables class loader release processing when an application is undeployed.

Signed-off-by: endtak <endo.takafumi@jp.fujitsu.com>


bshannon (commented):
This fix looks plausible to me, but someone more familiar with the code should review it.

smillidge (commented):
LGTM

smillidge (approved):

Issue #19059, opened 2012-09-05
Deprecate use of TimerWelcomeServlet


glassfishrobot:
Admin GUI should be able to show the same info as the TimerWelcomeServlet.

PR #22852, opened 2019-10-30
fix for deprecate use of TimerWelcomeServlet #19059

141801:
Signed-off-by: LIANG ZHANG <zhang-liang@jp.fujitsu.com>


smillidge (approved):

Cousjava:
The original issue this is in answer to (#19059) says to deprecate TimerWelcomeServlet and add the content of it to the admin console. Has equivalent functionality been added to the admin console already?

smillidge:
There is some functionality which shows timers in the admin console probably needs testing that it works. @141801 are you sure the replacement timers interface in the admin console works?

PR #22855, opened 2019-10-31
Minimal changes including the following changes to run with JDK 11.

hs536:
Minimal changes including the following changes to run with JDK 11.
 - Remove JVM options (java.ext.dirs and java.endorsed.dirs) from domain template and osgi.properties.
 - Move the packages from modules/endorsed/ to modules/ or lib/.
 - Add -Xbootclasspath option to load grizzly-npn-bootstrap.jar.
 - Add JVM options for reference to internal APIs and enable to add long JVM options (such as --add-opens).


smillidge:
I will approve this but it looks like we need some jenkins work to get this tested in the future.

smillidge (approved):

Cousjava (approved):

PR #22861, opened 2019-11-18
Fix for context root(eclipse-ee4j#22860)

takeda0121:
Signed-off-by: takeda0121 <takeda.rikuho@jp.fujitsu.com>  

I changed key of valueMap to which context root is preserved when we redeploy a web application by value of valueMap corresponding from "contextroot" to "contextRoot" because correct key of valueMap is "contextRoot".


bshannon (approved):
This looks good to me, but we have to figure out why the CI tests are failing.
I don't think the failure is related to this change.  I'm going to work on that
issue separately and if it's not solved in a few days I may just merge this anyway.

hs536 (approved):

PR #22865, opened 2019-12-17
Introducing few changes to improve CI pipeline. As we have observed stop-domain was causing issue, hence we have modified scripts/files to handle the case

RohitKumarJain:
Signed-off-by: RohitKumarJain <rohit.ku.jain@oracle.com>

### DIFF
--- a/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/deploy/FilterDefDecorator.java
+++ b/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/deploy/FilterDefDecorator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1997, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1997, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -90,7 +90,8 @@ public class FilterDefDecorator extends FilterDef {
     }
 
     public boolean isAsyncSupported() {
-        return decoree.isAsyncSupported();
+        Boolean decoreeAsyncFlag = decoree.isAsyncSupported();
+        return (decoreeAsyncFlag == null ? false : decoreeAsyncFlag);
     }   
 
 }


bshannon:
Let me see if I understand what was done here...

In some cases you changed the script to first try "asadmin stop-domain"
and if that fails to try "asadmin stop-domain --force=true".

In some other cases you change the script to just use "asadmin stop-domain --kill=true".

I'm not sure what the difference is between --force and --kill.
Why do you use one in some places and the other in other places?
Is there a strategy for which to use when?

This all seems like a reasonable workaround.  I only wish we could determine
why the simple "asadmin stop-domain" is failing to begin with.

appserver/tests/appserv-tests/config/common.xml
```diff
@@ -1821,14 +1821,38 @@ AS_ADMIN_MASTERPASSWORD=${master.password}
...
+           <arg line="stop-domain"/>
+        </exec>
+        <condition property="cmd1failed" value="true">
+        <not>
+           <equals arg1="0" arg2="${ret1}"/>
+        </not>
+        </condition>
+        <exec executable="${ASADMIN}" xmlns:if="ant:if" if:true="${cmd1failed}">
+        <arg value="stop-domain"/>
+        <arg line="--force=true"/>
```
gurunrao: Please add a message when stop-domain is invoked with force option.

appserver/tests/appserv-tests/devtests/batch/run_test.sh
```diff
@@ -36,7 +36,7 @@ test_run(){
   echo "ANT_OPTS=${ANT_OPTS}"
   ant ${TARGET} | tee ${TEST_RUN_LOG}
   ${S1AS_HOME}/bin/asadmin stop-database
-  ${S1AS_HOME}/bin/asadmin stop-domain
+  ${S1AS_HOME}/bin/asadmin stop-domain --kill=true
```
gurunrao: check if stop-domain fails without force or kill, then invoke the force or kill option.

PratoshJain: > Let me see if I understand what was done here...
> 
> In some cases you changed the script to first try "asadmin stop-domain"
> and if that fails to try "asadmin stop-domain --force=true".
> 
> In some other cases you change the script to just use "asadmin stop-domain --kill=true".
> 
> I'm not sure what the difference is between --force and --kill.
> Why do you use one in some places and the other in other places?
> Is there a strategy for which to use when?
> 
> This all seems like a reasonable workaround. I only wish we could determine
> why the simple "asadmin stop-domain" is failing to begin with.


@bshannon , 

Let me try to give inputs for open questions, please find my comments inline starts with 
"Rohit".
1): In some other cases you change the script to just use "asadmin stop-domain --kill=true".

Rohit: We have tried to stop domain using force that didn't work. Hence we have opted for kill option.

2): I'm not sure what the difference is between --force and --kill.

Rohit:
--force
Specifies whether the domain is forcibly stopped immediately.
--kill
Specifies whether the domain is killed by using the functionality of the operating system to terminate the domain process.

3): why the simple "asadmin stop-domain" is failing to begin with.

Rohit: We are investigating it using the server logs and will keep you posted on the same.

The latest CI job has recent changes and the results can be seen under 
https://jenkins.eclipse.org/glassfish/blue/organizations/jenkins/glassfish/detail/PR-22865/1/pipeline/

There are test failures for these stages which are not fixed by this change -in nucleus_admin_all,  ql_gf_full_profile_all, ql_gf_nucleus_all, ql_gf_web_profile_all.

RohitKumarJain:
> Let me see if I understand what was done here...
> 
> In some cases you changed the script to first try "asadmin stop-domain"
> and if that fails to try "asadmin stop-domain --force=true".
> 
> In some other cases you change the script to just use "asadmin stop-domain --kill=true".
> 
> I'm not sure what the difference is between --force and --kill.
> Why do you use one in some places and the other in other places?
> Is there a strategy for which to use when?
> 
> This all seems like a reasonable workaround. I only wish we could determine
> why the simple "asadmin stop-domain" is failing to begin with.

@bshannon 

Let me try to give inputs for open questions, please find my comments inline starts with "Rohit".

1): In some other cases you change the script to just use "asadmin stop-domain --kill=true".

Rohit: We have tried to stop domain using force that didn't work. Hence we have opted for kill option.

2): I'm not sure what the difference is between --force and --kill.

Rohit:

--force
Specifies whether the domain is forcibly stopped immediately.
--kill
Specifies whether the domain is killed by using the functionality of the operating system to terminate the domain process.

3): why the simple "asadmin stop-domain" is failing to begin with.

Rohit: We are investigating it using the server logs and will keep you posted on the same.

The latest CI job has the recent changes and results can be seen under

https://jenkins.eclipse.org/glassfish/blue/organizations/jenkins/glassfish/detail/PR-22865/1/pipeline/

There are test failures for these stages which are not fixed by this change -in nucleus_admin_all,  ql_gf_full_profile_all, ql_gf_nucleus_all, ql_gf_web_profile_all.

appserver/tests/appserv-tests/devtests/batch/run_test.sh
RohitKumarJain: @gurunrao  I'll incorporate your comments... Thanks

bshannon:
So sometimes --force works, and sometimes --force doesn't work but --kill works,
and we have no idea why either of them are required in any of the individual cases
where we've made this change, right?

If I remember correctly, stop-domain just asks the server "please stop yourself" and
then waits for some time checking periodically to see if the server has stopped, failing
if the time limit is exceeded without the server actually stopping.  Have we changed it
to wait longer to see if the server just needs more time to stop?

Most likely there's some "shutdown" operation in the server that is hung or deadlocked
and that's why the server won't stop.  Can we force a JVM thread dump when the server
fails to stop so we can debug it and see what it's doing when it's supposed to be stopping?
This would be particularly interesting when using --force, since the server should be doing
very little before stopping itself.  You should be able to use "jps" to find the process ID, and
kill -QUIT to force the thread dump.

(BTW, the purpose of "Quote reply" is so that you can edit the quoted text and intersperse
your reply with the original text.  You don't need to copy the original text a second time in
the reply so that you can reply to it.  And with the quoted text all preceded with ">" and
your text *not* preceded with ">", you don't need to mark which text is your text because
your text is all the text that's not quoted.  Just a little tip for how to use GitHub and email
and so many others.  Watch how I do it below...)

> The latest CI job has the recent changes and results can be seen under
>
> https://jenkins.eclipse.org/glassfish/blue/organizations/jenkins/glassfish/detail/PR-22865/1/pipeline/
>
> There are test failures for these stages which are not fixed by this change -in nucleus_admin_all, ql_gf_full_profile_all, ql_gf_nucleus_all, ql_gf_web_profile_all.

Looking at the link above, I can't see the log file output for anything after the
"Check out from version control" step.  Why is that?

I guess there are different tests that are failing in each one of these stages?
Are these new test failures?  Or have these been failing all along?

alwin-joseph:
@RohitKumarJain @gurunrao 

Recent PR job ran : https://jenkins.eclipse.org/glassfish/blue/organizations/jenkins/glassfish/detail/PR-22865/4/pipeline

After the latest changes and the job run, I believe below are the current issues pending.

1.  Total 5 test failures in ql_gf_nucleus_all & nucleus_admin_all. Can you compare with other PR jobs to see if this failure is present there too. 

2.  There seem to be some more stop-domain failures. Search for "Command stop-domain failed." in the below logs. I  think they are missed in the fix.
https://jenkins.eclipse.org/glassfish/blue/rest/organizations/jenkins/pipelines/glassfish/branches/PR-22865/runs/4/nodes/76/steps/391/log/?start=0
https://jenkins.eclipse.org/glassfish/blue/rest/organizations/jenkins/pipelines/glassfish/branches/PR-22865/runs/4/nodes/75/steps/347/log/?start=0
https://jenkins.eclipse.org/glassfish/blue/rest/organizations/jenkins/pipelines/glassfish/branches/PR-22865/runs/4/nodes/77/steps/396/log/?start=0

3. There is a CI issue with log "java.nio.channels.ClosedChannelException" for random stages. I wonder if this is also related to any resource limitation, but that was told to be only for ci.eclipse. Maybe we would need to edit the Jenkins pipeline for this job too(I see that cpu=3 for main glassfish-ci container alone). Please check https://bugs.eclipse.org/bugs/show_bug.cgi?id=558486 to find out about that possibility. 
Since you have already logged a bug (https://bugs.eclipse.org/bugs/show_bug.cgi?id=558878) with same issue for your test job, can you update in this bug with this master PR job to get advice from Eclipse infra.

RohitKumarJain:
Hi Everyone,

As soon as I synched my fork with master this PR got closed automatically. Apologies for inconvenience.

Issue #22612, opened 2018-11-12
ConcurrentModificationException while starting process


tsyamamoto:
[//]: # "Please remove these comments"
[//]: # "Provide a general summary of the issue in the Title above"

## Environment Details
* GlassFish Version (and build number): 4.1,5.0,5.1.0-RC1
* JDK version:8
* OS:Windows,Linux
* Database:None

----------

## Problem Description
[//]: # "Describe the bug in detail highlighting current behavior vs expected behavior"
[//]: # "State if the problem is easily reproducible or happens intermittently"
[//]: # "Include stack traces or command outputs"
hk2-config has thread unsafe problems.
Starting GlassFish sometimes hungs up because of ConcurrentModificationException.

<pre>
[2018-11-09T15:09:25.904+0900] [glassfish 5.0] [INFO] [NCLS-CORE-00017] [javax.enterprise.system.core] [tid: _ThreadID=1 _ThreadName=main] [timeMillis: 1541743765904] [levelValue: 800] [[
  GlassFish Server Open Source Edition  5.0.1  (5.1.0-RC1) startup time : Felix (15,000ms), startup services(81,825ms), total(96,825ms)]]

[2018-11-09T15:09:25.905+0900] [glassfish 5.0] [SEVERE] [NCLS-CORE-00019] [javax.enterprise.system.core] [tid: _ThreadID=1 _ThreadName=main] [timeMillis: 1541743765905] [levelValue: 1000] [[
  Shutting down server due to startup exception
java.util.ConcurrentModificationException
	at java.util.ArrayList$Itr.checkForComodification(Unknown Source)
	at java.util.ArrayList$Itr.next(Unknown Source)
	at org.jvnet.hk2.config.Dom.element(Dom.java:513)
	at org.glassfish.grizzly.config.dom.NetworkListener$Duck.findThreadPool(NetworkListener.java:205)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.lang.reflect.Method.invoke(Unknown Source)
	at org.jvnet.hk2.config.Dom.invokeDuckMethod(Dom.java:1105)
	at org.jvnet.hk2.config.Dom.invoke(Dom.java:1056)
[...]

PR #22952, opened 2020-04-02
Fix for synchronization for Dom(#22612)

tsyamamoto:
1. add a synchronization block for constructor to avoid ConccurentModificationException for source.children
2. shrink the synchronization scope of the writeTo method

Signed-off-by: tsyamamoto <tsyamamoto@fujitsu.com>

### DIFF
--- a/nucleus/hk2/hk2-config/src/main/java/org/jvnet/hk2/config/Dom.java
+++ b/nucleus/hk2/hk2-config/src/main/java/org/jvnet/hk2/config/Dom.java
@@ -328,8 +328,10 @@ public class Dom extends AbstractActiveDescriptor implements InvocationHandler, 
     public Dom(Dom source, Dom parent) {
         this(source.getHabitat(), source.document, parent, source.model);
         List<Child> newChildren = new ArrayList<Child>();
-        for (Child child : source.children) {
-            newChildren.add(child.deepCopy(this));
+        synchronized (source) {
+            for (Child child : source.children) {
+                newChildren.add(child.deepCopy(this));
+            }
         }
         setChildren(newChildren);
         attributes.putAll(source.attributes);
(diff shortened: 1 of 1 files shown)


smillidge (approved):

nucleus/hk2/hk2-config/src/main/java/org/jvnet/hk2/config/Dom.java
```diff
@@ -328,8 +328,10 @@ public ServiceLocator getHabitat() {
     public Dom(Dom source, Dom parent) {
         this(source.getHabitat(), source.document, parent, source.model);
         List<Child> newChildren = new ArrayList<Child>();
-        for (Child child : source.children) {
-            newChildren.add(child.deepCopy(this));
+        synchronized (source) {
```
pzygielo, line 331: https://rules.sonarsource.com/java/tag/multi-threading/RSPEC-2445

tsyamamoto: Synchronized blocks of non-final fields are not necessarily a problem.
Unlike the URL example, this constructor does not change "source.children".

pzygielo: > Unlike the URL example, 

It'd be extremely interesting and less fun should the example considered this `Dom.java` exactly with that change.

> this constructor does not change "source.children".

Can `source.children` be changed externally to this Ctor?
Can it happen in different thread in not-synchronized (on the very same `source` object) way?
Can it be controlled from this point how `source` is used in other threads/places?

pzygielo: > Can `source.children` be changed externally to this Ctor?

So I was wrong in this part, as access to `children` is synchronized on `source`. 

> Can it happen in different thread in not-synchronized (on the very same source object) way?

Probably it can not (now).

This change indeed prevents `ConcurrentModificationException` from happening, as I saw in my unit test.  (CMEx happens in previous `Dom` version).

Sorry for the noise!

PR #22963, opened 2020-04-21
#22925: Remove Pruned API Jakarta XML RPC

lukasj:
fixes #22925

Signed-off-by: Lukas Jungmann <lukas.jungmann@oracle.com>

Issue #22925, opened 2020-02-29
Remove Pruned API Jakarta XML RPC


smillidge:
Gaurav is taking this one

PR #22968, opened 2020-04-25
#22925 Update to Metro without RPC after removing RPC in GF

arjantijms:
Followup to #22925 and PR https://github.com/eclipse-ee4j/glassfish/pull/22963 where JAX-RPC was removed, but the Metro version used still declared its usage. Added 3.0.0-M1 metro which is not yet the Jakarta-ized version, but the version without the JAX-RPC requirement.

### DIFF
--- a/nucleus/parent/pom.xml
+++ b/nucleus/parent/pom.xml
@@ -79,7 +79,7 @@
         <findbugs.version>3.0.3</findbugs.version>
         <findbugs.glassfish.logging.validLoggerPrefixes>javax.enterprise</findbugs.glassfish.logging.validLoggerPrefixes>
 
-        <webservices.version>2.4.3</webservices.version>
+        <webservices.version>3.0.0-M1</webservices.version>
         <glassfish-corba.version>4.2.0</glassfish-corba.version>
         <stax-api.version>1.0-2</stax-api.version>
         <slf4j.version>1.7.21</slf4j.version>
@@ -107,7 +107,7 @@
         <jersey.version>2.30.1</jersey.version>
         <jackson.version>2.10.2</jackson.version>
         <jettison.version>1.4.0</jettison.version>
-        <jaxb-api.version>2.3.2</jaxb-api.version>
+        <jaxb-api.version>2.3.3</jaxb-api.version>
         <jax-rs-api.spec.version>2.1</jax-rs-api.spec.version>
         <jax-rs-api.impl.version>2.1.5</jax-rs-api.impl.version>
         <mimepull.version>1.9.12</mimepull.version>
--- a/nucleus/parent/pom.xml
+++ b/nucleus/parent/pom.xml
@@ -79,7 +79,7 @@
         <findbugs.version>3.0.3</findbugs.version>
         <findbugs.glassfish.logging.validLoggerPrefixes>javax.enterprise</findbugs.glassfish.logging.validLoggerPrefixes>
 
-        <webservices.version>3.0.0-M1</webservices.version>
+        <webservices.version>3.0.0-alpha1</webservices.version>
         <glassfish-corba.version>4.2.0</glassfish-corba.version>
         <stax-api.version>1.0-2</stax-api.version>
         <slf4j.version>1.7.21</slf4j.version>


bshannon (approved):
LGTM

nucleus/parent/pom.xml
```diff
@@ -79,7 +79,7 @@
         <findbugs.version>3.0.3</findbugs.version>
         <findbugs.glassfish.logging.validLoggerPrefixes>javax.enterprise</findbugs.glassfish.logging.validLoggerPrefixes>
 
-        <webservices.version>2.4.3</webservices.version>
+        <webservices.version>3.0.0-M1</webservices.version>
```
lukasj, line 82: just for the record - this brings in dependencies NOT yet approved by Eclipse EMO, therefore also NOT available in maven central yet. They won't get to central before the approval from EMO will be given.

lukasj:
wsit 3.0.0-M1 was supposed to contain jakartized jaxb, saaj, jaxws, be dependent on jakartized eclipselink (jaxws, thus also metro, has a dependency on eclipselink), has JAXR/JAX-RPC implementations removed and going out around next weekend (currently it is ~70-75% complete)

Since this is going to have a dependency on jakartized APIs as well - persistence, transactions, connectors, servlets, ejbs, security, activation, mail, annotations as well as webservices related ones - jaxws, jaxb, jws and saaj. it is likely that once current M1 gets replaced by the upcoming M1, things get broken more than ever. IMHO, the only correct solution to the problem was to revert offending change and let me rework it instead.

Since you staged M1 (btw giving the project at least a chance to respond would be really appreciated next time, also consider that not everyone works 24/7/366 and at least from time to time wants to enjoy free weekend) what change do you propose to the original plan for wsit? remove the mess and revert original offending commit on the GF side to get back on track where MX versions are at least somehow dependent at least on their own jakartized API artifacts or keep current M1(*) having dependency on >10 non yet approved final artifacts (thus possibly legal issues for a ~week at least) and delay "jakarta" stuff to M2?

nucleus/parent/pom.xml
smillidge: yeah we are just trying to get GlassFish 6.0.0 RCs built so not too bothered by EMO approval at the moment as long it is in jakarta staging.

smillidge:
@lukasj I can wait for the jakartarised jar when you are ready

nucleus/parent/pom.xml
arjantijms: There has been some confusion yes, this is indeed all in staging only. 

Since Metro is not an API project (and I think doesn't use the spec plug-in, but I double check), I can use another suffix as well instead of -M1 if that makes things more clear?

Although it hasn't been set at the platform level I kinda assumed that -RC1 would be used for the Jakarta-ized version, but maybe I'm wrong there.

arjantijms:
> wsit 3.0.0-M1 was supposed to contain jakartized jaxb, saaj, jaxws, be dependent on jakartized eclipselink (jaxws, thus also metro, has a dependency on eclipselink), has JAXR/JAX-RPC implementations removed and going out around next weekend (currently it is ~70-75% complete)

That sounds great. For a fully jakartized wsit/webservices-osgi some more underlying dependencies have to be jakartized I guess (Servlet too, and Jakarta Authentication for instance). 

The WSIT in this PR just concerns a tmp WSIT compatible with only the removal of RPC. I'll see if I can re-stage using a name like "-RPC1", if that's okay?

lukasj:
this 3.0.0-whatever (note that -RPC will always be treated newer than RC since 'P' comes after 'C') includes jaxb/jaxws impls version 2.3.3 and these have not been approved/released[1] and I'm really not willing to redo them (and some of their dependenices) if they accidentally get to central sooner than they should. That would simply too much work to re-do.

If you really cannot wait and need something which would live in staging only and will be dropped once "real" M1 shows up, something like "-alpha" or "-b..." are better choices. 

[1]: https://bugs.eclipse.org/bugs/show_bug.cgi?id=561498
https://bugs.eclipse.org/bugs/show_bug.cgi?id=561497

arjantijms:
> this 3.0.0-whatever (note that -RPC will always be treated newer than RC since 'P' comes after 'C') 

Good point, although "newest" might not matter here since it's a one off build, I'll change to -alpha1 then. Thx!

> includes jaxb/jaxws impls version 2.3.3 and these have not been approved/released* and I'm really not willing to redo them (and some of their dependenices)

As Steve mentioned, that's okay. It's all in staging. 

I know it's all messy. I wish it was different, but for Mojarra I also had to do messy releases that weren't ready on request since others really needed it. I'll write a heads-up to the platform list to make it a little bit more clear for everyone (hopefully).

nucleus/parent/pom.xml
lukasj: there were many discussions and I believe the conclusion was use whatever suffix project wants but RC is recommended for APIs and M for impls. Ie hibernate artifacts are using ".Alpha" suffix to give different example

lukasj:
> Good point, although "newest" might not matter here since it's a one off build, I'll change to -alpha1 then. Thx!

even in that one off build case `mvn -U -C versions:display-property-updates -Pstaging` should keep showing correct version updates when appropriate

> As Steve mentioned, that's okay. It's all in staging.

as long as this won't leave staging, it should be fine.

nucleus/parent/pom.xml
arjantijms: Feedback addressed by re-staging as `-alpha1`. Thanks for your valuable feedback @lukasj

arjantijms:
> as long as this won't leave staging, it should be fine.

yes, it's staging only. Thx again!

PR #22971, opened 2020-04-28
Integrate Metro and Jersey 3.0.0-M1

lukasj:
Fixes: #22912, #22897, #22904 and #22916

with this change ~~and my - at this point local only - [changes](https://github.com/eclipse-ee4j/metro-wsit/pull/74) in metro,~~ glassfish builds with Jakarta XML Binding and XML Web services related APIs

few notes:
* there is no "jakarta" xjc maven plugin (and I do not plan to provide one), so what used to be generated by various maven-xjc plugins, is now being generated by xjc ant task (with one exception till [the bug](https://github.com/eclipse-ee4j/jaxb-ri/issues/1436) in the task is fixed)
* metro as such already requires "jakarta.servlet" packages, so to make jsr109-impl module compilable, I had to use both - javax.servlet:javax.servlet-api:4.0.1 as well as jakarta.servlet:jakarta.servlet-api:5.0.0-M1 - for the build
* I made all required jakarta apis not included in this PR, like jakarta.ejb,..., optional in webservices-osgi.jar assuming that as long as a deployment does not contain web service, jsr-109 and webservices-osgi modules will be still able to start and thus not blocking deployment for simple apps (untested)

Questions:
* ~~do you want me to relocate xml-binding and webservices related APIs out from the current 'endorsed' folder?~~ **DONE**
* I may also need to cover #22898 (persistence) as part of this ~~and maybe also #22899 (common annotations)~~ (#22899 is **DONE**) . Or is someone already working on it?

PR #22973, opened 2020-04-29
clean up metro/jersey integration

lukasj:
follow-up from #22971 to not forget:

- [x] remove snapshot repo
- [x] remove change in osgi.properties
- [x] update jersey to 3.0.0-M2
- [x] update metro to 3.0.0-M1

### DIFF
--- a/nucleus/osgi-platforms/felix/src/main/resources/config/osgi.properties
+++ b/nucleus/osgi-platforms/felix/src/main/resources/config/osgi.properties
@@ -379,10 +379,8 @@ jre-1.6=\
  org.xml.sax.ext, \
  org.xml.sax.helpers, ${endorsed-standard-packages}
 
-#TODO revert the version change in javax.annotation !!!
 endorsed-standard-packages=\
- javax.annotation; version=1.4, \
- javax.annotation.security; version=1.4, \
+ javax.annotation, \
  javax.xml.bind, \
  javax.xml.bind.annotation, \
  javax.xml.bind.annotation.adapters, \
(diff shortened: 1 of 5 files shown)


lukasj:
Metro 3.0.0-M1 has been already released few days ago. @arjantijms @smillidge I'll leave it up to you if you want to wait for Jersey 3M2 or rather get rid of snapshot dependency sooner

smillidge:
Currently we need Jersey 3 M2 for GlassFish to start.

nucleus/osgi-platforms/felix/src/main/resources/config/osgi.properties
```diff
@@ -379,10 +379,8 @@ jre-1.6=\
  org.xml.sax.ext, \
  org.xml.sax.helpers, ${endorsed-standard-packages}
 
-#TODO revert the version change in javax.annotation !!!
 endorsed-standard-packages=\
- javax.annotation; version=1.4, \
- javax.annotation.security; version=1.4, \
+ javax.annotation, \
```
smillidge, line 383: Needs to be Jakarta now

lukasj: I do not think so. This refers to JDK packages to be allowed for overriding by standard endorsed mechanism within OSGi. I explicitly added version number for GF to manage to resolve jersey bundles. The problem is that while this helps to some extent, Jersey requires Priority annotation which has never made it to JDK/Java SE, so with this hack, jersey is just failing a bit later then it would done without this.

smillidge (approved):

Issue #22884, opened 2020-01-27
Some options in domain.xml are mutually exclusive between JDK 8 and JDK 11.


hs536:
The current 5.1.0-run-with-JDK11 branch allows starting on JDK 11, but not JDK 8.  
  
I'm reconsidering to make GlassFish work with JDK 8 as well as JDK 11.  
First time, I thought I could implement it just like Payara Server or Maven.  
But now, I'm thinking about whether this is really appropriate...  
  
* Idea 1:(Now implementing: https://github.com/hs536/glassfish/commit/b5cbd833375af72949cfbb5328c916758a408f98 )
  Prefix the version range of JDK to the JVM option and filter them when reading domain.xml.
  ```
  domain.xml:
    <java-config>
      <jvm-options>[,8]-Djava.endorsed.dirs=XXX</jvm-options>
      <jvm-options>[9,]--add-opens=XXX</jvm-options>
      :
  ```
  + pros)
    - Only one source file needs to be managed.
  + cons)
    - This means that domain.xml contains JVM options that are not actually enabled on JDK 8 or JKD 11.
    - I think we should explain the meaning of this prefix to GlassFish users in documents or the like.
  
* Idea 2:
  Prepare a domain.xml for JDK8 (such as domain-lte-jdk8.xml) file and replace domain.xml when GlassFish start on JDK8. (On first time only)
  ```
  when user execute start-domain on JDK8 on first time,
    1. if static initialized_flag is false
    2.   if run on JDK8
    3.     replace domain-lte-jdk8.xml to domain.xml
    4.   change initialized_flag to true
    5. read domain.xml
  ```
  + pros)
    - GlassFish users can use domain.xml as before.
  + cons)
    - We'll end up managing multiple domain.xml files.
  
* Idea 3:
  Prepare a domain.xml for JDK8 (such as domain-lte-jdk8.xml) file and replace domain.xml when create-domain is executed with --jdk8 option.
  + pros)
    - GlassFish users can use domain.xml as before.
  + cons)
    - GlassFish users who want to run GlassFish on JDK 8 will have to re-create the domain.
    - We'll end up managing multiple domain.xml files.
  
* Idea4:
  Create a new profile for JDK 8 such as glassfish-jdk8.zip
  + pros
    - GlassFish users can use domain.xml as before.
    - There is no need to implement JDK version branching in scripts such as wsimport command.
  + cons
    - We'll end up managing multiple domain.xml files and command scripts.
  
[...]

PR #23018, opened 2020-05-20
WIP:Merge jdk11 work into master

hs536:
Note that GlassFish will not work with JDK 8 if this PR is approved due to an incompatibility of JVM options. (refs #22884)
In particular, Jenkins-build requires a container with both JDK 8 and JDK 11 installed. (JDK 8 for builds and JDK 11 for testing.)


hs536:
Unfortunately, these conflicts seem to need to be resolved locally. However, it is inconvenient to merge the master branch to the 5.1.0-run-with-JDK11 branch, so I will create PR again from another branch.

PR #22954, opened 2020-04-04
Fixes #22926 Removes Jakarta Deployment API from GlassFish 6.0.0

smillidge:
Passes Quick Lokk Tests

Signed-off-by: smillidge <steve.millidge@payara.fish>

PR #23035, opened 2020-05-27
Fixes admin console application deployment gui

jGauravGupta:
Signed-off-by: Gaurav Gupta <gaurav.gupta@payara.fish>


hs536 (approved):
it seems to be restoring the DFDeploymentProperties class as part of the admin console that was removed in the process of removing the Deployment API (#22954) . so LGTM.

PR #23063, opened 2020-06-02
Package Web Service apis in Web Feature Set

smillidge:
Required to fix OSGI errors on deployment in web profile 

Signed-off-by: smillidge <steve.millidge@payara.fish>

`Unable to resolve org.glassfish.main.web.weld-integration [128](R 128.0): missing requirement   [org.glassfish.main.web.weld-integration [128](R 128.0)] osgi.wiring.package; (&(osgi.wiring.package=jakarta.xml.ws)(version>=3.0.0)(!(version>=4.0.0))) Unresolved requirements: [[org.glassfish.main.web.weld-integration [128](R 128.0)] osgi.wiring.package; (&(osgi.wiring.package=jakarta.xml.ws)(version>=3.0.0)(!(version>=4.0.0)))]`

### DIFF
--- a/appserver/featuresets/web/pom.xml
+++ b/appserver/featuresets/web/pom.xml
@@ -55,6 +55,14 @@
         <!-- glassfish-jmx -->
         <!-- glassfish-management -->
         <!-- glassfish-corba-base -->
+        
+        <!-- Web Services APIs in Jakarta Namespace -->
+        <dependency>
+        <!-- Contains jakarta.jws, jakarta.xml.soap, jakarta.xml.ws -->
+            <groupId>org.glassfish.metro</groupId>
+            <artifactId>webservices-api-osgi</artifactId>
+            <version>${webservices.version}</version>
+        </dependency>
 
         <!-- glassfish-common -->
         <dependency>


smillidge:
@lukasj you will know more about this than me

lukasj:
which profile defines availability of jws/soap/xml.ws APIs? If web-profile then this change is correct. If full profile then this change is wrong and  weld-integration module should mark these imports as optional. I believe that the latter case is true.

lukasj (commented):
Looking at Java EE 6-8 where `profiles` are defined, web services based APIs are not being part of web profile. Based on that I believe that this change is wrong and if weld depends on these APIs, such dependency needs to be optional instead. Unless Jakarta EE 9 makes SOAP web services mandatory part of web-profile.

smillidge:
I think the problem likely arises in that they used to be in the JVM so there was no problem. I;m not sure this is coming from Weld as the module references is Weld Integration which is a GlassFish module.

smillidge:
@lukasj looks like you added the dependency into Weld Integration

https://github.com/eclipse-ee4j/glassfish/blame/master/appserver/web/weld-integration/pom.xml#L142

lukasj:
because of: https://github.com/eclipse-ee4j/glassfish/blob/master/appserver/web/weld-integration/src/main/java/org/glassfish/weld/services/InjectionServicesImpl.java#L37

lukasj:
and there is also https://github.com/eclipse-ee4j/glassfish/blob/master/appserver/web/weld-integration/src/main/java/org/glassfish/weld/services/InjectionServicesImpl.java#L256

smillidge:
Looks like we will have to ship the api in Web Profile. Not sure we have time to unpick Weld Integration

lukasj:
easy/fast way would be to mark that dependency "optional", so optional import is generated and replace offending calls either by reflection or by some helper interface

smillidge:
Not sure there is time before an M1 release. We could merge and then raise an issue to unpick for the Final release.

lukasj:
I can find an hour or two for this later today if you want...

smillidge:
Feel free to close my PR if you have a better one.

lukasj:
I'm not GF committer, so I can not close this...

PR #23066, opened 2020-06-02
Web-profile should not include soap web services

lukasj:
@smillidge this is what I've had in mind instead of yours #23063 Probably not the best solution I could come up with (this would be based on ServiceLocator) but I believe it is good enough for M1


lukasj:
Btw: I did not test this...

smillidge (approved):

PR #23075, opened 2020-06-04
Fix webservices tester and related QL failures

lukasj:
Signed-off-by: Lukas Jungmann <lukas.jungmann@oracle.com>


jGauravGupta (approved):
Thanks @lukasj

PR #23084, opened 2020-06-09
Added test script for local usage 

arjantijms:
This script works like the Jenkinsfile tests, but targets usage locally on the command line.

Update Readme to explain script usage

### DIFF
--- a/gftest.sh
+++ b/gftest.sh
@@ -1,0 +1,105 @@
+#
+# DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+#
+# Copyright (c) 2020 Oracle and/or its affiliates. All rights reserved.
+#
+# The contents of this file are subject to the terms of either the GNU
+# General Public License Version 2 only ("GPL") or the Common Development
+# and Distribution License("CDDL") (collectively, the "License").  You
+# may not use this file except in compliance with the License.  You can
+# obtain a copy of the License at
+# https://oss.oracle.com/licenses/CDDL+GPL-1.1
+# or LICENSE.txt.  See the License for the specific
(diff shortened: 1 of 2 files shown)


gftest.sh
```diff
@@ -0,0 +1,105 @@
+#
+# DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+#
+# Copyright (c) 2020 Oracle and/or its affiliates. All rights reserved.
```
smillidge, line 4: Shouldn't this be the Eclipse Foundation and shouldn't the license be the EPL?

smillidge (changes requested):

Pandrex247 (approved):
Script seems to work (though the tests don't seem to like my local env - couldn't find some logs)

gftest.sh
arjantijms: Yes, it probably should. I'll update it, thx!

gftest.sh
```diff
@@ -1,41 +1,17 @@
 #
-# DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+# Copyright (c) 2020 Eclipse Foundation and/or its affiliates. All rights reserved.
```
smillidge, line 2: Strictly the copyright is yours if it is original work see https://www.eclipse.org/projects/handbook/#ip-copyright-headers you license the work to the Eclipse foundation under the EPL

smillidge (approved):

gftest.sh
arjantijms: Okay, I'll keep that in mind for a next commit. I copied the loop from another test file, though I was the one (and only one) to have written that loop.

PR #23121, opened 2020-07-15
Fix for problem of parameter-encoding in glassfish-web.xml(eclipse-ee4j/glassfish#23108)

### LINKED ISSUE #23108: parameter-encoding of glassfish-web.xml is applied to response
## Environment Details
* GlassFish Version (and build number): 5.1
* JDK version: 8
* OS: Windows/Linux
* Database: N/A

----------

## Problem Description
[//]: # "Describe the bug in detail highlighting current behavior vs expected behavior"
[//]: # "State if the problem is easily reproducible or happens intermittently"
[//]: # "Include stack traces or command outputs"

## Problem 1
In GlassFish 5.1, the value specified in the default-charset attribute of the glassfish-web.xml parameter-encoding tag is applied not only to the parameter parsing of the request, but also to the default encoding of the response.
On the other hand, GlassFish 4.1 only applied to request parameter parsing.
As the GlassFish 5.1 documentation (https://eclipse-ee4j.github.io/glassfish/docs/5.1.0/application-deployment-guide/dd-elements.html#GSDPG00218) states, the behavior of GlssFish 5.1 seems to be wrong.

"Specifications the default request character encoding and how the web container codes parameters from forms according to a hidden field value."

The current behavior of GlassFish 5.1 is probably caused by source code modifications that allow web.xml to specify request-character-encoding and response-character-encoding tags.
https://github.com/javaee/glassfish/commit/dc28d4a375edaa9040849aefb35a81b5bc646754

However, since the name of the tag in glassfish-web.xml is "parameter-encoding" the current behavior of the "default-charset" attribute of the "parameter-encoding" tag affecting the default encoding of the response is strange.

In particular, when an application is migrated from GlassFish 4.1 to GlassFish 5.1, the Content-Type header of a response to static content such as html is given a charset attribute, which may cause garbled characters when displayed in a browser.

## Problem 2

For GlassFish 4.1, the setI18nInfo method of WebModule.java sets the value of parameter-encoding as defined in glassfish-web.xml.
On the other hand, in the Glassfish 5.1, the settings are performed in the following order.
1. WebModule#setI18nInfo Sets the web.xml value.
2. WebModule#setI18nInfo Sets the glassfish-web.xml value.
3. Sets the value of WebModule#start Sets the web.xml value.

The setI18nInfo method of 1, 2 will set the glassfish-web.xml value if both web.xml and glassfish-web.xml are defined. 
However, the start method in 3 sets the value of web.xml again.
Thus, if you specify both the request-character-encoding tag of web.xml and the default-charset attribute of the parameter-encoding tag of glassfish-web.xml, the value of web.xml is used.

Other properties whose values can be set in both web.xml and glassfish-web.xml use the glassfish-web.xml value.
So does WebModule#start need to override the value set by setI18nInfo?

## Steps to reproduce

[//]: # "Step by step instructions to reproduce the problem"
[//]: # "Provide sample code/application if relevant"

1. deploy hello.war
```
C:\glassfish5\glassfish\bin>asadmin start-domain
Waiting for domain1 to start ....................................................
Successfully started the domain : domain1
domain  Location: C:\glassfish5\glassfish\domains\domain1
Log File: C:\glassfish5\glassfish\domains\domain1\logs\server.log
Admin Port: 4848
Command start-domain executed successfully.

C:\glassfish5\glassfish\bin>asadmin deploy hello.war
Application deployed with name hello.
Command deploy executed successfully.
```

2. access "http://localhost:8080/hello/hello.html"
However, the content isn't displayed correctly. (We expect to be displayed ”Konnichiwa” in Japanese.)

3. check the Content-Type of response header. 
; charset="UTF-8" is set.

[hello.zip](https://github.com/eclipse-ee4j/glassfish/files/4876922/hello.zip)

tsyamamoto:
Fix #23108
Signed-off-by: tsyamamoto <tsyamamoto@fujitsu.com>


smillidge:
Should this fix also go into master?

gerdogdu:
I think we can merge these fixes to master.

smillidge:
@tsyamamoto can you also make a PR to master?

PR #23134, opened 2020-07-21
Update Hibernate Validator version

starksm64:
Update Hibernate Validator version to use final BV API/schemas.

Signed-off-by: Scott M Stark <starksm64@gmail.com>


starksm64:
Adding the signing line from the web editor is not working. I'll have to do a proper fork.

PR #23225, opened 2020-10-06
Update several dependencies to latest version.

arjantijms:

### DIFF
--- a/appserver/pom.xml
+++ b/appserver/pom.xml
@@ -158,7 +158,7 @@
 
         <!-- CDI -->
         <cdi-api.version>3.0.0</cdi-api.version>
-        <weld.version>4.0.0.Alpha3</weld.version>
+        <weld.version>4.0.0.Beta1</weld.version>
         <jboss.classfilewriter.version>1.2.1.Final</jboss.classfilewriter.version>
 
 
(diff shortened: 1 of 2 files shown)


smillidge:
Not sure I want to merge just a bunch of changes like this when we are getting close to a Full TCK run

arjantijms:
> Not sure I want to merge just a bunch of changes like this when we are getting close to a Full TCK run

If this would cause extra TCK failures it would be a matter of a quick revert, though maybe it's a moot point in this case as the PR is already failing. I'll take a very quick look at the failures tomorrow, and if it's not something extremely obvious I'll close the PR for now.

arjantijms:
p.s. the Weld version definitely would need to be updated from Alpha.3 to Beta.1 I think. I'll do a separate PR for that.

jGauravGupta (approved):

smillidge:
I'd prefer individual changes. I agree Weld will need an update and a few others. Best to do individual PRs as I got burnt by this a month or so back doing a bunch at the same time. I was also going to pick up a few.

arjantijms:
Ok, no worries. I'll close this in favour of individual PRs.

PR #23231, opened 2020-10-08
Adopt Jakarta Restful Web Services API 3.0.0 and Jersey 3.0.0-RC1

jansupol:


jansupol:
Jersey 3.0.0-RC1 does not contain Jackson module

PR #23251, opened 2020-10-22
Update GlassFish Admin Console Help Pages - Start

gerdogdu:
I have started to change help documents. Here is the initial commit for updating some help pages. Can someone also check that it is correct to continue?


smillidge (approved):

smillidge:
Looks good we should look how to split the work across the team

PR #23285, opened 2020-11-29
Prevent loading interceptors explicitly, as they might be private.

arjantijms:
For instance, the Jakarta Validation implementation doesn't export its
interceptor package: org.hibernate.validator.cdi.internal.interceptor

Loading anything from this package will crash GF.

Edit: the IDE did some auto-whitespace changes. Easiest to review using 

https://github.com/eclipse-ee4j/glassfish/pull/23285/files?diff=split&w=1


jGauravGupta (approved):
LGTM

PR #23349, opened 2021-02-09
JDK 11 compile

arjantijms:
This contains all the work to make GlassFish compile using JDK 11.

Compilation is the first step, and getting everything to work and pass all the tests are the following step.s


smillidge (approved):

smillidge:
OK this is too big to review but will approve

Issue #23377, opened 2021-02-18
Illegal reflective access by org.glassfish.web.loader.WebappClassLoader


arjantijms:
During deployment of for example `cdi-servlet-annotation-with-web-inf-lib-with-ordering-web` from the CDI_ALL test suite that's part of the GF code, the following warning is logged:

```
[#|2021-02-18T17:04:12.441+0100|INFO|glassfish 6.1|jakarta.enterprise.web|_ThreadID=73;_ThreadName=admin-listener(2);_TimeMillis=1613664252441;_LevelValue=800;_MessageID=AS-WEB-GLUE-00282;|
  Finished JSP Precompile...|#]

WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.glassfish.web.loader.WebappClassLoader (file:/glassfish/bundles/glassfish6/glassfish/modules/war-util.jar) to field java.lang.Thread.threadLocals
WARNING: Please consider reporting this to the maintainers of org.glassfish.web.loader.WebappClassLoader
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

hussainnm:
There are few instances of using `java.lang.reflect.Field::setAccessible` in `org.glassfish.web.loader.WebappClassLoader`

- java.lang.Thread
- java.lang.ThreadLocal
- sun.rmi.transport.Target
- sun.rmi.transport.ObjectTable
- java.util.ResourceBundle

Adding the following jvm-options in domain.xml will hide the message:
```
		<jvm-options>--add-opens=java.base/java.lang=ALL-UNNAMED</jvm-options>
		<jvm-options>--add-opens=java.base/java.util=ALL-UNNAMED</jvm-options>
		<jvm-options>--add-opens=java.rmi/sun.rmi.transport=ALL-UNNAMED</jvm-options>
```

smillidge:
Closed by #23380

PR #23380, opened 2021-02-20
#23377 Illegal reflective access by org.glassfish.web.loader.WebappClassLoader

hussainnm:


arjantijms (approved):
Eventually for JDK12+ we may need to address this further.

PR #23079, opened 2020-06-05
Revert jakarta.xml.bind-api & webservices-api-osgi from bootclasspath

jGauravGupta:
Signed-off-by: Gaurav Gupta <gaurav.gupta@payara.fish>

PR #23071, opened 2020-06-04
Fixes GlassFish Quicklook testcases

jGauravGupta:
Signed-off-by: Gaurav Gupta <gaurav.gupta@payara.fish>

PR #23382, opened 2021-02-21
Remove module jars from Xbootclasspath

hussainnm:
The below Xbootclasspath entries are not present in 6.0 and is added recently.

- jakarta.annotation-api.jar
- jakarta.xml.bind-api.jar
- webservices-api-osgi.jar

The server seems to start fine without these entries. Is there any reason for these jars to be added to the bootclasspath?

### DIFF
--- a/appserver/admin/template/src/main/resources/config/domain.xml
+++ b/appserver/admin/template/src/main/resources/config/domain.xml
@@ -204,9 +204,6 @@
         <jvm-options>-XX:NewRatio=2</jvm-options>
         <!-- Woodstox property needed to pass StAX TCK -->
         <jvm-options>-Dcom.ctc.wstx.returnNullForDefaultNamespace=true</jvm-options>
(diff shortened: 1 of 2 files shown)


hussainnm:
This also resolves the following warning thrown on startup

```
JAXBContext implementation could not be found. WADL feature is disabled
```

arjantijms:
These came in with the following commit:

Commit 17f259f

Essentially those moved from endorsed to modules. If all test still run after this and the admin console boots up it should be fine to remove them.

hussainnm:
See comments in #23071, per Lukas these are not required. Tests were fixed in #23075 and these entries removed in #23079. These commits happened 15 days after Hiroki's commit.

arjantijms:
@hussainnm I see, that makes sense.

arjantijms (approved):

smillidge:
Admin console no longer starts

PR #23385, opened 2021-02-23
Fix QuickLook Full Profile Test

hussainnm:
The Apache Derby 10.15.2.0 seems to require `permission java.lang.RuntimePermission "getenv.SOURCE_DATE_EPOCH"` to be added to java.policy file to start the database in Network Mode.

The below file needs to be created in the user home directory
~/.java.policy
```
grant {
  permission java.lang.RuntimePermission "getenv.SOURCE_DATE_EPOCH";
};
```

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -21,7 +21,8 @@ env.label = "glassfish-ci-pod-${UUID.randomUUID().toString()}"
 
 // list of test ids
 def jobs = [
-  "cdi_all"
+  "cdi_all",
+  "ql_gf_full_profile_all"
 ]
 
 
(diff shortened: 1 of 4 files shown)


smillidge (approved):

hussainnm:
I think this permission is required only in my system. I tried to run the cdi_all test in local and it failed with the same error.

arjantijms:
Indeed, that permission wasn't needed here (macOS 11.2). I did notice Derby now runs by default with a security manager, which it sets programmatically. On macOS this has a side-effect that it runs through all your so-called "utun" devices, which takes 5 seconds per device. This on its turn causes a time-out for the process that starts Derby.

PR #23404, opened 2021-03-01
Integrate xml binding API 3.0.1, impl 3.0.1-b01

lukasj:
activation 2.0.1

this also removes `javax` deps which were added after 6.0

NOTE: API dependencies are in staging only and should not be published yet

@arjantijms jaxb-impl 3.0.1-b01 has been updated in staging and QL are passing for me now.


arjantijms:
Thanks @lukasj indeed, the QL tests that run as part of this PR passed.

arjantijms (approved):

PR #23410, opened 2021-03-07
Fix the web_jsp devtests

arjantijms:


hussainnm (changes requested):
Is there a need to check-in the below class file?
- appserver/tests/appserv-tests/devtests/web/jspReloadGeneratedServletIfUpdated/test_jsp.class

Do we need to check-in the jar files if they can be generated at build time?

- appserver/tests/appserv-tests/devtests/web/jspConfigUsePrecompiled2/lib/jsps.jar
- appserver/tests/appserv-tests/devtests/web/jspCustomTaglibBundledAsJar/lib/mytaglib.jar
- appserver/tests/appserv-tests/devtests/web/jspInWebappWithBundledXercesImpl/lib/mytaglib.jar

arjantijms:
> Do we need to check-in the jar files if they can be generated at build time?

Technically I'd rather not, but that's the way the tests have already been setup, so this is the least effort way to make them run again. If you'd want to update the PR to generate the jar at build time, then +1 from me.

hussainnm (approved):

smillidge (approved):

PR #23416, opened 2021-03-10
Update test dependencies (junit/testng/easymock)

arjantijms:

### DIFF
--- a/appserver/pom.xml
+++ b/appserver/pom.xml
@@ -171,7 +171,7 @@
         <schema2beans.version>6.7</schema2beans.version>
         <derby.version>10.15.2.0</derby.version>
         <wsdl4j.version>1.5.0</wsdl4j.version>
-        <easymock.version>3.5</easymock.version>
+        <easymock.version>4.2</easymock.version>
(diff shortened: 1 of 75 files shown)


smillidge (changes requested):
Would it be easier to pull the version number into a parent pom property or parent pom dependency rather than update 75 files :-)

arjantijms:
> Would it be easier to pull the version number into a parent pom property or parent pom dependency rather than update 75 files :-)

Normally, sure, but many (all?) of those files don't have a parent to begin with.

smillidge:
Could we add one?

arjantijms:
> Could we add one?

I guess we could, but may be some work ;)

dmatej:
You can add junit.version to aggregator and add junit to dependeny management. And also set default scope to test scope. Then you don't need to declare version and scope in each child.

smillidge:
Your call. I will approve

smillidge (approved):

arjantijms:
Okay, I'll merge this now as-is, as it updated just the previous version. For a new PR we can then introduce a parent and move the version there. Thanks!

Issue #23371, opened 2021-02-15
Weld failure: java.lang.ClassFormatError: Illegal class name 


arjantijms:
During deployment when injecting a CDI bean into an EJB.

Code:

```java
@RequestScoped
public class ImplicitTestBean {

}
```

```java
@Stateless
public class SlessEJB implements Sless {

    @Inject
    ImplicitTestBean testBean;
}
```

Exception:

```
java.lang.ClassFormatError: Illegal class name "/Sless$365510969$Proxy$_$$_Weld$EnterpriseProxy$" in class file /Sless$365510969$Proxy$_$$_Weld$EnterpriseProxy$
	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1016)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.jboss.weld.util.bytecode.ClassFileUtils.toClass2(ClassFileUtils.java:143)
	at org.jboss.weld.util.bytecode.ClassFileUtils.toClass(ClassFileUtils.java:112)
Caught: java.lang.RuntimeException: java.lang.ClassFormatError: Illegal class name "/Sless$365510969$Proxy$_$$_Weld$EnterpriseProxy$" in class file /Sless$365510969$Proxy$_$$_Weld$EnterpriseProxy$
	at org.jboss.weld.util.bytecode.ClassFileUtils.toClass(ClassFileUtils.java:116)
	at org.jboss.weld.bean.proxy.ProxyFactory.createProxyClass(ProxyFactory.java:481)
	at org.jboss.weld.bean.proxy.ProxyFactory.getProxyClass(ProxyFactory.java:370)
Caught: org.jboss.weld.exceptions.WeldException: WELD-001524: Unable to load proxy class for bean Session bean [class SlessEJB with qualifiers [@Any @Default]; local interfaces are [] with class class SlessEJB
	at org.jboss.weld.bean.proxy.ProxyFactory.getProxyClass(ProxyFactory.java:378)
[...]

PR #23434, opened 2021-03-26
#23371 Update to Weld 4.0.1.Final

hussainnm:


arjantijms:
PRs currently fail because the fs where the maven cache resides is not longer writable. See https://bugs.eclipse.org/bugs/show_bug.cgi?id=572345

In addition, I think 4.0.1 can not just be added, but first needs an update of the Weld SPI implementation.

hussainnm:
I withdraw this pull request as `cdi_all` tests fail.

PR #23466, opened 2021-04-28
Weld cleanup

arjantijms:
Some amount of formatting and cleanup for the Weld integration code.

For review it's easiest to look at the two individual commits separately:

# Formatting only:

* https://github.com/eclipse-ee4j/glassfish/pull/23466/commits/92c085726ca49c1664b0c4fd4b1467c507b95b93?diff=split&w=1

# Organising imports, missing braces and other minor cleanups:

* https://github.com/eclipse-ee4j/glassfish/pull/23466/commits/a23b7df61907a5151f53af650b163309985cf747?diff=split&w=1


gerdogdu:
This type of large commits are not easy to follow :) Is it possible to create a several issue and then patch small PRs? WDYT?

arjantijms:
@gerdogdu I'm actually already smaller issues ;) This is actually mostly just formatting, and automated cleaning. I'm trying to do that module by module. If done for the entire code base, it would be much larger indeed.

It should help a lot already if ignoring whitespace:

https://github.com/eclipse-ee4j/glassfish/pull/23466/files?diff=split&w=1

arjantijms:
@gerdogdu If you want, you could also look at just the first say 20 files. They're very quick to scroll through, and then say here: first X files LGTM. Then the next reviewer can look at the remaining files. Would that work?

gerdogdu:
@arjantijms What is your formatting xml? Do we have a project wide ? Which tool do you use for cleaning?

arjantijms:
@gerdogdu for the general formatting in Java I use the code convention files from https://github.com/eclipse-ee4j/ee4j/tree/master/codestyle I use the Eclipse version, but the IntelliJ one should give similar results.

For XML formatting I use Eclipse as well, and use the defaults with spaces (as per the Jakarta convention) instead of tabs.

jGauravGupta (approved):

PR #23469, opened 2021-05-04
Update TCK-Results-6.0.md

edbratt:
Fix Web Profile test doc. mistake also.
(Was noticed on platform results but not web-profile but incorrect in both :-( )


m0mus (approved):
LGTM

PR #23472, opened 2021-05-18
Trim all trailing whitespaces and easy formatting when necessary

dmatej:
Fixed all trailing spaces and mixed whitespaces issues and added checkstyle plugin to the default lifecycle. **Ldapbp** build fixed and integrated to the build (if not supported, kill and delete it, but until that don't ignore it).

- all tabs were replaced with 3-8 spaces
-- default is 4, when I have seen that some module uses other value more often, I changed it)
- removed all trailing spaces from known textual files (java|properties|xml|sh|sql|bat)
- bat files use CRLF
- all other text files use LF
- properties files 
-- if there was a trailing space, my script made last space visible by using unicode value `\u0020` - all such changes were reviewed manually, because sometimes it was a mistake, sometimes the space was intentional
- fixed some i18n messages (see commits, usually missing {0} for a parameter)
- sometimes updated imports (usually removed unused imports)
- sometimes removed redundant `public` keyword from interfaces
- all changes were reviewed manually, when the formatting was too bad to read it, I changed it using Eclipse + manually to improve it. In many cases I avoided it, because
-- this PR would be too big
-- it would take too much time
-- further changes should better go over modules/features/etc before some standard change, so it would be easily readable
-- this is just a step an the long path to remove extreme technical debt; it also reduces size of future PRs, so it will be more visible what changed, while this PR functionally doesn't change anything ... except ...
- added checkstyle rules to prevent fixed issues (**zombies**) comming back
- another improvements are waiting after this would be merged


arjantijms:
Good change! Yes, this was really needed. A couple of other projects have already gone through the same of similar, but GF wasn't done yet. Indeed, because of the size of the resulting PR.

Using https://github.com/eclipse-ee4j/glassfish/pull/23472/files?diff=split&w=1 for the initial look, and a local client for some of the bigger files made it easier to look through.

arjantijms (approved):

PR #23473, opened 2021-05-19
Use current Ant version with maven-antrun-extended-plugin for faster rejar

dmatej:
- original plugin uses 1.6.5
- 1.10.10 is much faster
- result is nearly the same
  - smaller file, same content, more meaningful file permissions


arjantijms (approved):
LGTM! Nice speed increase ;)

PR #23486, opened 2021-06-07
Issue #23485 Removals + easier updating version id

dmatej:
- Removed unused or duplicit modules
- added profile and script for easier changing the version id - the script may be removed later after we "resuscitate" all test modules in the repository, which will be some larger amount of work.

PR #23487, opened 2021-06-10
Small cleanups in build

dmatej:
- In https://github.com/eclipse-ee4j/glassfish/pull/23486 I forgot to update the minor version from 1 to 2
- Checkstyle had issues with the project root resource folder added for LICENSE and NOTICE files, which should be added into each artifact. Checkstyle used the directory, but ignored the filter and replaced it with it's own - `**/*.properties`. This PR resolves it by excluding paths added by mistake; probably it would be better later replace that adding external resources with some assembly artifact, so we would be compliant with maven rules.
- While I was working on that, I noticed that when I use the `-pl `parameter, LICENSE and NOTICE files are not added to the artifact, because Maven resolves the directory to the basedir of the project defined by the `-pl `parameter. However since Maven 3.3.1 it is possible to use another parameter: `maven.multiModuleProjectDirectory`
- I removed those 'auto-clean' executions, because they obey standard Maven rules and can be simply done by standard execution of the `clean` phase.

### Tests
```
mvn clean install -Pstaging -TC8 -pl :acc-config
mvn install -Pstaging -TC8 -pl :acc-config
mvn clean install -Pstaging -TC8
Total time:  04:01 min (Wall Clock)
mvn install -Pstaging -TC8
Total time:  03:30 min (Wall Clock)
mvn clean install -Pstaging -TC16
Total time:  04:08 min (Wall Clock)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,6 @@ nb-configuration.xml
 .DS_Store
 .project
 .settings/
-appserver/appclient/client/acc-config/temp/
 appserver/security/jaspic-provider-framework/auth.conf
 appserver/tests/appserv-tests/devtests/web/JSESSIONID
 /bundles/
--- a/appserver/admingui/cluster/pom.xml
+++ b/appserver/admingui/cluster/pom.xml
@@ -31,27 +31,6 @@
     <name>Admin Console Clustering Support Plugin</name>
     <description>Clustering support plugin bundle for GlassFish Admin Console</description>
 
-    <developers>
-        <!-- See parent POM -->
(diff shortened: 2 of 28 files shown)


arjantijms:
> probably it would be better later replace that adding external resources with some assembly artifact, so we would be compliant with maven rules.

Yes, I was thinking exactly this. If I'm not mistaken, this is what the GF license plugin did or at least there was a license dependency before in GF (before switching over to Eclipse).

arjantijms (approved):
LGTM!

PR #23506, opened 2021-07-05
Use GF-customized OpenMQ conf file

### LINKED ISSUE #23499: Customize OpenMQ's `IMQ_DEFAULT_VARHOME` for the needs of GlassFish
Consider performing GF customization of `MQ`:`etc/imqenv.conf` in this project.
Up to currently bundled OpenMQ 6.1.1, the `imqenv.conf` file was distributed with the following line added:
```
IMQ_DEFAULT_VARHOME=../../glassfish/domains/domain1/imq
```

If this is no longer present, different `varhome` would be used by standalone broker.
Embedded one seems to use domain directory in such case.

If the `IMQ_DEFAULT_VARHOME` pointing to `domain1` location is needed - please update GF build process to not depend on the mentioned variable being set.
If it isn't - feel free to close this issue with no action.

pzygielo:
Closes #23499

### DIFF
--- a/appserver/distributions/glassfish/src/main/mq-etc/imqenv.conf
+++ b/appserver/distributions/glassfish/src/main/mq-etc/imqenv.conf
@@ -1,0 +1,18 @@
+#!/bin/sh
+#
+# Copyright (c) 2021 Contributors to the Eclipse Foundation
+#
+# This program and the accompanying materials are made available under the
+# terms of the Eclipse Public License v. 2.0, which is available at
+# http://www.eclipse.org/legal/epl-2.0.
+#
+# This Source Code may also be made available under the following Secondary
+# Licenses when the conditions for such availability set forth in the
+# Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+# version 2 with the GNU Classpath Exception, which is available at
+# https://www.gnu.org/software/classpath/license.html.
+#
+# SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+#
+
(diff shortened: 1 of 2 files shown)


appserver/distributions/glassfish/src/main/mq-etc/imqenv.conf
```diff
@@ -0,0 +1,18 @@
...
+# This Source Code may also be made available under the following Secondary
+# Licenses when the conditions for such availability set forth in the
+# Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+# version 2 with the GNU Classpath Exception, which is available at
+# https://www.gnu.org/software/classpath/license.html.
+#
+# SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+#
+
+IMQ_DEFAULT_VARHOME=../../glassfish/domains/domain1/imq
```
dmatej, line 18: What if I would use create-domain command to create a new domain with different name? (I am asking before I would try it)

pzygielo: 1. With this change, `imqenv.conf` bundled with GF is EXACTLY (with the exception for (c)) as before. But the `IMQ_DEFAULT_VARHOME` will (probably) no longer be set in `mq.zip` in next releases. This PR is to make GF ready for that.
1. As noted in linked issue - embedded broker uses domain dir, regardless of its name or location. It's up to you then, to configure broker I guess, as it was before.

arjantijms (approved):
Thanks, this mirrors what's in OpenMQ now, even though the static domain set looks weird in this context.

PR #23517, opened 2021-07-09
Issue #23516 Ant replaced by Maven in Admin GUI build

### LINKED ISSUE #23516: Admin GUI build: don't use Ant where you can use Maven
## Problem Description

* In 8cf09e6259f427138022fe7c997a60dc9c413f11 we broke UI - sun-web.xml still referes the commons-io 2.8.0
* Nothing in build warns that UI will not work, you have to start it and watch logs.
* The build should be more reliable at least

dmatej:
- dependencies are managed by Maven now, so upgrades of dependencies will be
  more easy next time
- fixed commons-io dependency broken in 8cf09e6259f427138022fe7c997a60dc9c413f11 and fixes #23516
- javahelp removed from dependencies - unused, not maintained since 2007
- jsf-extensions removed too - they were on build classpath, but were not distributed

### DIFF
--- a/appserver/admingui/core/pom.xml
+++ b/appserver/admingui/core/pom.xml
@@ -32,86 +32,8 @@
 
     <dependencies>
         <dependency>
-            <groupId>org.glassfish.main.admingui</groupId>
-            <artifactId>console-plugin-service</artifactId>
-            <version>${project.version}</version>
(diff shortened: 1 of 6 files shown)


arjantijms (approved):
Nice! :)

PR #23518, opened 2021-07-10
Bump OpenMQ version

pzygielo:
This OpenMQ version:
- needs at least Java 11 to run,
- has no GlassFish customization in `etc/imqenv.conf` (but this file is adjusted for GF in this project now).


dmatej:
Depends on #23506

dmatej (approved):

PR #23519, opened 2021-07-11
Resolve `build.id` with git.* properties

pzygielo:
Before
```
$ ./glassfish6/glassfish/bin/asadmin version
Version string could not be obtained from Server [localhost:4848].
(Turn debugging on e.g. by setting AS_DEBUG=true in your environment, to see the details.)
Using locally retrieved version string from version class.
Version = Eclipse GlassFish  6.2.0  (build ${git.branch}-b${git.closest.tag.commit.count}-g${git.commit.id.abbrev} ${git.build.time})
Command version executed successfully.
```
After
```
 ./glassfish6/glassfish/bin/asadmin version
Version string could not be obtained from Server [localhost:4848].
(Turn debugging on e.g. by setting AS_DEBUG=true in your environment, to see the details.)
Using locally retrieved version string from version class.
Version = Eclipse GlassFish  6.2.1  (build revid-b463-gd891f35 2021-07-11T22:14:56+0200)
Command version executed successfully.
```
and for web console: 
revid-console

Fixes #23489.

PR #23524, opened 2021-07-12
Simple project to run test for resolved build_id

pzygielo:
Related to #23519, to avoid regression.

There is most probably better way to do it, but with `<packaging>distribution-fragment</packaging>` I'm not willing to (re)configure test compilation and execution. Filtered resources are not in default test-classpath in that module anyway.

### DIFF
--- a/appserver/distributions/glassfish-common/pom.xml
+++ b/appserver/distributions/glassfish-common/pom.xml
@@ -89,6 +89,21 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <artifactId>maven-invoker-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>integration-test</id>
+                        <goals>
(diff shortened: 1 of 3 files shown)


dmatej:
I'm still thinking about this - it seems to me like an overkill ... I would rather resuscitate some integration tests soon and then simply execute the asadmin command and also verify that the same is in UI:
```
/app/appservers/glassfish6/glassfish/bin/asadmin version
Version = Eclipse GlassFish  6.2.1  (build tests-b481-g794a9a2 2021-07-20T21:23:46+0200)
Command version executed successfully.
```

pzygielo:
> resuscitate some integration tests soon and then simply execute the asadmin command and also verify that the same is in UI

It can still be done to perform (heavy) integration testing. This PR is intended to test in the place where the thing happens, not where it is used.

PR #23533, opened 2021-07-29
Integrate Exousia 1.0.0 M2 and adjust enterprise beans code

arjantijms:

### DIFF
--- a/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
+++ b/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1997, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1997, 2021 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -33,6 +33,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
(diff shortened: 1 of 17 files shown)


appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
```diff
@@ -119,110 +114,84 @@
...
         STATELESS, STATEFUL, SINGLETON, MESSAGE_DRIVEN, ENTITY, READ_ONLY
     };
 
-    protected static final Logger _logger  = LogFacade.getLogger();
+    protected static final Logger _logger = LogFacade.getLogger();
 
-    @LogMessageInfo(
-        message = "The feature {0} requires Full Java EE Profile to be supported",
-        level = "WARNING")
+    @LogMessageInfo(message = "The feature {0} requires Full Java EE Profile to be supported", level = "WARNING")
```
dmatej, line 164: The original formatting was more readable ...

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
```diff
@@ -234,33 +203,30 @@
...
+    static final int EJBLocalHome_create = 15;
 
     // true if home method, false if component intf method.
     // Used for setting info on invocation object during authorization.
-    private static final boolean[] EJB_INTF_METHODS_INFO =
-    { true,  true,  true,  true,  true,
-      false, false, false, false, false,
-      false, false, false, false,
-      true,  true };
+    private static final boolean[] EJB_INTF_METHODS_INFO = { true, true, true, true, true, false, false, false, false, false, false, false,
```
dmatej, line 225: I dislike previous array of booleans as I have no idea what it is, and this is probabaly even worse. But ok, the biggest issue is that it is an array of 16 booleans.

appserver/ejb/ejb-container/src/main/java/org/glassfish/ejb/security/application/EJBSecurityManager.java
```diff
@@ -785,21 +359,19 @@ public Principal getCallerPrincipal() {
...
              * own permission cache, which must be unregistered from factory to
-             * avoid leak.
+             * avoid leaks.
              */
             PermissionCacheFactory.removePermissionCache(uncheckedMethodPermissionCache);
             uncheckedMethodPermissionCache = null;
-            roleMapperFactory.removeAppNameForContext(this.contextId);
+            roleMapperFactory.removeAppNameForContext(contextId);
 
-        } catch (PolicyContextException pce) {
```
dmatej, line 801: This smells, catching just ISE? What about other RTE? They should or should not be propagated? At least comment why is this catch block so special would be fine.

dmatej (changes requested):
See individual comments. Also some explanation would be fine, and some info about usage and testing in the PR description ;)

appserver/ejb/ejb-container/src/main/java/org/glassfish/ejb/security/application/EJBSecurityManager.java
arjantijms: The previous code caught PolicyContextException as a special case. This has been replaced by an IllegalStateException. It just maintains that behaviour.

dmatej: Hmm, I don't like it much, but ok.

dmatej (approved):
I have yet another thought about @LogMessageInfo annotation. This formatting is ugly, but the level is used just for documentation purpose, which is not regenerated now (or am I wrong?) and the original work around logging never finished.
So I accept it as it is (build passed).

arjantijms:
> I have yet another thought about @LogMessageInfo annotation. This formatting is ugly, but the level is used just for documentation purpose, which is not regenerated now (or am I wrong?) and the original work around logging never finished.

Yes, I just peeked at it. Don't know what to do with it atm.

Issue #23507, opened 2021-07-05
Give tests some love and move them >18 years forward


dmatej:
## Problem Description

Before we start some more rapid development, we should get the best from existing tests. So the target of this issue is to do what is possible:

- [x] Migrate to JUnit5
- [x] hk2-junitrunner is probably obsoleted, investigate alternatives
- [x] Clean up test dependencies (junit4 is everywhere implicitly even if unused)
- [x] Transform TestNG tests to JUnit5 (not maintained for a long time)
- [x] Resuscitate or delete disabled tests if possible (may have it's own issues, it depends on future analysis and some future plan depending on result of previous steps)
- [x] Integrate maven profile to run tests with JaCoCo (target: to see lines uncovered by any test, detect risks and be able to improve the state)
- [x] Burial of console-devtests maven project (uses obsoleted selenium and firefox, unused too long)
- [x] Resuscitate nucleus-tests
- [x] Resuscitate appserver/tests/... + quicklook

## Impact of Issue

- better coverage
- more capabilities - ordering, parallel assertions
- maintenance

## Notes

This issue will be probably updated later as I don't see all consequences and targets yet, but from my experience they will be visible soon.

PR #23535, opened 2021-07-30
Issue #23507 TestNg is out

dmatej:
- the first step of the quest - replace TestNg by JUnit5
- I didn't bother much with tests which are not executed at all and are broken and not much maintainable, but the rest was migrated
- the work will continue and I hope that many currently disabled tests will be active again
- testng utils were not used by anyone, based on statistics of mvnrepository


arjantijms (approved):
Yes, major milestone! :)

Issue #23542, opened 2021-08-17
Investigate if we still require fixup.xml files


dmatej:
@arjantijms noticed that manifest files packaged into jars are randomly radically different and resulting jar then crashes some tests. After we have read documentation of the [Jar Task](https://ant.apache.org/manual/Tasks/jar.html) we tried to avoid the timing issue by using the roundup attribute, but it did not help enough, so Arjan just increased the sleep time in 92bf32e8b5c1d575 .

The question is that after years of evolution especially of the maven jar plugin - can we get rid of this "hack"?

dmatej:
Done

PR #23543, opened 2021-08-17
Issue #23542 Removed fixup.xml

dmatej:
- added dependecies and configured bundle+jar plugins instead of explicit
  naming in manifest and fixing classpath by Ant
- aix-repackaged is not a part of GlassFish and this was the only place where
  it was mentioned, so I removed it


arjantijms (approved):
LGTM

PR #23554, opened 2021-08-20
Separate words in logged messages

pzygielo:


nucleus/admin/util/src/main/java/com/sun/enterprise/admin/util/AdminLoggerInfo.java
```diff
@@ -50,7 +51,7 @@ public static Logger getLogger() {
     @LogMessageInfo(message = "unable to read instance state file {0}, recreating", level = "FINE")
     final static String mISScannotread = LOGMSG_PREFIX + "-00003";
 
-    @LogMessageInfo(message = "unable to create instance state file: {0}, exception: {1}", cause = "The instance state file is missing and the system is trying to"
+    @LogMessageInfo(message = "unable to create instance state file: {0}, exception: {1}", cause = "The instance state file is missing and the system is trying to "
```
pzygielo, line 54: Please suggest rewording.

nucleus/admin/util/src/main/java/com/sun/enterprise/admin/util/AdminLoggerInfo.java
```diff
@@ -50,7 +51,7 @@ public static Logger getLogger() {
...
             + "recreated it but and exception was raised.", action = "Check the server logs and contact Oracle support", publish = true, level = "SEVERE")
```
pzygielo, line 55: Not sure if Oracle support should be referenced here...

dmatej: Are those messages still used? When I was in Payara, I found that many of them are simply unused, dead. But verifying them one by one takes too much time.

dmatej: I haven't seen it since SGES2.

dmatej (approved):

nucleus/admin/util/src/main/java/com/sun/enterprise/admin/util/AdminLoggerInfo.java
arjantijms: I noticed that same message. Indeed, time consuming to check everything. My preference would be to not have messages in separate files, and have them directly in the logger, or as constants in the class that uses them.

PR #23555, opened 2021-08-20
Revert "Additional other small cleanups - add missing @Override etc"

dmatej:
This reverts commit 21c31b9423c8bda9fae878fbd7427d7f3f6c74fd.

With the 21c31b9423c8bda9fae878fbd7427d7f3f6c74fd GlassFish starts with exception in the server.log. 
I wasn't able to find to exact cause, so I'm reverting the commit.

image


dmatej:
@arjantijms If you can find another solution of the bug without reverting it, we can close this PR then, but at least we have some solution :)

arjantijms:
+1 For revert for now, if the CI clears. I'll try to do the parts of the PR even more individually next. Like "@override" in a separate commit, etc.

arjantijms (approved):

PR #23560, opened 2021-08-22
Synchronize Maven prerequisite in doc

pzygielo:
https://github.com/eclipse-ee4j/glassfish/blob/efe3cd18411e991e0af9561f0a170c1414831280/pom.xml#L86-L88


arjantijms (approved):
Good catch!

PR #23561, opened 2021-08-23
Use CS:RedundantModifier

pzygielo:


pzygielo:
Needs something like #23561.

PR #23562, opened 2021-08-23
Organize imports

pzygielo:


appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/connectors/deployment/annotation/handlers/ConfigPropertyHandler.java
```diff
@@ -508,7 +507,7 @@ public static void processParent(Class claz, Set configProperties) {
             if (property != null) {
                 String result = validateMethod(m, property);
                 if (!result.equals(SUCCESS)) {
-                    throw new IllegalStateException(result);
+                    throw new java.lang.IllegalStateException(result);
```
dmatej, line 510: Why???

qa/src/main/resources/org/glassfish/qa/config/checkstyle/checkstyle.xml
```diff
@@ -82,6 +82,7 @@
         <!-- Checks for imports                              -->
         <!-- See http://checkstyle.sf.net/config_import.html -->
         <module name="IllegalImport" /> <!-- defaults to sun.* packages -->
+        <module name="RedundantImport" />
```
dmatej, line 85: I agree, if it is fixed everywhere, but not by using package names like java.lang.Object in code. If there are conflicts, ie. org.glassfish.Object, I would prefer JDK in imports and conflicting class with full name in code (except if there would be 100:1 occurences).

appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/connectors/deployment/annotation/handlers/ConfigPropertyHandler.java
pzygielo: Reverted to not-FQ name.

qa/src/main/resources/org/glassfish/qa/config/checkstyle/checkstyle.xml
pzygielo: Conflicts resolved by replacing on-demand imports with individual ones.

pzygielo:
reworked

dmatej (approved):

PR #23563, opened 2021-08-23
Issue #23507 Integration of JaCoCo and SonarQube

dmatej:
- SonarQube parses XML reports, not exec files
- added also com.sun.enterprise packages

Signed-off-by: David Matějček <dmatej@seznam.cz>


arjantijms (approved):
LGTM, hope we can have the parent as a real parent one day ;)

PR #23564, opened 2021-08-23
Remove references to MQ:jaxm-api.jar

pzygielo:
The upcoming OpenMQ version will not deliver `jaxm-api.jar`.
This PR serves as reminder to remove references to that archive.


pzygielo:
Will re-open once new OpenMQ is proposed here.

PR #23579, opened 2021-09-01
Fix #23578 Cannot enbale Security Manager from the admin console

### LINKED ISSUE #23578: Cannot enbale Security Manager from the admin console.
An error occurs when enabling Security Manager from the admin console.

## Environment Details
* GlassFish Version (and build number): Eclipse GlassFish  6.2.1  (build master-b568-g0f9a1e7 2021-08-30T05:04:52+0000)
* JDK version: OpenJDK 11.0.12+7
* OS: Red Hat Enterprise Linux release 8.0
* Database:

----------

## Problem Description
If you enable Security Manager from the admin console, you will get an error:
image

## Steps to reproduce
1. Create a new Cluster, eg SecMgrTest1
2. Navigate to new cluster's config, Configurations > SecMgrTest1-config
3. Navigate to cluster's Security page
4. Check the Enabled checkbox for "Security Manager"
5. Click "Save" button.

## Impact of Issue
We cannot enable Security Manager from the admin console.

11rx4f:
Fixed that failed to enable Security Manager on admin console.  
It may become unnecessary soon because of JEP 411, but it is a bug at this time, so I fixed it.

QuickLookTest was successful in local environment.

### DIFF
--- a/appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/SecurityHandler.java
+++ b/appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/SecurityHandler.java
@@ -770,7 +770,8 @@ public class SecurityHandler {
             Map<String, Object> payload = new HashMap<String, Object>();
             payload.put("target", configName);
             for (String option : newOptions) {
-                ArrayList kv = InstanceHandler.getKeyValuePair(option);
+                String option1 = UtilHandlers.escapePropertyValue(option);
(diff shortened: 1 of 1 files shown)


arjantijms:
Looks good, but as a convention, we try to avoid names like [somename]x, but instead try to explain its difference, e.g. "escapedOption" instead of "option1". But that's a small nit in this context. 

Thanks again!

arjantijms (approved):

PR #23581, opened 2021-09-06
Fix #23580 Refer to parameter-encoding only if charset is not specified.

### LINKED ISSUE #23580: parameter-encoding in glassfish-web.xml overrides the charset specified by the client.
## Environment Details
* GlassFish Version (and build number): 6.2.1
* JDK version: OpenJDK 11.0.12+7
* OS: Cent OS 7
* Database: -

----------

## Problem Description
If parameter-encoding is set in glassfish-web.xml, GlassFish overwrites the value of charset specified by the client, causing garbled characters in the application.
This problem does not occur on GlassFish 4, but on GlassFish 5 and later.

## Steps to reproduce
1) Deploy an application that returns a response with POST data strings that is not in iso-8859,
   specifying iso-8859 in glassfish-web.xml.

    TestServlet.java
    ```
    @WebServlet("/TestServlet")
    public class TestServlet extends HttpServlet {
    	protected void doPost(HttpServletRequest request, HttpServletResponse response)
    			throws ServletException, IOException {
    		String requestContentType = request.getContentType();
    		response.setCharacterEncoding("UTF-8");
    		PrintWriter out = response.getWriter();
    		out.println(requestContentType);
    		BufferedReader reader = request.getReader();
    		try {
    			String line;
    			while ((line = reader.readLine()) != null) {
    				sb.append(line).append('\n');
    			}
    		} finally {
    			reader.close();
    		}
    		out.println(sb.toString());
    		out.close();
    	}
    }
    ```

    glassfish-web.xml
    ```
    <!DOCTYPE glassfish-web-app PUBLIC "-//GlassFish.org//DTD GlassFish Application Server 3.1 Servlet 3.0//EN" "http://glassfish.org/dtds/glassfish-web-app_3_0-1.dtd">
    <glassfish-web-app>
        <parameter-encoding default-charset="iso-8859-15" />
    </glassfish-web-app>
    ```

2) Post HTTP Request with strings that do not exist in iso-8859.  
  * GF 6.2.1 (and 5.1.0)
    ```
    $ curl -X POST -H 'Content-Type: text/plain; charset=UTF-8' -d "〇△□" http://localhost:8080/Test/TestServlet
    text/plain;charset=iso-8859-15
    ãâ³â¡
    ```
  * ※GF 4.1.2
    ```
    # curl -X POST -H 'Content-Type: text/plain; charset=UTF-8' -d "〇△□" http://localhost:8080/Test/TestServlet
    text/plain; charset=UTF-8
    〇△□
    ```

## Impact of Issue
Incompatible behavior that causes applications to fail.

hs536:
Fix: #23580

### DIFF
--- a/appserver/web/web-core/src/main/java/org/apache/catalina/connector/Request.java
+++ b/appserver/web/web-core/src/main/java/org/apache/catalina/connector/Request.java
@@ -704,17 +704,20 @@ public class Request
             if (p != null) {
                 hostValve = p.getBasic();
             }
-            try {
-                String reqEncoding = this.servletContext.getRequestCharacterEncoding();
-                if (reqEncoding != null) {
-                    setCharacterEncoding(reqEncoding);
+            String charsetEncoding = getCharacterEncoding();
+            if (charsetEncoding == null) {
+                try {
+                    String reqEncoding = this.servletContext.getRequestCharacterEncoding();
+                    if (reqEncoding != null) {
+                        setCharacterEncoding(reqEncoding);
+                    }
+                    String resEncoding = this.servletContext.getResponseCharacterEncoding();
+                    if (resEncoding != null) {
+                        getResponse().getResponse().setCharacterEncoding(resEncoding);
+                    }
+                } catch(UnsupportedEncodingException e) {
+                    throw new RuntimeException(e);
                 }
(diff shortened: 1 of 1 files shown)


hs536:
failed to modify the signature of the commit, so I'll close it.

PR #23587, opened 2021-09-06
Prepare next development cycle for 6.2.2-SNAPSHOT - fixed update_version

dmatej:


dmatej:
Tested locally, the version in server.log is correct now

arjantijms (approved):

PR #23591, opened 2021-09-07
Fix #23590, Fix typo.

kaido207:
Signed-off-by: kaido207 <kaido.hiroki@fujitsu.com>

Fix #23590, Fix typo.
`log.fileCouldntFound` -> `log.fileCouldntbeFound`

PR #23592, opened 2021-09-07
Extract and test method to construct message

pzygielo:
Follow-up to https://github.com/eclipse-ee4j/glassfish/pull/23591#issuecomment-914138130


arjantijms:
Thanks! Btw, at some point we may want to remove i18n for the log files completely and just have all messages directly where they are consumed.

arjantijms (approved):

PR #23605, opened 2021-09-14
Explicit level for echo task

pzygielo:
This does not seem to be needed of WARNING level.

Sadly, it saves only one warning.


arjantijms (approved):
Thanks, much better, if only 1 message.

PR #23621, opened 2021-09-24
Issue 23616 Fixed generators on jdk11 + tests

### LINKED ISSUE #23616: Some usages of generateAndLoad in GlassFish fail on JDK11+
## Problem Description

When I wrote several tests to reproduce #23606 , I found another bug - the way how EjbUtils generate classes+interfaces breaks JDK+ rules.

After my tests failed, I migrated them back to Payara5, which still uses JDK8. Tests passed (except one, which fails due to leaky method API, which test misuses). But in GlassFish on JDK11/16/17 tests failed.

The cause: 

- PFL used the Unsafe class in JDK8, which generates classes in native code and doesn't do any validation.
- on JDK11+ PFL uses MethodHandles instead of Unsafe. This impl does the following check while Unsafe accepted anything:
```
            if (!pn.equals(lookupClass.getPackageName())) {
                throw new IllegalArgumentException("Class not in same package as lookup class");
            }
```
- the real cause is that some of generators generate classes with incorrect package name.
- see also: https://github.com/eclipse-ee4j/orb-gmbal-pfl/commit/7508c568e77b2660f98a95a6f89b6e7b4a379a78

## Suggested Solution

- [x] Do all necessary validations before we try generate and register generated class.
- [x] Use valid names

## Environment Details
* GlassFish Version (and build number): commit 178af6400c343de
* JDK version: OpenJDK 17+35-2724, AdoptOpenJDK-16.0.1+9, Zulu11.48+21-CA for GF, Zulu 8.54.0.21-CA for Payara
* OS: Kubuntu 21.04

dmatej:
Originally PFL used Unsafe class to generate classes. Since JDK11 it uses MethodInvocations, which verifies that the generated class has the same package as the anchor class. Generators were breaking the rule and under some circumstances it caused exceptions.

This PR also includes several tests and new test dependency: OpenJDK JMH. It's license is [here](https://github.com/openjdk/jmh/blob/master/LICENSE).

### DIFF
--- a/appserver/ejb/ejb-container/pom.xml
+++ b/appserver/ejb/ejb-container/pom.xml
@@ -222,6 +222,18 @@
             <groupId>org.hamcrest</groupId>
             <artifactId>hamcrest</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-core</artifactId>
+            <version>1.32</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-generator-annprocess</artifactId>
+            <version>1.32</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/appserver/ejb/ejb-container/pom.xml
+++ b/appserver/ejb/ejb-container/pom.xml
@@ -225,14 +225,10 @@
         <dependency>
             <groupId>org.openjdk.jmh</groupId>
             <artifactId>jmh-core</artifactId>
-            <version>1.32</version>
-            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.openjdk.jmh</groupId>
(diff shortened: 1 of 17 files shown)


dmatej:
Current status: 4 tests of the TCK fail with StackOverflow when JDK does some class loading. 
Hypothesis: Perhaps caused by a cyclic dependency between classloaders of a generated interface and generated proxy for a method of the interface.

dmatej:
I was wrong, the StackOverflow was caused by this property used in ts.override.properties:
`-Djava.protocol.handler.pkgs=javax.net.ssl`
When I removed it, tests run (didn't finish yet).

dmatej:
Ok, we are back, after I fixed my TCK test configuration, all failing tests passed, ufff:
`Not passed: 0/20 /media/data-samsung/data/dmatej/work/repo/git/jakartaeetck-runner/cts_home/jakartaeetck-report/ejb_ee_bb_localaccess/text/summary.txt`

arjantijms:
Note: see targeted version: 6.2.3, which means don't merge until 6.2.2 is released.

arjantijms (approved):

PR #23636, opened 2021-09-29
Fix #23634 Change @ExecuteOn targets for some of security commands.

### LINKED ISSUE #23634: "create-file-user --target server" fails if other instances is running.
## Environment Details
* GlassFish Version (and build number): GlassFish 6.2.1
* OS: CentOS Linux release 7
* Database: none

----------

## Problem Description
"create-file-user --target server" fails if other instances is running.
And the message "Configuration for target server not found." is displayed.

Output
```
[root /]# ./glassfish6/glassfish/bin/asadmin  create-file-user
(...)
remote failure: An error occurred during replication
Failure: Command create-file-user failed on server instance [Instance name]: remote failure: Configuration for target server not found.
Command create-file-user failed.
```

## Steps to reproduce
1. create and start other server instance.
2. execute `asadmin create-file-user`
  (default target value is `server`)

## Impact of Issue
If more than one instance is running, file-user cannot be created for DAS.

kaido207:
Signed-off-by: kaido207 <kaido.hiroki@fujitsu.com>

Fix #23634.  
If set `@ExecuteOn` to `RuntimeType.ALL`, some of security commands will also be executed on other instances when it is executed on DAS, which causes the commands to fail.

### DIFF
--- a/nucleus/security/core/src/main/java/com/sun/enterprise/security/cli/CreateFileUser.java
+++ b/nucleus/security/core/src/main/java/com/sun/enterprise/security/cli/CreateFileUser.java
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 1997, 2021 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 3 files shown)


arjantijms (approved):

arjantijms:
LGTM!

PR #23638, opened 2021-09-30
Fix #23637 Memory leak when redeploy same app repeatedly

### LINKED ISSUE #23637: Memory leak when redeploy same app repeatedly.
## Environment Details
* GlassFish Version (and build number): Eclipse GlassFish  6.2.1  (build master-b577-g50d68a3 2021-09-01T02:07:52+0000)
* JDK version: OpenJDK 64-Bit Server VM Temurin-11.0.12+7
* OS: CentOS Linux release 7.9.2009 (Core)
* Database:

----------

## Problem Description
Memory leak occured when redeploy same app repeatedly.  
Heap dumps captured by jmap are shown as below:
* Before redeploy
image
* redeploy same app 100 times
image
* redeploy same app 200 times
image

## Steps to reproduce
1. start-domain
2. execute following command repeatedly.
```
# asadmin deploy --force=true SampleAPP.war
```
## Impact of Issue
memory leak occured

11rx4f:
Fix: 23637

If the \<application\> DOM object is nested, only the parent element is removed, but the child elements leak when the application is undeployed.
Fix it to remove recursively.

QuickLook has passed in local.

### DIFF
--- a/nucleus/hk2/hk2-config/src/main/java/org/jvnet/hk2/config/Dom.java
+++ b/nucleus/hk2/hk2-config/src/main/java/org/jvnet/hk2/config/Dom.java
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2007, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
  *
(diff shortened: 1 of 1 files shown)


arjantijms (approved):
LGTM

PR #23675, opened 2021-10-23
Issue #23042 Fixed jdbc pools js error

### LINKED ISSUE #23042: TypeError: table.getAllSelectedRowsCount is not a function
## Environment Details
* GlassFish Version 5.1.0 (default private):
* JDK version: 1.8.0_202 (64bit)
* OS: Windows 7 (64bit)
* Database: n/a

----------

## Problem Description
In the GF Admin Console web app, Navigate to Resource/JDBC/JDBC ConnectionPools.  The table of current connection pools is rendered then the error is thrown, and rendered to the GUI client:
`TypeError: table.getAllSelectedRowsCount is not a function`.

Now, ignore the error and add a new JDBC Connection.  Complete the setup of the JDBC Connection, and the Admin GUI returns you back to the JDBC Connection Pools table where the new entry does not appear, and an hour glass is show indefinitely along with the same error message.  

No exception is logged to the Glassfish Console / Logs

It may be worth noting that the CSS is very slow to render and sometimes does not render at all for the admin pages.

## Steps to reproduce
See _Problem Description_

## Impact of Issue
Cannot effectively setup a new application domain under Glassfish 5.1.0 with JDK 1.8.0_202

image

dmatej:
- guiTemplateOnLoad used tables.js before it was loaded. It seems it wasn't
  even required as everything is set correctly by default.
- the code was removed also from other files where I have found same expectation

### DIFF
--- a/appserver/admingui/common/src/main/resources/js/adminjsf.js
+++ b/appserver/admingui/common/src/main/resources/js/adminjsf.js
@@ -2093,19 +2093,15 @@ admingui.table = {
     },
 
     changeButtons : function (buttons,tableId){
-        try {
-            var table = document.getElementById(tableId);// + ":_table");
-            var selections = table.getAllSelectedRowsCount();
-            var disabled = (selections > 0) ? false : true;
-            for (count=0; count < buttons.length; count++) {
-                var element = document.getElementById(buttons[count]);
-                if (element) {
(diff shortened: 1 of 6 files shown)


hs536 (approved):
LGTM

PR #23694, opened 2021-11-02
#23692: improve build time by avoiding unnecessarry resources filtering

### LINKED ISSUE #23692: Improve build time (fix regression)
running `mvn -U -C clean install -Pstaging -Dcheckstyle.skip=true` on my box used to take <30mins some time ago. The build of current master (204d3af):

```
...
[INFO] GlassFish OSGi Console Plugin ...................... SUCCESS [ 14.059 s]
[INFO] GlassFish Flashlight Client ........................ SUCCESS [ 12.284 s]
[INFO] Flashlight ......................................... SUCCESS [  0.049 s]
[INFO] Ant tasks .......................................... SUCCESS [ 12.109 s]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:28 h
[INFO] Finished at: 2021-11-01T13:08:55+01:00
```

re-running the build after few changes:

```
[INFO] Glassfish Distribution ............................. SUCCESS [ 41.555 s]
[INFO] Security Related Implementation for GlassFish ...... SUCCESS [  0.347 s]
[INFO] GlassFish Tests .................................... SUCCESS [  0.149 s]
[INFO] GlassFish Tests - TCK .............................. SUCCESS [  0.149 s]
[INFO] GlassFish Tests - TCK - Authorization .............. FAILURE [01:45 min]
[INFO] GlassFish Tests - TCK - Embedded EJB - Smoke ....... SKIPPED
[INFO] GlassFish JavaTest install ......................... SKIPPED
[INFO] EJB Lite Basic ..................................... SKIPPED
[INFO] GlassFish Tests - TCK - Run embedded EJB smoke ..... SKIPPED
[INFO] Console extensions for OSGi platforms .............. SKIPPED
[INFO] Felix Web Console Extension for GlassFish .......... SKIPPED
[INFO] GlassFish OSGi Console Plugin ...................... SKIPPED
[INFO] GlassFish Flashlight Client ........................ SKIPPED
[INFO] Flashlight ......................................... SKIPPED
[INFO] Ant tasks .......................................... SKIPPED
[INFO] BUILD FAILURE
[INFO] Total time:  02:13 h
[INFO] Finished at: 2021-11-01T17:14:03+01:00
```

Seriously? Does one have to wait for hours to find out he has done something wrong somewhere?

https://ci.eclipse.org/glassfish/view/GlassFish/job/glassfish_weekly_build/buildTimeTrend shows that I'm not alone who sees the significant slowdown of the build - it went from 10mins to 20mins

lukasj:
having legal resources in an extra artifact improves the speed of the build by not filtering source tree in every module

Relates to #23692

### DIFF
--- a/appserver/distributions/pom.xml
+++ b/appserver/distributions/pom.xml
@@ -172,6 +172,7 @@
                         <attach>false</attach>
                         <appendAssemblyId>false</appendAssemblyId>
                         <useProjectArtifact>false</useProjectArtifact>
+                        <ignoreMissingDescriptor>true</ignoreMissingDescriptor>
                     </configuration>
                 </plugin>
                 <plugin>
--- a/appserver/distributions/pom.xml
+++ b/appserver/distributions/pom.xml
@@ -154,25 +154,15 @@
                 </plugin>
                 <plugin>
                     <artifactId>maven-assembly-plugin</artifactId>
-                    <executions>
-                        <execution>
-                            <id>default-single</id>
-                            <phase>process-resources</phase>
-                            <goals>
-                                <goal>single</goal>
-                            </goals>
-                        </execution>
-                    </executions>
                     <configuration>
                         <descriptors>
-                            <descriptor>${basedir}/src/main/assembly/${project.artifactId}.xml</descriptor>
+                            <descriptor>src/main/assembly/${project.artifactId}.xml</descriptor>
                         </descriptors>
                         <ignoreMissingDescriptor>false</ignoreMissingDescriptor>
                         <finalName>${stage.dir.name}</finalName>
                         <attach>false</attach>
                         <appendAssemblyId>false</appendAssemblyId>
                         <useProjectArtifact>false</useProjectArtifact>
-                        <ignoreMissingDescriptor>true</ignoreMissingDescriptor>
                     </configuration>
                 </plugin>
                 <plugin>
--- a/nucleus/distributions/pom.xml
+++ b/nucleus/distributions/pom.xml
@@ -130,25 +130,15 @@
 
                 <plugin>
                     <artifactId>maven-assembly-plugin</artifactId>
-                    <executions>
-                        <execution>
-                            <id>default-single</id>
-                            <phase>process-resources</phase>
-                            <goals>
-                                <goal>single</goal>
-                            </goals>
-                        </execution>
-                    </executions>
                     <configuration>
                         <descriptors>
-                            <descriptor>${basedir}/src/main/assembly/${project.artifactId}.xml</descriptor>
+                            <descriptor>src/main/assembly/${project.artifactId}.xml</descriptor>
                         </descriptors>
                         <ignoreMissingDescriptor>false</ignoreMissingDescriptor>
                         <finalName>${stage.dir.name}</finalName>
                         <attach>false</attach>
                         <appendAssemblyId>false</appendAssemblyId>
                         <useProjectArtifact>false</useProjectArtifact>
-                        <ignoreMissingDescriptor>true</ignoreMissingDescriptor>
                     </configuration>
                 </plugin>
 
--- a/nucleus/distributions/pom.xml
+++ b/nucleus/distributions/pom.xml
@@ -148,6 +148,7 @@
                         <attach>false</attach>
                         <appendAssemblyId>false</appendAssemblyId>
                         <useProjectArtifact>false</useProjectArtifact>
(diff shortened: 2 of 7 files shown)


nucleus/distributions/pom.xml
```diff
@@ -148,6 +148,7 @@
                         <attach>false</attach>
                         <appendAssemblyId>false</appendAssemblyId>
                         <useProjectArtifact>false</useProjectArtifact>
+                        <ignoreMissingDescriptor>true</ignoreMissingDescriptor>
```
dmatej, line 151: The same property is set to false here. Also I believe false is correct - if the descriptor is missing, the build is broken.

nucleus/parent/pom.xml
```diff
@@ -928,11 +930,7 @@
                         <configuration>
                             <resources>
                                 <resource>
-                                    <directory>${maven.multiModuleProjectDirectory}</directory>
```
dmatej, line 931: It worked, because the maven.multiModuleProjectDirectory pointed to the root of the project. But it was rather emergency temporal solution. This is better.

nucleus/parent/pom.xml
```diff
@@ -962,6 +960,51 @@
...
+                        <configuration>
+                            <descriptors>
+                                <descriptor>src/main/assembly/resources.xml</descriptor>
+                            </descriptors>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
```
dmatej, line 983: groupId is redundant (implicit in both cases)

appserver/distributions/pom.xml
```diff
@@ -172,6 +172,7 @@
                         <attach>false</attach>
                         <appendAssemblyId>false</appendAssemblyId>
                         <useProjectArtifact>false</useProjectArtifact>
+                        <ignoreMissingDescriptor>true</ignoreMissingDescriptor>
```
dmatej, line 175: dtto, ignoreMissingDescriptor is already set to false, which I believe is correct

dmatej (changes requested):
It worked because the assembly plugin is just in plugin management, while you added it to the implicit lifecycle, just without the not inherited common-resources execution (now it executes default execution from plugin management even where it wasn't desired).

It is really interesting, I had to compare both builds :)

dmatej:
Funny, I would think that this is Maven's bug - even when I add <inherit>false</inherit> both to plugin and your new execution, it is always executed when it is mentioned just in pluginManagement. Evil :D

lukasj:
>Funny, I would think that this is Maven's bug - even when I add false both to plugin and your new execution, it is always executed when it is mentioned just in pluginManagement. Evil :D

No, it is not a bug. It is caused by the fact that the execution id defined in the plugin management is `default-single` - that implies that it gets executed every time. Changing the id to something else, probably with related refactoring, should help as well

nucleus/parent/pom.xml
lukasj: that does not mean one cannot use it ;-)

nucleus/parent/pom.xml
lukasj: there was a discussion about that property, see: https://github.com/eclipse-ee4j/eclipselink/pull/1244#discussion_r693939448

dmatej:
Without your changes it wasn't executed - and it shouldn't be executed here. Plugin management touches just configured plugins in the lifecycle and there wasn't any such plugin. You added it just to the nucleus-parent and anything from there should not be inherited, but it is. The plugin then takes configuration from the plugin management.

dmatej:
Ok, nice, nucleus-common used wrong (meaningless) id, now it starts making sense ... I hope we are on the good path :)

dmatej:
Wow. Finally it is not a bug but feature ... maven assembly plugin simply brings default-single by default if there is no execution active, but the plugin is there (but still ... if said inhertited=false, I would expect that even that plugin will not be active). I am curious how far we can get when this will join #23696 

`Total time:  01:36 min (Wall Clock)`

dmatej:
I have created PR for your PR (I didn't want to aggresively push here, just take a look for sure):
https://github.com/lukasj/glassfish-1/pull/1

nucleus/parent/pom.xml
dmatej: It is still somehow nasty, but less nasty than before ;-)

dmatej (approved):

PR #23714, opened 2021-11-21
Update maven plugins for JDK 18 build compatibility

arjantijms:


appserver/persistence/cmp/support-sqlstore/pom.xml
```diff
@@ -117,7 +117,7 @@
         </resources>
         <plugins>
             <plugin>
-                <groupId>org.codehaus.mojo</groupId>
+                <groupId>org.omnifaces</groupId>
```
dmatej, line 120: Is this completely new plugin or is it based on codehause's dead or is it based no this one?
https://www.antlr.org/api/maven-plugin/latest/plugin-info.html

dmatej (approved):

appserver/admin/admin-core/pom.xml
```diff
@@ -23,7 +23,7 @@
     <parent>
         <groupId>org.glassfish.main.admin</groupId>
         <artifactId>admin</artifactId>
-        <version>6.2.3-SNAPSHOT</version>
+        <version>6.2.4</version>
```
dmatej, line 26: Shouldn't it be rather 6.2.4-SNAPSHOT?

dmatej (changes requested):

appserver/admin/admin-core/pom.xml
dmatej: ... or 7.0.0-SNAPSHOT?

pzygielo: ... or do not change this in this very PR?

dmatej: > ... or do not change this in this very PR?

I expect it will be rebased after the release, so these changes will go away, but yeah, you are right.

jGauravGupta (approved):

appserver/persistence/cmp/support-sqlstore/pom.xml
arjantijms: It's the codehouse plug-in, updated to all the latest versions of everything, and with the correct settings to not use the security manager to prevent system.exit, but instead throw an exception.

arjantijms:
Already merged for 6.x, will cherry pick 6.x commits/pr to master instead.

PR #23737, opened 2021-12-10
Jersey ejb component provider for 6.x

dmatej:
This PR ...
- reduces cyclic dependencies between GF and Jersey (removes dependency on jersey-gf-ejb which depended on GF)
- reduces differences between master and 6.x


arjantijms (approved):
LGTM

PR #23781, opened 2022-02-06
For tags - build all dependencies of impl

pzygielo:


pzygielo:
_EJB tests seem flaky_ :disappointed:

arjantijms:
I've already staged the dependencies for Tags, which will be a more stable solution to incorporate these. We also should move the Tags implementation away from the API project so that it will always build from the root.

PR #23794, opened 2022-02-08
#23507 Removed console devtests

dmatej:
- unused and unusable with current firefox versions
- obsoleted, used deprecated methods of old Selenium version,  which were removed in newer releases
- poor design, but good coverage
- Admin UI will have to change in GF7 so it is not worth to fix this


arjantijms (approved):
No point in keeping those around

PR #23807, opened 2022-02-15
MicroProfile Config 3.0 Support

MattGill98:
This change supports the deployment of MicroProfile Config 3.0 compatible applications, and modifies GF full profile to pass the MP Config 3.0.1 TCK by using the Smallrye Config 3.0.0.RC1 implementation.

The Smallrye dependencies are currently inlined in the implementation module, with the `org.glassfish.microprofile.config` and `io.smallrye.config` packages exported. If further Smallrye dependencies are used in future, it may be helpful to export the common utilities to their own module to reduce bundle size. I've inlined the dependencies to improve performance rather than loading nested JARs, although I'm happy to change this if required.

A sniffer module has been created to enable the respective MP implementations when they are detected, although the module defers the specific deployment steps to Smallrye.

The TCK needed a few modifications which are detailed in the commit messages:
- The TCK assumes that empty beans.xml files imply the `all` bean discovery mode. Since CDI 4.0 this is no longer the case, so all empty beans.xml files are replaced with a correct implementation by an Arquillian extension.
- The TCK depends on some system properties, which an initial custom test is used to configure in a container neutral way.
- The official Arquillian GlassFish connector throws all exceptions as `GlassFishClientException`s, which fails the deployment assertions. An Arquillian extension is used to cast these exceptions when the message is correct.
- Hamcrest is required by some deployments, but not included. The Arquillian extension handles this.

### DIFF
--- a/appserver/featuresets/glassfish/pom.xml
+++ b/appserver/featuresets/glassfish/pom.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-
+    Copyright (c) 2022 Contributors to Eclipse Foundation.
     Copyright (c) 2015, 2022 Oracle and/or its affiliates. All rights reserved.
 
     This program and the accompanying materials are made available under the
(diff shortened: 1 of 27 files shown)


dmatej:
Wow, that is a nice work! I will first fix my PR and then I will try to help you to fix them (if it would be needed) ;)

arjantijms:
Great work indeed @MattGill98!

Maybe though take into consideration alignment with Piranha Cloud:

* https://github.com/piranhacloud/piranha/pull/2391/files
* https://github.com/omnifaces/jwt
* https://github.com/omnifaces/omniconfig

OmniConfig packages MP Config from Helidon, while JWT is the one I coded at Payara, somewhat repackaged to be reusable (it works in Tomcat too etc when the right dependencies are present).

That said, regarding things like Tracing, SmallRye is already reusable out of the box. For Helidon some work may be needed for that. Piranha Cloud did use SmallRye for MP Config before.

jansupol:
Added [CQ for Helidon Config 3.0.0-M1](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23994)
Added [CQ for Smallrye Config 3.0.0-RC1](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23996)
Glassfish can piggybag.

jansupol:
Helidon Config requires JDK 17, Smallrye JDK 8.

arjantijms:
@jansupol 
> Glassfish can piggybag.

Thanks!

Btw, how do you keep track of which dependency has been CQ'ed and which one has not been? Or is it easier to just CQ everything and if it already has been CQ'ed ipzilla will let you know?

Is there a tool available that simply CQ's every dependency from a project, or do you manually create the bugs for each  one of them?

jansupol:
They have a query https://dev.eclipse.org/ipzilla/query.cgi for the CQs.

jansupol:
I try to CQ everything I cannot find the CQ for. There is no tool I am aware of, some questions are complicated, the repository of the source code, all the licenses, description of the dependency, project home page, this information can be difficult to get automatically, I grab them from multiple sources. It is easier when a newer version is CQed, most of the information can be used from the previous CQ...

MattGill98:
I've swapped out Smallrye Config for Helidon config - had to make a few small changes (mostly relating to classloading) but in general very little needed changing

What else needs completing for this change to make it in?

arjantijms (approved):

dmatej:
Wait for a while with the merge, my yesterday PR broke something - servlet TCK doesn't pass at this moment.

dmatej:
Wait for a while with the merge, my yesterday PR broke something - servlet TCK doesn't pass at this moment.

dmatej (approved):

PR #23808, opened 2022-02-18
Integrate Expressly

arjantijms:
Uses a snapshot for now since we're not able to stage doing to a
Sonatype access issue.


snapshots/pom.xml
```diff
@@ -34,6 +34,7 @@
 
     <modules>
         <module>jbatch</module>
+        <module>expressly</module>
```
pzygielo, line 37: Could the execution at https://github.com/eclipse-ee4j/glassfish/blob/8796647bb57e70610b79bc85f4a0d5f35dbbe186/snapshots/pom.xml#L58 be renamed?

dmatej: Mea culpa :)

dmatej (approved):

PR #23823, opened 2022-02-21
Integrate Faces 4.0.0-M3

arjantijms:

### DIFF
--- a/appserver/pom.xml
+++ b/appserver/pom.xml
@@ -80,8 +80,8 @@
         <!-- Jakarta API Versions -->
 
         <!-- Jakarta Faces -->
-        <jakarta.faces-api.version>4.0.0-M2</jakarta.faces-api.version>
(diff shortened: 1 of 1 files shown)


dmatej (approved):

hs536 (approved):

dmatej:
Breaks CDI for some reason, Arjan knows about it.

arjantijms:
> Breaks CDI for some reason, Arjan knows about it.

Yes, I/we made a mistake in the initialiser code, which is a somewhat confusing thing where annotations that Mojarra uses to determine whether it needs to activate itself are mixed with the set of all annotations that Mojarra uses. Among them is "@Resource", which is used by Mojarra but should not activate it. All the tests that fail contain "@Resource" annotations without any CDI activation.

As Faces needs CDI to be present, this it what fails.

PR #23843, opened 2022-03-04
Merge 6.x to master

dmatej:

PR #23845, opened 2022-03-05
Integrate Jersey 3.1.0-M3

arjantijms:

### DIFF
--- a/appserver/featuresets/debug/pom.xml
+++ b/appserver/featuresets/debug/pom.xml
@@ -46,6 +46,24 @@
 
     <dependencies>
         <dependency>
+            <groupId>org.osgi</groupId>
+            <artifactId>org.osgi.service.resolver</artifactId>
+            <version>1.1.1</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.felix</groupId>
+            <artifactId>org.apache.felix.resolver</artifactId>
+            <version>2.0.4</version>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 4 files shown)


hs536 (approved):

hs536:
a little bit concerned about Jersey still depends on JAXB3:  
https://github.com/eclipse-ee4j/jersey/blob/3.1.0-M3/pom.xml#L2196-L2197

arjantijms:
> https://github.com/eclipse-ee4j/jersey/blob/3.1.0-M3/pom.xml#L2196-L2197

Good catch, @jansupol any idea why that is?

dmatej:
Just FYI: https://github.com/eclipse-ee4j/jersey/issues/5006
So if you would be able to work on jackson, we could probably meet on some stable point of compatibility of 4 modules: jersey, grizzly, metro-wsit and jackson ... and finally GlassFish, which is waiting for that here https://github.com/eclipse-ee4j/glassfish/pull/23843 - tests detected incompatibility :-)

jansupol:
[JAX-B version update](https://github.com/eclipse-ee4j/jersey/pull/5008) is merged. For Jackson, we need to have both JAX-B/3 and JAX-B/4 on [classpath](https://github.com/eclipse-ee4j/jersey/blob/3.1/tests/osgi/functional/src/test/java/org/glassfish/jersey/osgi/test/basic/JsonJacksonTest.java#L45). Looks like OSGi is ok with that.

PR #23930, opened 2022-05-04
Upgrade jbatch to v2.1.0

scottkurz:
Signed-off-by: Scott Kurz <skurz@us.ibm.com>

There shouldn't be any functional change here.   We had just been waiting for Weld & the CDI API dependencies to release final versions, and now that they have, we've released 2.1.0.


arjantijms (approved):
Thanks for the PR!

PR #23934, opened 2022-05-11
Fix #23933 by adding a proxy for the Policy when Security Managed used

### LINKED ISSUE #23933: GlassFish 7 has 40 failures running platform authorization tests
See https://gist.github.com/scottmarlow/67a2c64b88f8453a622931235c12ad4e

The issue is the combination of a security manager and the fact jakarta authorization uses a policy. As soon as the policy itself needs any kind of Java SE permission, the security manager will ask the policy, which is itself. This then leads to recursion.

Ultimately the design of mixing Java SE code level permissions and permissions wrt the logged-in user are wrong. We'll address this in EE 11. For now we can install a proxy that doesn't send Java SE permissions to the Jakarta Authorisation Policy.

Note that the standalone TCK passes sinds a) it doesn't use the security manager, and b) uses wars, which seemly just happen to not ask for a certain class loading permission.

CC @scottmarlow

arjantijms:
Fixes #23933

### DIFF
--- a/nucleus/security/core/pom.xml
+++ b/nucleus/security/core/pom.xml
@@ -103,10 +103,10 @@
             <groupId>org.glassfish.gmbal</groupId>
             <artifactId>gmbal</artifactId>
         </dependency>
-		<dependency>
-			<groupId>org.javassist</groupId>
-			<artifactId>javassist</artifactId>
-		</dependency>
+        <dependency>
+            <groupId>org.javassist</groupId>
+            <artifactId>javassist</artifactId>
+        </dependency>
         
         <dependency>
             <groupId>org.glassfish.annotations</groupId>
--- a/nucleus/security/core/pom.xml
+++ b/nucleus/security/core/pom.xml
@@ -103,6 +103,11 @@
             <groupId>org.glassfish.gmbal</groupId>
             <artifactId>gmbal</artifactId>
         </dependency>
+		<dependency>
+			<groupId>org.javassist</groupId>
+			<artifactId>javassist</artifactId>
+		</dependency>
+        
         <dependency>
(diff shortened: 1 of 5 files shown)


scottmarlow (approved):

nucleus/security/core/pom.xml
```diff
@@ -103,6 +103,11 @@
             <groupId>org.glassfish.gmbal</groupId>
             <artifactId>gmbal</artifactId>
         </dependency>
+		<dependency>
```
dmatej, line 106: Formatting ;)

dmatej (approved):

nucleus/security/core/src/main/java/com/sun/enterprise/security/PolicyLoader.java
```diff
@@ -143,15 +160,92 @@ public void loadPolicy() {
...
+                clazz.toClass(
+                    Thread.currentThread()
+                          .getContextClassLoader()
+                          .loadClass(System.getProperty(POLICY_CONF_FACTORY))));
 
         if (!(javaPolicyInstance instanceof Policy)) {
             throw new RuntimeException(SM.getString("enterprise.security.plcyload.not14"));
         }
 
+        javaPolicyInstance.toString();
```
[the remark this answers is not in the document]
arjantijms: It was to make sure the Proxy absolutely creates the underlying class.

nucleus/security/core/pom.xml
arjantijms: Maybe we need to install a check style rule for tabs in pom.xml files? Somehow a tab got inserted again :O

PR #23939, opened 2022-05-18
Omnifish

dmatej:
* Using OmniFish maven repository instead of building own snapshots
* Added note about OmniFish support to the list


dmatej:
The previous build failed because Felix could not handle it's own version with classifier. That is fixed here and since nobody uses the M1 version, I will redeploy it to include the fix.
https://github.com/apache/felix-dev/pull/152

arjantijms (approved):

smillidge:
I don't think this should be merged. If we need to host patched 3rd party content I believe it should use some Eclipse mechanism. Please revert while I seek clarification.

dmatej:
> I believe it should use some Eclipse mechanism

I would really appreciate if you would be more concrete and if you could lead us to that mechanism, because we didn't find that. Until now each and every build had to rebuild the patch from someone's repository and use it. But those branches were sometimes deleted by mistake, sometimes updated, it was quite risky. So now we have at least publicly available repository, provided by JFrog cloud and maintained by company owned by GlassFish developers.

**But it would be really nice if you could discuss another perhaps ideal solution with someone from Eclipse.**

However these patches are just temporary, GF7 will not be released with any staging/snapshot/nonstandard dependency. We also did not find any Eclipse rule which could be broken by this change. See also: https://www.eclipse.org/projects/handbook/#resources-external

This PR ensures that

1. Snapshot builds are repeatable
2. Builds need less time
3. We can continue with development without blocking

Revert would make GlassFish uncompilable as the fix was already merged and the patch branch deleted. You would have to rebuild it from Felix master (including other changes or from the exact commit). 

We are simply cooperating with other projects, however we can't wait until Apache releases another Felix and stop developing GlassFish until that.

PR #23941, opened 2022-05-20
Fix #23940 Users can override classes of the jakarta package.

### LINKED ISSUE #23940: Users can override classes of the jakarta package.
## Environment Details
* GlassFish Version (and build number):7.0.0-M4
----------

## Problem Description
GlassFish defines the following list as exceptions to classloader delegation, but this list does not have "jakarta", which allows users to override classes of the jakarta package.

https://github.com/eclipse-ee4j/glassfish/blob/7.0.0-M4/appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java#L116-L126

kosakak:
Signed-off-by: kosaka.koki <kosaka.koki@fujitsu.com>

- Fixes #23940


dmatej (approved):
Wow, nice! Thank you!

PR #23945, opened 2022-05-22
Add services file to indicate which BuildServices impl we want.

arjantijms:

### DIFF
--- a/appserver/web/weld-integration/pom.xml
+++ b/appserver/web/weld-integration/pom.xml
@@ -173,9 +173,7 @@
             <resource>
                 <directory>src/main/resources</directory>
                 <includes>
-                    <include>**/faces-config.xml</include>
-                    <include>**/com.sun.faces.spi.FacesConfigResourceProvider</include>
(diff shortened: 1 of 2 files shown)


dmatej (approved):

appserver/web/weld-integration/pom.xml
```diff
@@ -173,9 +173,7 @@
             <resource>
                 <directory>src/main/resources</directory>
                 <includes>
-                    <include>**/faces-config.xml</include>
-                    <include>**/com.sun.faces.spi.FacesConfigResourceProvider</include>
-                    <include>**/jakarta.enterprise.inject.spi.Extension</include>
+                    <include>**/*</include>
```
pzygielo, line 176: I think
```xml
          <resource>
              <directory>src/main/resources</directory>
          </resource>
```
(w/o explicit `include`-everything) would do.

arjantijms: I'll try that, thanks!

PR #23948, opened 2022-05-26
Fix of default context root with a generated suffix.

OndroMih:
This is caused by commit 8262ac43857d14c6186bd27009517a977df2073e which saves the uploaded deployment file to a temporary file, which automatically generates a suffix.

This also fixes Jakarta Security TCK tests that are failing with GlassFish 7.0.0-M5


dmatej (approved):

dmatej:
Yeah, I remember that - my fix fixed glassfish's old tests (of the zombie army), and while I did not see any other failure, so I expected the context is resolved by some other mechanism, which was probably wrong expectation OR the mechanism is still broken.
But this fixes it. I'm not sure about other consequences, but as the code really prays for refactoring, we will revisit it later with more tests.

PR #23952, opened 2022-05-29
Fix build compatible extensions not running after portable ones.

arjantijms:


dmatej (approved):
Nice!

PR #23955, opened 2022-05-30
Using released Felix instead of temporary prerelease by OmniFish

dmatej:


dmatej:
Just FYI, as it is not critical for us now, we will wait until Felix 7.0.5 will make it to Maven Central, then I will remove the Apache repository reference too. We just wanted to see if it passes the build on Jenkins - if not, we would help Felix again without any need of revert :-)

arjantijms (approved):

PR #23957, opened 2022-05-31
Support scanning applibs

arjantijms:
Refactoring first:

* Renaming variables
* Formatting
* Simplifying logic
* Etc

### DIFF
--- a/nucleus/deployment/common/src/main/java/org/glassfish/deployment/common/InstalledLibrariesResolver.java
+++ b/nucleus/deployment/common/src/main/java/org/glassfish/deployment/common/InstalledLibrariesResolver.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2022, 2022 Contributors to the Eclipse Foundation.
  * Copyright (c) 1997, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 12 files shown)


arjantijms:
@anajosep what do you think? Is this a good approach? @senivam any opinion? This touches the deployer code.

gurunrao:
@arjantijms, we can add following details for wider review of the PR:
- changes proposed in this pull request(or issue (link) that is solved with pull request or link to mailing list discussion thread)
- changes needed or Describe the change
- PR tested(like CI link or logs)

dmatej (approved):
Just a question - is there some fix I did not notice or is it really just cleanup? (if not, the title is a bit misleading)

arjantijms:
The fix / change is in the second commit. It adds the libraries identified by extension-list to the collection of libraries returned.

The first commit is only cleanup.

arjantijms:
@gurunrao thanks for your suggestion, you are right. This fixes the CDI TCK test for EarInstalledLibs which has 3 variants. It's the test where the CDI TCK wants the integrator / test runner to put the ext-cdi-Lib in a server extension folder (in gf this is domains/domain1/Lib/applied). The test then checks that the beans inside this jar are scanned.

Ultimately the change is small (2nd commit). The first commit represents the code I had to read and study to ultimately be able to make that small change.

gurunrao (approved):

dmatej:
> The fix / change is in the second commit. It adds the libraries identified by extension-list to the collection of libraries returned.
> 
> The first commit is only cleanup.

Oh, yeah, mea culpa, I should take a look on the list of commits, now I understand. Yeah, seems logical to me.

nucleus/deployment/common/src/main/java/org/glassfish/deployment/common/InstalledLibrariesResolver.java
```diff
@@ -16,325 +17,298 @@
...
+     * @return a list of libraries included in the archive
      */
-    private static Vector getArchiveLibraries(ReadableArchive archive) {
-
+    private static List<String> getEmbeddedLibraries(ReadableArchive archive) {
         Enumeration<String> entries = archive.entries();
-        if (entries == null)
-            return null;
+        if (entries == null) {
+            emptyList();
```
pzygielo, line 201: This is unclear for me.
- What happens with that empty list?
- What happens in `while` below with `entries == null` then?

dmatej: Seems to me there is missing the `return` keyword. 
And I had the same mistake yesterday in my code, seems Eclipse editor sometimes does something ... not ideal :-D

arjantijms: Thanks so much for noticing, do I see a PR coming? ;)

PR #23965, opened 2022-06-02
Test bean attributes for session bean

arjantijms:
Install support for delegating some EJB tasks to the right bean archive.

Without this, when the a bean manager for the root bean archive is used,
it will not find the EJB definitions in a sub-archive, and will treat
the bean as a normal CDI bean.

For EJB beans a few special rules have to be taken into account, and
without applying these rules
CreateBeanAttributesTest#testBeanAttributesForSessionBean fails.

### DIFF
--- a/appserver/web/weld-integration/src/main/java/org/glassfish/weld/WeldDeployer.java
+++ b/appserver/web/weld-integration/src/main/java/org/glassfish/weld/WeldDeployer.java
@@ -34,7 +34,11 @@ import static org.jboss.weld.bootstrap.api.Environments.SERVLET;
 import static org.jboss.weld.bootstrap.spi.EEModuleDescriptor.ModuleType.CONNECTOR;
 import static org.jboss.weld.bootstrap.spi.EEModuleDescriptor.ModuleType.EJB_JAR;
 import static org.jboss.weld.bootstrap.spi.EEModuleDescriptor.ModuleType.WEB;
+import static org.jboss.weld.manager.BeanManagerLookupService.lookupBeanManager;
 
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -72,12 +76,15 @@ import org.glassfish.weld.services.SecurityServicesImpl;
 import org.glassfish.weld.services.TransactionServicesImpl;
 import org.glassfish.weld.util.Util;
 import org.jboss.weld.bootstrap.WeldBootstrap;
+import org.jboss.weld.bootstrap.api.ServiceRegistry;
 import org.jboss.weld.bootstrap.spi.BeanDeploymentArchive;
 import org.jboss.weld.bootstrap.spi.BeanDiscoveryMode;
 import org.jboss.weld.bootstrap.spi.EEModuleDescriptor;
 import org.jboss.weld.bootstrap.spi.helpers.EEModuleDescriptorImpl;
 import org.jboss.weld.ejb.spi.EjbServices;
 import org.jboss.weld.injection.spi.InjectionServices;
+import org.jboss.weld.manager.BeanManagerImpl;
+import org.jboss.weld.module.EjbSupport;
 import org.jboss.weld.resources.spi.ResourceLoader;
 import org.jboss.weld.security.spi.SecurityServices;
 import org.jboss.weld.serialization.spi.ProxyServices;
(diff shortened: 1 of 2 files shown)


arjantijms:
@senivam do you have any feedback on the PR?

arjantijms:
@russgold Any insight why this worked before, and why there's a proxy needed now?

hs536 (approved):
LGTM

senivam (approved):
for me, it is strange that suddenly it stopped working, but since the PR fixed the problem should be fine.

gurunrao (approved):

appserver/web/weld-integration/src/main/java/org/glassfish/weld/WeldDeployer.java
```diff
@@ -353,6 +360,71 @@ public String getName() {
...
+                    if (!deploymentImpl.getBeanDeploymentArchives().isEmpty()) {
+
+                        BeanDeploymentArchive rootArchive = deploymentImpl.getBeanDeploymentArchives().get(0);
+                        ServiceRegistry rootServices = bootstrap.getManager(rootArchive).getServices();
+
+                        EjbSupport originalEjbSupport = rootServices.get(EjbSupport.class);
+                        if (originalEjbSupport != null) {
+
+                            // We need to create a proxy instead of a simple wrapper, since EjbSupport
+                            // references the type "EnhancedAnnotatedType", which the Weld OSGi bundle doesn't
```
dmatej, line 379: So it is just a temporary fix until Weld fixes that, do they know?

dmatej (approved):

appserver/web/weld-integration/src/main/java/org/glassfish/weld/WeldDeployer.java
arjantijms: I'm not 100% sure if it's a Weld problem, might need to check how (if) Weld passes on other servers.

starksm64: The current WildFly 27.0.0.Alpha1 release being used for the core profile passes these tests in the web profile tck run.

arjantijms: Thanks, that's a pointer. It remains a mystery though why this old test passed for years on GF, but then with Weld 5 started to fail. Of course there's the option something else in GF is the root cause.

What happens is that the root bean archive doesn't have the EJB descriptors, but the one representing the war has. The bean manager injected in the test archive (which btw is crazy difficult to track where it exactly comes from) represents this root archive. The Weld bean manager doesn't try to find the bean archive that actually contains the EJB class, but directly looks in the current archive.

If that's a fault of Weld (it should lookup the correct archive) or some fault of GlassFish (the root archive should also reference the ejb descriptors) is hard to say, or perhaps the injected bean manager for the test should be the war one, and not the root one.

As a test, I tried giving the root archive the ejb descriptors as well, but Weld absolutely choked on that, and everything started to fail randomly.

starksm64: I'll ask @manovotn to comment as it seems like Weld did have a change relative to the archive identification, but it seems like it was for the build time extension handling.

manovotn: I cannot think of any Weld change that would cause this.
Is [this](testBeanAttributesForSessionBean) the test in question?

One thing I can see is that the test archive doesn't have any regular CDI beans, so I am not sure how that affects discovery for GF. Prior to CDI 4, this archive had `all` discovery whereas now it default to `annotated` (with 0 cdi beans). Could that make a difference for you? I.e. try changing the test in question to contain `beans.xml` with discovery `all` and re-run it @arjantijms

manovotn: Actually, there is at least one class with a `TundraStereotype ` stereotype on it. Still, the discovery is the only difference I can think of in context of this test.

arjantijms: I guess if nobody just knows what the reason could be, the only thing to do would be to do some extensive analysis of Weld 5 running on WildFly and finding out how it can pass there, and then running the previous TCK on GF 6 with Weld 4 and finding out what happens there and how it passed.

manovotn: Rather than extensive analysis, I'd first try to find out if the discovery causes that. GF might have different integration of EJB support altogether in which case any analysis of WFLY might not help here.

arjantijms: I think I tried that before. But the discovery doesn't seem to be the issue. The root archive by definition has no beans, and the war archive in question does (it has at least the EJB).

What now happens is this

Structure:

```
Weld BeanManager for root_CreateBeanAttributesTest63816be3596ce2e9f4b71ee22a457711159b93c
|
|—> services [1] —-> EjbSupport.class —> ejbDescriptors —> [empty]
|
Weld BeanManager for CreateBeanAttributesTest63816be3596ce2e9f4b71ee22a457711159b93c 
|
|—> sServices [2] —-> EjbSupport.class —> ejbDescriptors —> ejb bean class
```


Code:

```java
    @Override
    public <T> BeanAttributes<T> createBeanAttributes(AnnotatedType<T> type) {
        EnhancedAnnotatedType<T> clazz = services.get(ClassTransformer.class).getEnhancedAnnotatedType(type, getId());
        if (services.get(EjbSupport.class).isEjb(type.getJavaClass())) {
            return services.get(EjbSupport.class).createSessionBeanAttributes(clazz, this);
        }
        return BeanAttributesFactory.forBean(clazz, this);
    }
```

The test gets the bean manager for BDA "root_CreateBeanAttributesTest63816be3596ce2e9f4b71ee22a457711159b93c". Inside that bean manager it does `services.get(EjbSupport.class)`, which doesn't contain an `EJBSupport` instance which has the right EJB Descriptors. 

Some methods in the Weld bean manager do lookup the class in the correct bean archive, e.g.

```
@Override
    public <T> AnnotatedType<T> createAnnotatedType(Class<T> type) {
        // first, make sure to use the right BDA ID for this class
        String bdaId = BeanManagerLookupService.lookupBeanManager(type, this).getId();
        return getServices().get(ClassTransformer.class).getBackedAnnotatedType(type, bdaId);
    }
```

manovotn: > The test gets the bean manager for BDA

Is it the same BM you got for prior versions? Can there be some change in what GF considers to be a "current" BM for that deployment?

> `BeanManagerLookupService.lookupBeanManager(type, this).`

Yes, this is an internal way to lookup proper BM for given class. In this case I think it is used to avoid re-creating the AT which might already exist somewhere else so you try to look it up first and only create it if it doesn't already exist. I guess it is needed here because you are not allowed to have duplicit ATs in the deployment unless they have different IDs.

`EjbSupport` is a per-archive service, so if your archive doesn't have that service, I'd look into where it gets registered and why it gets skipped when it was previously included. Like I said before, I didn't change anything in Weld in this regard, at least not consciously :)

PR #23816, opened 2022-02-20
Add EE 10 schemas

OndroMih:
Adds new schemas for EE 10 from https://jakarta.ee/xml/ns/jakartaee/#10

Needed in order to deploy test applications for the Jakarta Platform TCK.

Recreated https://github.com/eclipse-ee4j/glassfish/pull/23812 against the master branch.

PR #23968, opened 2022-06-05
Add EE 10 schemas from https://jakarta.ee/xml/ns/jakartaee/#10

hs536:
* follow-up: https://github.com/eclipse-ee4j/glassfish/pull/23816

Add the missing schema files for EE10 published from https://jakarta.ee/xml/ns/jakartaee/#10

### DIFF
--- a/appserver/deployment/schemas/src/main/resources/glassfish/lib/schemas/connector_2_1.xsd
+++ b/appserver/deployment/schemas/src/main/resources/glassfish/lib/schemas/connector_2_1.xsd
@@ -1,99 +1,103 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!--
-
-    Copyright (c) 2009, 2020 Oracle and/or its affiliates. All rights reserved.
-
-    This program and the accompanying materials are made available under the
-    terms of the Eclipse Public License v. 2.0, which is available at
-    http://www.eclipse.org/legal/epl-2.0.
-
-    This Source Code may also be made available under the following Secondary
-    Licenses when the conditions for such availability set forth in the
-    Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
-    version 2 with the GNU Classpath Exception, which is available at
-    https://www.gnu.org/software/classpath/license.html.
-
-    SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
-
--->
-
 <xsd:schema xmlns="http://www.w3.org/2001/XMLSchema"
-	    targetNamespace="https://jakarta.ee/xml/ns/jakartaee"
-	    xmlns:jakartaee="https://jakarta.ee/xml/ns/jakartaee"
-	    xmlns:xsd="http://www.w3.org/2001/XMLSchema"
-	    elementFormDefault="qualified"
-	    attributeFormDefault="unqualified"
-	    version="2.1">
-
+            targetNamespace="https://jakarta.ee/xml/ns/jakartaee"
+            xmlns:jakartaee="https://jakarta.ee/xml/ns/jakartaee"
+            xmlns:xsd="http://www.w3.org/2001/XMLSchema"
+            elementFormDefault="qualified"
+            attributeFormDefault="unqualified"
+            version="2.1">
   <xsd:annotation>
     <xsd:documentation>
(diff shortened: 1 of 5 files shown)


pzygielo:
It looks that committed files were re-formatted, as they differ between this PR and https://jakarta.ee/xml/ns/jakartaee/#10.
I'd prefer to have them copied 1:1.

hs536:
@pzygielo Thanks for your comment. The preceding MR was formatted, so I went along with that, but would it be preferable to store the raw file as is?

pzygielo:
> The preceding MR was formatted,

I have no time to complain about everything :grin: 

> would it be preferable to store the raw file as is just for this ones?

I only expressed my preference - to have no difference found between **files** published by Jakarta EE and stored by GlassFish. I used `diff`, and it was not happy.

I didn't compare **the contents**, which is the thing that matters in the end, so it might be OK - but I don't know.

arjantijms:
As the previous one was formatted, let's do it here too. If it appears to be troublesome later, we can unformat them all?

arjantijms (approved):

dmatej (dismissed):

dmatej:
I would prefer even those original files would be without trailing spaces and well formatted ;)
Otherwise it is always an issue here and there. 
Some tools can live with that (kdiff3 in Kubuntu can compare even server.log files).

appserver/deployment/schemas/src/main/resources/glassfish/lib/schemas/connector_2_1.xsd
```diff
@@ -0,0 +1,1163 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (c) 2009, 2020 Oracle and/or its affiliates. All rights reserved.
```
pzygielo, line 4: At https://jakarta.ee/xml/ns/jakartaee/#10 - it is `2009, 2021`.
So this isn't just copied and re-formatted.
What is the source of this file, then?

appserver/deployment/schemas/src/main/resources/glassfish/lib/schemas/connector_2_1.xsd
```diff
@@ -0,0 +1,1163 @@
...
+	    targetNamespace="https://jakarta.ee/xml/ns/jakartaee"
+	    xmlns:jakartaee="https://jakarta.ee/xml/ns/jakartaee"
+	    xmlns:xsd="http://www.w3.org/2001/XMLSchema"
+	    elementFormDefault="qualified"
+	    attributeFormDefault="unqualified"
+	    version="2.1">
+
+  <xsd:annotation>
+    <xsd:documentation>
+      <xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="license.inc" parse="text"/>
```
pzygielo, line 30: Where one can find this `license.inc`?

pzygielo: Just guessing - all files in this PR were copied without a change from https://github.com/eclipse-ee4j/jakartaee-schemas/tree/master/xml/src, from `xsds`es.
I'd expect them to be processed and taken as published instead.

appserver/deployment/schemas/src/main/resources/glassfish/lib/schemas/connector_2_1.xsd
hs536: As you commented later, the source of this file is [the file below https://github.com/eclipse-ee4j/jakartaee-schemas/tree/master/xml/src.](https://github.com/eclipse-ee4j/jakartaee-schemas/blob/master/xml/src/connector_2_1.xsds).

appserver/deployment/schemas/src/main/resources/glassfish/lib/schemas/connector_2_1.xsd
hs536: Thank you for the catch. I was not aware that the original file did not match the published file.
So... for now, I'll just format the currently published xsd file according to [Eclipse codestyle](https://github.com/eclipse-ee4j/ee4j/tree/master/codestyle).

pzygielo (approved):
Checked that new files in 7d8d1526bdeb817c926774d590c6e2cc214f3735 are the same as at https://jakarta.ee/xml/ns/jakartaee/#10.

dmatej (approved):

PR #23973, opened 2022-06-07
Cleanups

### LINKED ISSUE #23979: Possible NPEx due to missing return
_Originally posted by @pzygielo in https://github.com/eclipse-ee4j/glassfish/pull/23957#discussion_r892681962_

dmatej:
- resources, formatting, loops, javadoc, ...
- classloaders, containers, archives, ...

### DIFF
--- a/appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
+++ b/appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
@@ -21,27 +21,16 @@ package org.glassfish.web.loader;
 import com.sun.appserv.BytecodePreprocessor;
 import com.sun.appserv.server.util.PreprocessorUtil;
 import com.sun.enterprise.loader.ResourceLocator;
-import com.sun.enterprise.util.io.FileUtils;
-import org.apache.naming.JndiPermission;
-import org.apache.naming.resources.DirContextURLStreamHandler;
-import org.apache.naming.resources.JarFileResourcesProvider;
-import org.apache.naming.resources.ProxyDirContext;
-import org.apache.naming.resources.WebDirContext;
-import org.apache.naming.resources.Resource;
-import org.apache.naming.resources.ResourceAttributes;
-import org.glassfish.api.deployment.InstrumentableClassLoader;
-import org.glassfish.web.util.ExceptionUtils;
-import org.glassfish.web.util.IntrospectionUtils;
-import org.glassfish.hk2.api.PreDestroy;
 import com.sun.enterprise.security.integration.DDPermissionsLoader;
 import com.sun.enterprise.security.integration.PermsHolder;
+import com.sun.enterprise.util.io.FileUtils;
 
-import javax.naming.Binding;
-import javax.naming.NameClassPair;
-import javax.naming.NamingEnumeration;
-import javax.naming.NamingException;
-import javax.naming.directory.DirContext;
-import java.io.*;
+import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.FilePermission;
+import java.io.IOException;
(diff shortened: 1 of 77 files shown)


arjantijms (approved):

appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
```diff
@@ -3271,19 +3217,19 @@ private boolean isDelegateFirstResource(String name) {
...
-            return !useMyFaces && !name.startsWith("jakarta/faces");
+        if (name.startsWith("jakarta/faces") || name.startsWith("com/sun/faces")) {
+            return !useMyFaces;
         }
-        if (name.startsWith("com/sun/faces")) {
-            return !useMyFaces && !name.startsWith("com/sun/faces/extensions");
+        if (name.startsWith("jakarta") || name.startsWith("javax")) {
+            return true;
         }
-        if (name.startsWith("org/apache/taglibs/standard")) {
```
hs536, line 3280: @dmatej Do you remember the reason why this line was removed?

dmatej: The reason is that I thought that these resources moved to jakarta.* packages, but now I see that was just api, impl is in jakarta.servlet.jsp.jstl-3.0.0.jar and that still uses the `org.apache.taglibs` package. However ...

- in those packages are no resources except classes and resource bundles.
- all TCK and GlassFish tests passed
- tld files are under the path `META-INF/*.tld`, so then this delegation filter still wasn't applied

What do you think, is it alright or should we put these "exceptions" back? Did you notice some issue? If I would put that back, at least I would like to document why is it there.

hs536: @dmatej 
Maybe there is no problem so far.
However, inconsistencies in the following package list may cause problems in the future I think:

- for classes
https://github.com/eclipse-ee4j/glassfish/blob/f2dd7ea034f467b9c96c808c4ae6c68bd597b542/appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java#L145-L156

- for resources
https://github.com/eclipse-ee4j/glassfish/blob/f2dd7ea034f467b9c96c808c4ae6c68bd597b542/appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java#L3216-L3230

appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
```diff
@@ -652,18 +635,14 @@ public void setWorkDir(File workDir) {
 
     public void setUseMyFaces(boolean useMyFaces) {
         this.useMyFaces = useMyFaces;
-        if (useMyFaces) {
```
hs536, line 655: @dmatej 
Without this clause, users cannot override the faces package even if they enable useMyFaces.

appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
dmatej: Yeah, I noticed inconsistencies even in the original code, it's quite confusing what is useful, what was useful 15 years ago, and what is even harmful. I would like to have more tests for that as it seems to me it was created iteratively just when some TCK test failed or some customer complained that something is not working as he expected. 

That myFaces is another example. Why was the attribute created? What about PrimeFaces and BootsFaces? And is it still useful in GlassFish7?

dmatej: I think now I have the answer in the next comment, thanks!

PR #24000, opened 2022-06-17
Helidon Config - CDI TCK Fix 2

MattGill98:
Injection points were being created by the Config CDI extension regardless of if an `@ConfigProperty` annotation was present, which was causing the CDI TCK to fail unexpectedly. This change just checks for an annotation before creating an injection point.


arjantijms (approved):

arjantijms:
Thanks a bunch!

PR #24002, opened 2022-06-19
Jdk17 Profile for MicroProfile dependency

arjantijms:
[Add JDK17 profile for MP modules](https://github.com/eclipse-ee4j/glassfish/commit/484345d5b3e6945c756244f7efadb7c193f04715)

### DIFF
--- a/appserver/featuresets/glassfish/pom.xml
+++ b/appserver/featuresets/glassfish/pom.xml
@@ -799,42 +799,52 @@
                 </exclusion>
             </exclusions>
         </dependency>
-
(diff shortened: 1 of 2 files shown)


OndroMih:
Looks good. Note that GF distribution built with Java 11 won't contain the modules that require Java 17 (MicroProfile), while if built with Java 17, the distribution will contain them.

dmatej (approved):
This is one option, second is to resolve that when starting the server, but as I am digging right in these places now, we can revert this change later again.

PR #24021, opened 2022-06-23
Fixed default httpOnly value - should be true

hs536:
* Follow-up: https://github.com/eclipse-ee4j/grizzly/pull/2168


dmatej (approved):
Hmmm, and again my mistake. I would like to ask myself at that time what led me to change the dfault value in 5f22f7a307bae2875f1172620b11755fa26d347f

PR #24035, opened 2022-06-29
Rules for resource and class loading resolution aligned

dmatej:
- overridablePackages are not added, but set, and then just used. Because it is just configured and then read, we don't need ConcurrentLinkedQueue
- useMyFaces now affects both classes and resources
- org.apache.commons.logging doesn't exist in GlassFish - removed
- com.sun.enterprise.overrideablejavaxpackages renamed to org.glassfish.main.webappCL.overridablePackages
  - I don't think anyone used it for years, but it really could be useful.
  - overridable packages are now set only by this system property.
- removed redundant log
- removed synchronized on setOverridablePackages as it is set just from a single thread which creates the instance.
- based on review from @hs536 : 
  - https://github.com/eclipse-ee4j/glassfish/pull/23973#pullrequestreview-1019545385
  - https://github.com/eclipse-ee4j/glassfish/pull/23973#pullrequestreview-1021094261

### DIFF
--- a/appserver/web/war-util/src/main/java/com/sun/enterprise/glassfish/web/WarHandler.java
+++ b/appserver/web/war-util/src/main/java/com/sun/enterprise/glassfish/web/WarHandler.java
@@ -168,14 +168,8 @@ public class WarHandler extends AbstractArchiveHandler {
 
         } catch(XMLStreamException xse) {
             logger.log(Level.SEVERE, xse.getMessage(), xse);
-            if (logger.isLoggable(Level.FINE)) {
-                logger.log(Level.FINE, xse.getMessage(), xse);
-            }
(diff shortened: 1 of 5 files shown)


dmatej:
TCK servlet/api passed locally:
`[runcts] OUT => [javatest.batch] Number of tests completed:  849 (849 passed, 0 failed, 0 with errors)`

hs536 (approved):
LGTM

hs536:
@dmatej I'm sorry later but useMyFaces doesn't seem to work for classes with GF7-nightly... //for resources, works fine

dmatej:
Hmmm, bad news. We have issues with Jenkins now, but I will create an integration test next week, so we can resolve this once forever.

PR #24055, opened 2022-07-30
Fixes #24054 - Move ApplicationValidator messages to DOLUtils class

### LINKED ISSUE #24054: An error message logged by application validator during deployment prints a message key instead of a human-readable message
## Environment Details
* GlassFish Version (and build number): 7.0.0-M7
* JDK version: Java 17
* OS: Linux
* Database: N/A

----------

## Problem Description
[//]: # "Describe the bug in detail highlighting current behavior vs expected behavior"
[//]: # "State if the problem is easily reproducible or happens intermittently"
[//]: # "Include stack traces or command outputs"

When application is detected to be invalid by the application validator (class ApplicationValidator), GlassFish prints a SEVERE message `enterprise.deployment.util.application.invalid.namespace`. It should instead print: `DEP0004:Deployment failed because a conflict occured for jndi-name: some/jndi-name for application: some-application`, according to the definition in [LogStrings.properties](https://github.com/eclipse-ee4j/glassfish/blob/18953db0fd7b28648314df13d7b349dc44d33f2a/appserver/deployment/dol/src/main/resources/com/sun/logging/enterprise/system/tools/deployment/LogStrings.properties)

OndroMih:
So that they are found and translated by DOLUtils.getDefaultLogger() logger, which uses `org.glassfish.deployment.LogMessages` as the bundle name.
Fixes https://github.com/eclipse-ee4j/glassfish/issues/24054.

Where I didn't know what is exactly the cause and action for the message, I left the cause and action properties "Unknown". If anybody has a clue, please suggest. I wasn't able to reproduce the conditions under which those messages are logged so I don't understand what's the cause and action.

I also fixed when the JNDI_LOOKUP_FAILED is logged. without this fix, it was logged when an application defined a resource with a global JNDI name or without a context (e.g. `env/X`) and the resource didn't exist on the server already (lookup failed). but this is an expected situation. The problem, on the other hand, is when an application tries to define such resource and the global JNDI name is already defined on the server (i.e. lookup doesn't give an exception).

The case for JNDI_LOOKUP_FAILED is easy to reproduce. 

Behavior without this fix:
* deploy an application that defines a resource with that JNDI name, e.g.: `@DataSourceDefinition(name = "env/X", className = "some.Driver2")`
* deployment succeeds
* but there's a SEVERE message in the log: "enterprise.deployment.util.application.lookup"

Then:
* undeploy
* create a custom resource on the server: `asadmin create-custom-resource --restype=java.lang.String --factoryclass=org.glassfish.resources.custom.factory.PrimitivesAndStringFactory --property value="myvalue" 'env/X'`
* deploy again
* deployment fails
* there are other SEVERE messages but the "enterprise.deployment.util.application.lookup" message isn't printed

After the fix:
* delete the custom resource if it's defined: `asadmin delete-custom-resource 'env/X'`
* deploy an application that defines a resource with that JNDI name, e.g.: `@DataSourceDefinition(name = "env/X", className = "some.Driver2")`
* deployment succeeds
* there's no SEVERE message about JNDI look up

Then:
* undeploy
* create a custom resource on the server: `asadmin create-custom-resource --restype=java.lang.String --factoryclass=org.glassfish.resources.custom.factory.PrimitivesAndStringFactory --property value="myvalue" 'env/X'`
* deploy again
* deployment fails
* you should see the message `DEP0003:The jndi-name is already used in the global tree failed for given jndi-name: env/X` in the log


dmatej:
You have 9 Checkstyle violations, probably trailing spaces or tabs in the code.
Otherwise it looks good!

arjantijms (approved):

PR #24067, opened 2022-08-04
Asciidoctor plugin now supports JDK17 too

dmatej:
1. Updated dependencies
2. Set the TOC level from 1 for HTML and 2 for PDF to 3 for both - until we add some indexing service able to search in documentation it is easier to find something just by rolling through the TOC menu.

Tested both with JDK11 and JDK17. 

After the merge I will visit yet [Jenkins](https://ci.eclipse.org/glassfish/view/all/job/glassfish-docs-publish/configure) as it uses JDK8 which was mandatory before this PR.

### DIFF
--- a/docs/parent/pom.xml
+++ b/docs/parent/pom.xml
@@ -35,7 +35,7 @@
         <asciidoctor.maven.plugin.version>2.2.2</asciidoctor.maven.plugin.version>
         <asciidoctorj.version>2.5.3</asciidoctorj.version>
         <asciidoctorj.pdf.version>2.1.6</asciidoctorj.pdf.version>
-        <jruby.version>9.2.20.1</jruby.version>
+        <jruby.version>9.3.2.0</jruby.version>
         <!-- status: DRAFT, BETA, etc., or blank for final -->
         <status>DRAFT</status>
         <pdf.fileName>${project.artifactId}.pdf</pdf.fileName>
--- a/docs/parent/pom.xml
+++ b/docs/parent/pom.xml
@@ -39,7 +39,7 @@
         <!-- status: DRAFT, BETA, etc., or blank for final -->
         <status>DRAFT</status>
         <pdf.fileName>${project.artifactId}.pdf</pdf.fileName>
-        <pdf.toclevels>2</pdf.toclevels>
+        <pdf.toclevels>3</pdf.toclevels>
         <productName>Eclipse GlassFish</productName>
         <bookDirectory>${project.build.directory}/book</bookDirectory>
     </properties>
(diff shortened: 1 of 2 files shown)


docs/parent/pom.xml
```diff
@@ -32,14 +32,14 @@
...
         <maven.site.skip>true</maven.site.skip>
         <maven.deploy.skip>true</maven.deploy.skip>
-        <asciidoctor.maven.plugin.version>2.2.1</asciidoctor.maven.plugin.version>
-        <asciidoctorj.version>2.5.2</asciidoctorj.version>
-        <asciidoctorj.pdf.version>1.6.0</asciidoctorj.pdf.version>
-        <jruby.version>9.3.2.0</jruby.version>
+        <asciidoctor.maven.plugin.version>2.2.2</asciidoctor.maven.plugin.version>
+        <asciidoctorj.version>2.5.3</asciidoctorj.version>
+        <asciidoctorj.pdf.version>2.1.6</asciidoctorj.pdf.version>
+        <jruby.version>9.2.20.1</jruby.version>
```
pzygielo, line 38: Why downgrade JRuby? It seems to work fine with 9.3.2.0.

dmatej: First I tried to upgrade it to the latest, but there is some method missing, so I used the same as was used in asciidoctor. It probably doesn't matter too much.

dmatej: Ok, I will put the 9.3.2.0 back

pzygielo: > First I tried to upgrade it to the latest, but there is some method missing

Just guessing - it was due to
- https://github.com/jruby/jruby/issues/7261

pzygielo: My understanding of jruby#7261 - if plugin is bumped to 2.2.1 - its `jruby-complete` dependency 
https://github.com/eclipse-ee4j/glassfish/blob/95d5f2f2c9d931ed5dc551ea437e0e575dd1861e/docs/parent/pom.xml#L72-L76
should be switched to `jruby`. Otherwise - dependency with version as defined by plugin (`jruby` with different version) is used in first place, only then `jruby-complete` is added to CP.

dmatej: Oh, good to know! :-)

dmatej: Hmm, I was digging in asciidoctor poms now, I believe there are just two options:
* change the dependency from jruby-complete to jruby
  - then with every update we have to choose a compatible dependency set (plugin vs. asciidoctor vs. jruby)
  - needs some fiddling, but safe
* remove jruby dependency
  - then with updates the plugin decides the jruby version even if asciidoctorj would need different
  - it would be bit harder to find out what is happening

Reason: while the plugin uses the jruby.version property, the asciidoctor doesn't, it explicitly uses the version.
```
<dependency>
  <groupId>org.jruby</groupId>
  <artifactId>jruby</artifactId>
  <version>9.2.20.1</version>
  <scope>compile</scope>
</dependency>
```

pzygielo (approved):

PR #24076, opened 2022-08-11
More robust solution for Helidon

dmatej:
Implemented detection of compatibility of the CDI extension with the JDK runtime. If the JDK is not compatible, extension file is excluded.

Original state
- GF built with JDK17 could not be used with JDK11, because it contained Helidon which required JDK17.


arjantijms (approved):

appserver/web/weld-integration/src/main/java/org/glassfish/weld/FilteringClassLoader.java
```diff
@@ -0,0 +1,124 @@
...
+    public FilteringClassLoader(ClassLoader parent) {
+        super(parent);
+        LOG.log(Level.FINEST, "Parent: {0}", parent);
+    }
+
+
+    @Override
+    public URL getResource(final String name) {
+        LOG.log(Level.FINE, "getResource(): {0}", name);
+        return getParent().getResource(name);
```
OndroMih, line 59: This method doesn't need to filter the resources? Shouldn't this method gelegate to the `getResources` method to also filter resources here?

appserver/web/weld-integration/src/main/java/org/glassfish/weld/FilteringClassLoader.java
```diff
@@ -0,0 +1,124 @@
...
+
+import static java.util.Collections.enumeration;
+import static java.util.stream.StreamSupport.stream;
+
+/**
+ * This classloader filters out extensions which are incompatible with current environment.
+ *
+ * @author David Matejcek
+ */
+class FilteringClassLoader extends ClassLoader {
```
OndroMih, line 43: I would rename the classloader to something that describes better what it does. For example, `JdkFilteringClassLoader` or `JavaRuntimeFilteringClassLoader`.

appserver/web/weld-integration/src/main/java/org/glassfish/weld/FilteringClassLoader.java
dmatej: I have fallen to this trap recently too, but these methods are not on the same level. They usually follow the scenario when you first get all resources by one method, then load one of them by the second one. Then you would do the filtering twice with no reason.
The class has just limited visibility, so we can do this as we know it is always used just this way (but I should add some test to cover that - test containers).

appserver/web/weld-integration/src/main/java/org/glassfish/weld/FilteringClassLoader.java
dmatej: It would be too verbose (impl detail), rules can change later. But more probable is that we will kill this CL after moving to JDK17.

appserver/web/weld-integration/src/main/java/org/glassfish/weld/FilteringClassLoader.java
```diff
@@ -0,0 +1,124 @@
...
+
+    private boolean isCompatible(final URL extensionUrl) {
+        final URL manifestURL = toManifestURL(extensionUrl);
+        try (InputStream stream = manifestURL.openStream()) {
+            final Manifest manifest = new Manifest(stream);
+            final Version requiredMinVersion = getRequiredMinimalJavaVersion(manifest.getMainAttributes());
+            if (requiredMinVersion == null) {
+                return true;
+            }
+            return requiredMinVersion.compareTo(Runtime.version()) < 0;
```
pzygielo, line 97: This is unclear for me. It's possible there is some assumption for values returned by `getRequiredMinimalJavaVersion` and `Runtime.version()` involved here.

But if `requiredMinVersion == X` and `Runtime.version() == X` then `isCompatible` returns `0 < 0` (`false`) (as `X.compareTo(X)` is `0`).

dmatej: Hmmmmm, I should always write a test :-D
For now it works, because the runtime version is always greater than that from osgi requirement. But yeah, it is not correct.

`Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=17))"`
Runtime.version() gives ie. 11.0.14 for JDK11

I think that situation where those versions would be really equal is not much probable, but I will fix it.

PR #24082, opened 2022-08-15
DOL module vs. technical debt: generics

### LINKED ISSUE #23601: Need more descriptive error message in case of an invalid schema
## Environment Details
* GlassFish Version (and build number): 6.2.1 release

----------

## Problem Description

I am in the process of updating some old JEE projects to Jakarta EE 9(.1). As the projects are quite simple, it takes me only two steps: replacing `javax.` package name prefixes with `jakarta.` as appropriate, and updating the schemas. This issue concerns the latter step.

If I try to use the projects with GlassFish 6.2.1, I get the error message [thrown on this line of code](https://github.com/eclipse-ee4j/glassfish/blob/master/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/node/SaxParserHandler.java#L229). It is possibly loaded from [this file](https://github.com/eclipse-ee4j/glassfish/blob/master/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/node/LocalStrings.properties), however, the two texts are the same: `Requested schema  is not found in local repository, please ensure that there are no typos in the XML namespace declaration.`

This error message is not helpful enough, as it doesn't mention the name of the wrong schema, forcing me into a guessing game.

Suggestion: include the name of the missing/wrong schema's name in the error message like this: "Requested schema **wrong_schema.xsd** is not found in local repository, please ensure that there are no typos in the XML namespace declaration." Optionally, for further assistance, the message could also mention the path of the local repository.

Performing the change seems to be trivial: the required name should be one of the two arguments [on this line](https://github.com/eclipse-ee4j/glassfish/blob/master/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/node/SaxParserHandler.java#L202).

dmatej:
This PR is rather a "feature branch merge" than single fix, because all changes were done with a target to reduce the extreme tech debt of the DOL modules (deployment). Before the merge I would like to run all available tests to ensure that nothing broke, including TCK.

* I tried to add generics where it was possible and known
* Fixed issue #23601 and several similar unresolved messages
* Removed some deprecated or unused methods
* Fixed IO leaks in DOL


dmatej:
Note for myself - ejb_group_embedded passed locally twice.

dmatej:
I will pause now, or slow down ... I will create another parallel PR for the TCK runner ... so I can ensure that this huge PR will not break anything. I committed part by part last week, because I could not find mistake I did ... now I found it. When I replaced `"...Interceptors"` string by `Interceptor.class.getName()` ... yeah, that missing s was it :-D

Bad was that I could not find it so long, good is that it forced me to analyze the API (broken, leaking) deeply and add some notes how to clean it up so it would make sense. After finishing this PR DOL will not be a scary monster any more :-) - despite it doesn't look like that, the worst part is done. The more code changes will come yet, but they should be much safer.

But again, TCK first.

dmatej:
I'm yet passing TCK tests which failed yesterday, but it seems this should be the last fix. Refactoring will continue, but it would be good to have some safe point not just in the dev branch ...

arjantijms (approved):

PR #24089, opened 2022-08-24
Fixed comparison in JDK version requirement validation

dmatej:
- it did not cause any errors (yet), but it should be correct.
- see https://github.com/eclipse-ee4j/glassfish/pull/24076#discussion_r953517629 , thanks @pzygielo for the review!


pzygielo (approved):

pzygielo:
> It is sensitive to something, maybe not related to this PR, but it seems quite reproducible here ...

Not only here, the master fails with
```
LoggingRestITest.logFileNames
```
as well.

PR #24096, opened 2022-08-29
Remove unused local variables

pzygielo:


dmatej (approved):
It conflicts with my branch, but whatever, I can handle it.

PR #24118, opened 2022-09-27
Fix #24115 Reason phrase is not printed in the validation error message

### LINKED ISSUE #24115: Reason phrase is not printed in the validation error message
## Environment Details
* GlassFish Version (and build number): GlassFish 7.0.0
* OS: CentOS Linux release 7
* Database: none

----------

## Problem Description
Reason phrase is not printed in the validation error message.  
//Message key is printed instead of expected message when an illegal value is set.  
```
> asadmin set server-config.iiop-service.orb.message-fragment-size=-1
remote failure: Could not change the attributes: Constraints for this Orb configuration have been violated: on property [ messageFragmentSize ] violation reason [ {jakarta.validation.constraints.Min.message} ]
Constraints for this Orb configuration have been violated: on property [ messageFragmentSize ] violation reason [ {jakarta.validation.constraints.Min.message} ]
Command set failed.
```

## Steps to reproduce
1. `asadmin start-domain`
2. `asadmin set server-config.iiop-service.orb.message-fragment-size=-1`

kaido207:
Here should be `MessageInterpolatorImpl.class` rather than `MessageInterpolator.class` because message resources are loaded by Impl module, not API module.

https://github.com/eclipse-ee4j/glassfish/blob/9a68cc8076d14c77aba791fd5c955b7c7a175593/nucleus/hk2/hk2-config/src/main/java/org/jvnet/hk2/config/MessageInterpolatorImpl.java#L121
   

Signed-off-by: kaido207 <kaido.hiroki@fujitsu.com>

### DIFF
--- a/nucleus/hk2/hk2-config/src/main/java/org/jvnet/hk2/config/MessageInterpolatorImpl.java
+++ b/nucleus/hk2/hk2-config/src/main/java/org/jvnet/hk2/config/MessageInterpolatorImpl.java
@@ -113,22 +113,12 @@ public class MessageInterpolatorImpl implements MessageInterpolator {
         // if the message is not already in the cache we have to run step 1-3 of the message resolution
         if (resolvedMessage == null) {
             ResourceBundle userResourceBundle = new ContextResourceBundle(context, locale);
-            ClassLoader cl;
-            if (System.getSecurityManager()!=null) {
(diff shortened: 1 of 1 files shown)


dmatej:
Interesting error on Jenkins, I will take a look locally ... invalid character encoding seems suspicious (may be also caused by some other bug).

dmatej (approved):

PR #24140, opened 2022-10-25
Use spec-version plugin to check distributions

pzygielo:


dmatej:
I would rather put that directly to the maven build where appropriate. Just standard maven lifecycle, not just Jenkins.

pzygielo:
- I planned to propose only 28ca60d5474464a3864d94806031f1c7a7f29916, and use this PR just to confirm #24141

- Encouraged with above comment to open more complete one #24142

PR #24144, opened 2022-10-26
Try to build w/o separate container

pzygielo:


dmatej:
In this case this will not work - ant based tests are quite hungry, maven verify too, but it changes. I was tuning it some time ago to this state, it took some time :-)
I have yet plans with all those tests, that time will come soon, then the configuration will change.

dmatej:
Yet one thing - I have prepared another generation of the container, but I have found a bug in Temurin, so I am waiting for the next release.

pzygielo:
> In this case this will not work

I'm afraid I was not aware of that.

But can you be more specific and tell what isn't working [here](https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/job/PR-24144/12/), please?

dmatej:
> > In this case this will not work
> 
> I'm afraid I was not aware of that.
> 
> But can you be more specific and tell what isn't working [here](https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/job/PR-24144/12/), please?

Czech proverb: 

> When someone says something can't be done, there's always someone who doesn't know it can't be done and does it!

Nice work, no OOME, did we fix something so it has lower footprint, or does it some adaptation now, or was it affected by [this](https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2102)? (probably not as the issue is still open).

I love when I am wrong in such cases, thank you! :-)

dmatej (approved):

pzygielo:
Thanks for checking.

dmatej:
Thank you, I would not even try after I burned my fingers for the last time! 
Seems now we can drop those docker images.

Issue #23309, opened 2020-12-29
Remote clients cannot discover glassfish 5.1 EJBs due to spurious java.io.NotSerializableException


Blavo:
## Environment Details

Glassfish 5.1.0 (default-private)
Java version "1.8.0_161"
Mac OS Catalina 10.15.2
Java DB

## Problem Description

Remote clients cannot discover glassfish EJBs because they cannot list the 'java:global' naming context due to a spurious java.io.NotSerializableException.

What fails is the naming context list() operation. This is the facility used to discover remotely-invocable services, such as EJBs, on a remote server. These services are listed under a well-known context ('java:global'). In my case I am looking for remotely-invocable EJBs on a remote server. The remote (Glassfish) server binds these under a sub-context of java:global where the sub context is remote-server-dependent and generally unknown to the caller. So I have to list the java:global context to discover that sub-context.  (The sub-context is usually the deployment name of the application containing the EJBs. Multiple such applications would result in multiple sub-contexts of 'java:global'.)

My attempt to list the java:global context gives the same java.io.OptionalDataException described in #17220 or a more specific java.io.NotSerializableException naming class org.glassfish.kernel.javaee.MEJBNamingObjectProxy.  

The call signature for list() is 'NamingEnumeration<NameClassPair> list(String name) throws NamingException'. All elements of the signature, recursively, are serializable. So a correct response to the list() call is incapable of giving a CORBA error due to serialisation problems. So definitely something wrong with the implementation of list() in glassfish.

Part of the original response to #17220 is that the cause may be that the sub-context being listed contains JNDI entries that aren't serializable. Such entries are allowed (though pointless - see below). In fact it doesn't matter whether the elements of the sub-context are serializable or not because only their class names are being returned in a list() call. Serialisability doesn't come into play until the result of a list() call is used to make a lookup() call to obtain an instance of the bound class type. 

[...]

PR #24146, opened 2022-11-01
JNDI Names + Concurrency + DOL cleanups

dmatej:
In this PR I tried to separate names and JNDI names, because I noticed that some JNDI names are weird/wrong despite they pass all tests. It is much more complicated than I originally thought, so this is not yet complete work, but it is a intermediate state which passed all available tests.

This PR passed whole TCK and all tests we have.

## Main changes

- Upgrade to Soteria 3.0.1 (fixes some of tests, Soteria fixed because of them)
- Introduced SimpleJndiName
  - Value object with simple methods to make usual conversions easy
  - Reduced copy and paste
  - Strings named as jndiName usually wrapped by this object (not always, sometimes it would not be effective)
  - Cannot be simply replaced by CompositeName yet, because not all names are compliant and also CompositeName does parsing.
- GlassfishNamingManagerImpl - using new JavaNamespace instead of HashMap.
- SerialContext - testMode kicked out as it changed the beahvior for unit tests. Using EasyMock instead.
- ActiveJmsResourceAdapter - fixed setting port and timeout as they are numbers now (GF expected string setters)
- JdbcDataSource - implemented JDBC4 methods rather than NotImplementedException (rather strict impl for now)
- Fixed exception in ConnectorConnectionPoolAdminServiceImpl - getUserName doesn't exist,  getUser does
- Refactored sending HTTP 500 to the client in ContainerMapper - previous version had too wide try-catch blocks where in one of deployment_all test cases one exception caused another which was preventable. This version logs just the original exception and then a message if it cannot send the http 500 or another exception if it tried and it failed, but no NPE.
- More generics
- Reviewed many logs, their meaning and effectivity

## FIXMEs

- Related: https://github.com/eclipse-ee4j/glassfish/issues/23309
- One commit did part of this, but had to put it back: https://github.com/eclipse-ee4j/glassfish/issues/24171
- Discovered: https://github.com/eclipse-ee4j/glassfish/issues/24172
- Discovered: https://github.com/eclipse-ee4j/glassfish/issues/24169 (Soteria update resolved that, but someting in GF smells)
- Added many FIXMEs into the code 
  - they will be used to track related things on future pull requests.
  - some relate to old deprecated methods where already were some messages from past developers
  - some relate to things I noticed, which will need standalone PR
  - some are bound somehow to each other, but I wasn't sure with consequences so I just added a comment. 

## Directions

- make JNDI in GlassFish effective, logical and correct
- the hard part: JNDI can be used also remotely via IIOP/Corba; that has expectations - correct and logical names which are now magically translated inside the server.


dmatej:
I will yet rebase this to merged https://github.com/eclipse-ee4j/glassfish/pull/24174 before the merge, there are same files touched.

arjantijms (approved):
A wee bit too large to review every file, but LGTM and passes all tests.

PR #24148, opened 2022-11-02
Set generated keystore type to JKS

arjantijms:
GlassFish generates an internal keystore using the keytool command during the build. The type of the keystore depends on the JDK being used. From JDK12 onwards it defaults to PKS12, which throws the following exceptions on JDK 11:

```
Integrity check failed: java.security.NoSuchAlgorithmException: Algorithm HmacPBESHA256 not available
```

By setting the keystore explicitly to JKS, it's supported on both 11 and 17.

### DIFF
--- a/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/KeystoreManager.java
+++ b/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/KeystoreManager.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2022 Contributors to the Eclipse Foundation
  * Copyright (c) 1997, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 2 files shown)


pzygielo:
1. What piece of code complains about:
```
Integrity check failed: java.security.NoSuchAlgorithmException: Algorithm HmacPBESHA256 not available
```

How can I verify that it is happy after this change?

-----

2. `keytool`s, including the one from JDK11, issue
```
Warning:
The JKS keystore uses a proprietary format. It is recommended to migrate to PKCS12 which is an industry standard format using "keytool -importkeystore -srckeystore keystore.ks -destkeystore keystore.ks -deststoretype pkcs12".
```

arjantijms:
> How can I verify that it is happy after this change?

Build with JDK 17, then start GlassFish and deploy a war to it using an older version of JDK 11.  I used sdkman to install 11.0.2:

```
sdk install java 11.0.2-open
```

```
java --version
openjdk 11.0.2 2019-01-15
OpenJDK Runtime Environment 18.9 (build 11.0.2+9)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.2+9, mixed mode)
```

With this version the exception happens on master, and does not happen with this patch. Note that with a newer JDK 11 release the exception doesn't happen on master, e.g. with 11.0.15:

```
java --version
openjdk 11.0.15 2022-04-19 LTS
OpenJDK Runtime Environment Zulu11.56+19-CA (build 11.0.15+10-LTS)
OpenJDK 64-Bit Server VM Zulu11.56+19-CA (build 11.0.15+10-LTS, mixed mode)
```

arjantijms:
>  It is recommended to migrate to PKCS12 which is an industry standard format

Indeed, so once we raise the baseline to JDK 17 we should probably start using PKCS12. However, it's just an internal keystore for GF for the self-signed certificates, so the industry standard format aspect is not super important at the moment.

pzygielo:
Seems related:
- https://bugs.openjdk.org/browse/JDK-8267599:
> pkcs12 keystores created using JDK 11.0.12 or later can't be read by earlier JDK 11 versions

arjantijms:
Yes, that's the exact issue.

dmatej (approved):

PR #24149, opened 2022-11-04
Improve exception logging during startup

arjantijms:
Some exceptions would not get logged during startup, particularly multi exceptions from HK2 wrapped in other exceptions and/or being suppressed.

### DIFF
--- a/nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/ASMain.java
+++ b/nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/ASMain.java
@@ -17,16 +17,26 @@
 
 package com.sun.enterprise.glassfish.bootstrap;
 
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
 /**
  * Tag Main to get the manifest file
  */
 public class ASMain {
+
+    private static final Logger LOG = Logger.getLogger(ASMain.class.getName());
 
     /**
      * Most of the code in this file has been moved to MainHelper
      * and ASMainOSGi
      */
     public static void main(final String[] args) throws Exception {
-        GlassFishMain.main(args);
+        try {
+            GlassFishMain.main(args);
+        } catch (Throwable t) {
+            LOG.log(Level.SEVERE, "Error starting GlassFish", t);
+            throw t;
+        }
     }
 }
--- a/nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/ASMain.java
+++ b/nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/ASMain.java
@@ -17,15 +17,10 @@
 
 package com.sun.enterprise.glassfish.bootstrap;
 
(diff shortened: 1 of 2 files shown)


nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/osgi/EmbeddedOSGiGlassFishRuntime.java
```diff
@@ -144,20 +156,41 @@ public synchronized GlassFish newGlassFish(GlassFishProperties gfProps) throws G
...
+
+        while (currentThrowable != null) {
+            if (currentThrowable instanceof MultiException) {
+                return (MultiException) currentThrowable;
+            }
+
+            for (Throwable suppressed : currentThrowable.getSuppressed()) {
+                MultiException multiException = findMultiException(suppressed);
+                if (multiException != null) {
+                    return multiException;
```
dmatej, line 185: But this way you lose other possible suppressed exceptions describing what happened, right?
I need some more detailed explanation (javadoc), at least why do you think it is ok. And comparison of the output to see it too.

nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/ASMain.java
```diff
@@ -17,16 +17,26 @@
...
     /**
      * Most of the code in this file has been moved to MainHelper
      * and ASMainOSGi
      */
     public static void main(final String[] args) throws Exception {
-        GlassFishMain.main(args);
+        try {
+            GlassFishMain.main(args);
+        } catch (Throwable t) {
+            LOG.log(Level.SEVERE, "Error starting GlassFish", t);
```
dmatej, line 38: If GF fails to start, this cannot print anyhing, because GlassFish will not load logging.properties. So in this case t.printStacktrace() would be better.

arjantijms: It does actually gets added to the output (not the log file). So you do see what happened if GF doesn't start there. Of course, we could do `t.printStacktrace()` if that's better?

nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/osgi/EmbeddedOSGiGlassFishRuntime.java
arjantijms: > But this way you lose other possible suppressed exceptions describing what happened

I don't think so, as the original exception (throwable t) is still propagated:

```
throw new GlassFishException(bundleBuilder.toString(), t);
```

The first `MultiException` encountered is just used for `bundleBuilder.toString`.

dmatej: Oh, I see, thanks!

dmatej (approved):

Issue #24167, opened 2022-11-11
Admin Console - New JDBC Connection pool page doesn't contain list of database vendors


OndroMih:
## Environment Details
* GlassFish - built from the admingui7 branch on 11 Nov 2022
* JDK version: Zulu17.36+13-CA (build 17.0.4+8-LTS)
* OS: Ubuntu 22.04.1 LTS
* Database: N/A

----------

## Problem Description

When creating a new connection pool in Admin Console, the Database Driver Vendor listbox is empty and doesn't show a list of known database Vendors. It should show the list from [dbvendor.properties](https://github.com/eclipse-ee4j/glassfish/blob/571714739b23e16c4c90d5229b806433920b3bcb/appserver/jdbc/templates/src/main/resources/glassfish/lib/install/databases/dbvendormapping/dbvendor.properties)

## Steps to reproduce

Open Admin Console, Resources, JDBC, JDBC Connection Pools, click New button. 

Current behavior:

The `Database Driver Vendor` field shows an empty listbox.

Expected behavior:

The `Database Driver Vendor` field shows a list of database vendors configured in [dbvendor.properties](https://github.com/eclipse-ee4j/glassfish/blob/571714739b23e16c4c90d5229b806433920b3bcb/appserver/jdbc/templates/src/main/resources/glassfish/lib/install/databases/dbvendormapping/dbvendor.properties)

## Impact of Issue

User experience degraded. Users need to input the name of the driver manually

[//]: # "Please choose one each Type, Component, and Priority label"

Type: bug
Component: Admin Console

OndroMih:
This bug is introduced by https://github.com/eclipse-ee4j/glassfish/issues/23641 as I explained in this comment: https://github.com/eclipse-ee4j/glassfish/issues/23641#issuecomment-1311982378

dmatej:
Can you write some trivial test in admin-tests module which would work with templates and which would cover this scenario? Without admin-gui, just something for asadmin - it had to affect it somehow, right?
Hmmm, but I don't see any related asadmin commands ... what do you think?

avpinchuk:
If it is acceptable yet, I can add some test cases soon. Please do not close this issue.

OndroMih:
> EDIT: Yet another idea - there are also rest tests, I would expect there exists some related API ...

[...]

PR #24168, opened 2022-11-12
Add back JDBC templates to the GF distribution

OndroMih:
Fixes https://github.com/eclipse-ee4j/glassfish/issues/24167 and possibly other issues.

JDBC templates need to be a ZIP artifact to unpack them into the distribution. This reverts making it a JAR artifact in https://github.com/dmatej/glassfish/commit/a53e2d0a0f9b4e96526c706dbd9132bc08b4eeb1.

WARNING: When merging, please SQUASH the commits and use the message from the last one. The second commit only reverts the first one, and they both can be removed from the history.

### DIFF
--- a/appserver/distributions/glassfish/src/main/assembly/glassfish.xml
+++ b/appserver/distributions/glassfish/src/main/assembly/glassfish.xml
@@ -334,6 +334,7 @@
                 <exclude>jakartaee.jar</exclude>
                 <exclude>appserv-rt.jar</exclude>
                 <exclude>descriptors.jar</exclude>
+                <exclude>templates.jar</exclude>
                 <exclude>sun-as-jsr88-dm.jar</exclude>
                 <exclude>gf-client.jar</exclude>
                 <exclude>mejb.jar</exclude>
--- a/appserver/distributions/glassfish/src/main/assembly/glassfish.xml
+++ b/appserver/distributions/glassfish/src/main/assembly/glassfish.xml
@@ -334,7 +334,6 @@
                 <exclude>jakartaee.jar</exclude>
                 <exclude>appserv-rt.jar</exclude>
                 <exclude>descriptors.jar</exclude>
-                <exclude>templates.jar</exclude>
                 <exclude>sun-as-jsr88-dm.jar</exclude>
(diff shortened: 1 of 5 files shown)


avpinchuk:
Maybe we can adjust assembly descriptor without fallback to `zip`?

OndroMih:
> Maybe we can adjust assembly descriptor without fallback to zip?

I tried but it's not that easy. The [glassfish-build-maven-plugin](https://github.com/eclipse-ee4j/glassfish-build-maven-plugin/https://github.com/eclipse-ee4j/glassfish-build-maven-plugin/commit/145a8b9e917312b6453e7cea2717b3eda5715370/master/src/main/java/org/glassfish/build/FeatureSetsDependenciesMojo.java) can either unpack all JARs (with some excluded), or none. 

So we would have to exclude all those that we don't want to unpack or exclude them in the assembly descriptor.

Alternatively, we could unpack the template.jar copied by the glassfish build plugin, probably using the antrun plugin.

But is it worth it? The templates artifact is never used as a JAR, it's always unpacked as a ZIP file, in both featuresets/web and jdbc-ra artifacts.

avpinchuk:
@OndroMih, adding this to `full` and `web` assembly descriptor seems to solve the problem

```xml
 <dependencySets>
        <dependencySet>
            <includes>
                <include>org.glassfish.main.jdbc:templates:jar:${artifact.version}</include>
            </includes>
            <unpack>true</unpack>
            <unpackOptions>
                <excludes>
                    <exclude>META-INF/**</exclude>
                </excludes>
            </unpackOptions>
            <outputDirectory>${install.dir.name}</outputDirectory>
        </dependencySet>
    </dependencySets>
```

That's correct way?

OndroMih:
Yes, this works, @avpinchuk, thanks! I didn't know that the Maven Assembly plugin can also unpack ZIP files. I reverted my previous commit and applied your suggestion.

dmatej:
Aha, I was scared by an e-mail writing about a revert, which would break embedded again. Now it seems alright.

dmatej (approved):

PR #24176, opened 2022-11-17
Fixed Ant home resolution and PATH

dmatej:


appserver/tests/tck/tck-runner/src/main/java/org/glassfish/main/tests/tck/ant/TckRunner.java
```diff
@@ -204,7 +204,7 @@ private void configureEnvironment(Map<String, String> env) {
         env.put("AS_DEBUG", String.valueOf(cfg.isAsadminLoggingEnabled()));
 
         env.put("PATH", cfg.getJdkDirectory().getAbsolutePath() + "/bin:" + cfg.getAntDirectory().getAbsolutePath()
-            + "/bin/:/usr/bin");
+            + "/bin/:/usr/bin:/bin");
```
pzygielo, line 207: For readability - could this be broken to one-PATH-element per line?

appserver/tests/tck/tck-runner/src/main/java/org/glassfish/main/tests/tck/ant/TckConfiguration.java
```diff
@@ -125,10 +125,18 @@ public File getJdkDirectory() {
 
 
     /**
+     * Resolves to the first value set:
+     * <ol>
+     * <li>System property ant.home (set in command line or Maven profile)
+     * <li>Environment property ANT_HOME (set in bash environment and resolved via pom.xml)
+     * <li>Default /usr/share/ant
+     * </ol>
```
pzygielo, line 133: It seems to me half of this ({ `ant.home`, `ANT_HOME` } => `ant.directory`) happens in `cfg`...

pzygielo: I'm more confused learning that `cfg` is `j.u.Properties`.
So far I don't see how `ant.home` or `ANT_HOME` participate here.

pzygielo: Hmmm... perhaps with this
https://github.com/eclipse-ee4j/glassfish/blob/12c11913e29757a520c0de2b991c1046a68544cf/appserver/tests/tck/tck-runner/src/test/resources/tck.properties#L8
+
https://github.com/eclipse-ee4j/glassfish/blob/12c11913e29757a520c0de2b991c1046a68544cf/appserver/tests/tck/tck-runner/pom.xml#L35

dmatej: Yeah, I would call it a "waterfall logic", aka "if it is that, take it, if not, try something else, or yet ..."
I had yet idea to move it to TestContainers docker image, but it has it's own price, and the TCK started moving, so soon all these modules will be replaced by some more modern way as the old TCK will be split to standalone modules.

So this way I wanted to minimize efforts so it would be able to run on my machine with -Dant.home, on my Jenkins using docker nodes with ANT_HOME and on Eclipse's Jenkins using k8s with PATH + ant.home of different version for the TckRunner (httpclient issue in the TCK, requires Ant 1.9 while the rest uses 1.10).

appserver/tests/tck/tck-runner/src/main/java/org/glassfish/main/tests/tck/ant/TckRunner.java
dmatej: Hmm, yeah, I could use the String.join. I will update it ...

pzygielo (approved):

PR #24179, opened 2022-11-18
Embed EE4J parent pom until 1.0.8 is released.

arjantijms:
This is needed to be able to deploy using JDK 17.

### DIFF
--- a/ee4j/pom.xml
+++ b/ee4j/pom.xml
@@ -1,0 +1,338 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (c) 2018 Oracle and/or its affiliates. All rights reserved.
+
+    This program and the accompanying materials are made available under the
+    terms of the Eclipse Public License v. 2.0, which is available at
+    http://www.eclipse.org/legal/epl-2.0.
+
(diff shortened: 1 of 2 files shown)


pzygielo:
Will this result in release of ee4j/project (parent) under its original coordinates with new (-M1) version from [this] glassfish repository?

pzygielo:
I think much safer would be to manage required plugin(s) (I guess nexus-staging only) instead.

dmatej:
This is a wrong idea - as @pzygielo  wrote, you will have to release them parent too as M1, or it cannot pass the Maven Central rules again.

arjantijms:
Essentially these are just some (increasingly) desperate attempts to get GF to be deployed to staging.

arjantijms:
But indeed, managing the nexus plug-in instead is probably much safer.

pzygielo:
I think, JDK17 can be used to build GF, and then JDK11 to stage with old sonatype plugin.

----

This is all unnecessary work due to unreleased change
- https://github.com/eclipse-ee4j/ee4j/pull/79#issuecomment-1287658387
- https://github.com/eclipse-ee4j/ee4j/issues/83

PR #24180, opened 2022-11-18
Update poms to use embedded parent pom

arjantijms:

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -23,7 +23,7 @@
         <groupId>org.eclipse.ee4j</groupId>
         <artifactId>project</artifactId>
         <version>1.0.7</version>
-        <relativePath />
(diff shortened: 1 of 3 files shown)


docs/pom.xml
```diff
@@ -23,7 +23,7 @@
         <groupId>org.eclipse.ee4j</groupId>
         <artifactId>project</artifactId>
         <version>1.0.7</version>
```
pzygielo, line 25: Does this match the version in file? I'd prefer to see this updated rather than have `sed` in Jenkins job.

pzygielo: - https://github.com/eclipse-ee4j/glassfish/pull/24181

dmatej: There;s already a bash script for that, updateVersions ... however I did not run it for a while, so maybe it would need some update.

PR #24185, opened 2022-11-19
Generate missing config when asadmin set is called (2nd attempt)

### LINKED ISSUE #24158: Asadmin set command fails if the parent config object isn't already present
## Environment Details
* GlassFish 7.0.0-M9
* JDK version: Zulu17.36+13-CA (build 17.0.4+8-LTS)
* OS: Ubuntu 22.04.1 LTS
* Database: N/A

----------

## Problem Description

When running asadmin set command to set a dotted configuration and the parent dotted configuration doesn't exist, the set command fails to set the configuration.

## Steps to reproduce

On a fresh GlassFish installation, run:

```
 bin/asadmin set configs.config.server-config.network-config.protocols.protocol.http-listener-1.ssl.cert-nickname=s1as
```
This will fail with the following output:

```
remote failure: No configuration found for configs.config.server-config.network-config.protocols.protocol.http-listener-1.ssl
Command set failed.
```

This is because the default `domain.xml` file doesn't contain any `ssl` element for the `http-listener-1` protocol. It works for the `http-listener-2` protocol, which contains the `ssl` element.

A similar error can be reproduced in Admin Console if GlassFish is built from these branches: https://github.com/eclipse-ee4j/glassfish/issues/23782#issuecomment-1291023938. When editing the SSL tab of the http-listener-1, pressing the Save button leads to the same error and there's no way to fix this in the Admin Console.

## Impact of Issue
[//]: When editing the SSL configuration of the HTTP listener, it's not possible to save the configuration in Admin Console.

Type:Enhancement
Component: Kernel, Admin, Admin Console

OndroMih:
Fixes https://github.com/eclipse-ee4j/glassfish/issues/24158. Automatically creates missing config nodes in the `set` command, if they aren't present in domain.xml.

Added tests to cover multiple scenarios:

* setting an attribute directly (this test was already present)
* setting an attribute through an alias (a different dotted name leading to the same config node)
* seting a custom property `a=b`
* setting an attribute on a node which isn't in the current configuration and which will be created before setting the attribute (this is the new feature implemented)

This is a fix of https://github.com/eclipse-ee4j/glassfish/pull/24160, which was reverted by https://github.com/eclipse-ee4j/glassfish/pull/24170 to pass the tests.

### DIFF
--- a/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/SetCommand.java
+++ b/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/SetCommand.java
@@ -194,6 +194,319 @@ public class SetCommand extends V2DottedNameSupport implements AdminCommand, Pos
             }
             return dottedNameForResourceName.toString().replace('.', '/');
         }
+
+        private class ExecutionContext {
+            String pattern = SetOperation.this.pattern;
+            String attrName = SetOperation.this.attrName;
+            String prefix;
+            boolean lookAtSubNodes;
+            boolean applyOverrideRules;
+
+            private Map<Dom, String> findDottedNames(TreeNode[] parentNodes, String pattern) {
+                Map<Dom, String> dottedNames = new HashMap<Dom, String>();
+                this.applyOverrideRules = false;
+                if (lookAtSubNodes) {
+                    for (TreeNode parentNode : parentNodes) {
+                        dottedNames.putAll(getAllDottedNodes(parentNode.node));
+                    }
+                    this.applyOverrideRules = true;
+                } else {
+                    for (TreeNode parentNode : parentNodes) {
+                        dottedNames.put(parentNode.node, pattern);
+                    }
+                }
+                return dottedNames;
+            }
+
+            private void resetPattern(TreeNode[] parentNodes, String somePattern) {
+                this.pattern = somePattern;
+                this.lookAtSubNodes = true;
+                if (parentNodes[0].relativeName.length() == 0
+                        || parentNodes[0].relativeName.equals("domain")) {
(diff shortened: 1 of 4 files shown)


nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/SetCommand.java
```diff
@@ -194,6 +194,320 @@ private String getResourceName() {
...
+            String prefix;
+            boolean lookAtSubNodes;
+            boolean applyOverrideRules;
+
+            private Map<Dom, String> findDottedNames(TreeNode[] parentNodes, String pattern) {
+                Map<Dom, String> dottedNames = new HashMap<Dom, String>();
+                this.applyOverrideRules = false;
+                if (lookAtSubNodes) {
+                    for (TreeNode parentNode : parentNodes) {
+                        dottedNames.putAll(getAllDottedNodes(parentNode.node));
```
pzygielo, line 211: https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/view/change-requests/job/PR-24185/2/console
```
[ERROR] /home/jenkins/agent/workspace/_test-using-jenkinsfile_PR-24185/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/SetCommand.java:[211,44] non-static method getAllDottedNodes(org.jvnet.hk2.config.Dom) cannot be referenced from a static context
```

OndroMih: I'm sorry, I should have checked better. I rebased against `master`, that broke my build but I forgot to rebuild again. Should be fixed by my most recent commit

dmatej (approved):

PR #23874, opened 2022-03-29
Integrate Shoal 3.0.0-M1

arjantijms:
Signed-off-by: Arjan Tijms <arjan.tijms@gmail.com>

PR #24188, opened 2022-11-19
Integrate Shoal 3.0.0

arjantijms:

### DIFF
--- a/nucleus/parent/pom.xml
+++ b/nucleus/parent/pom.xml
@@ -129,7 +129,7 @@
         <glassfish-management-api.version>3.2.3</glassfish-management-api.version>
         <pfl.version>4.1.2</pfl.version>
         <gmbal.version>4.0.3</gmbal.version>
-        <shoal.version>3.0.0-M1</shoal.version>
+        <shoal.version>3.0.0</shoal.version>
         <ha-api.version>3.1.13</ha-api.version>
         <logging-annotation-processor.version>1.9</logging-annotation-processor.version>
         <command-security-plugin.version>1.0.15</command-security-plugin.version>


nucleus/parent/pom.xml
```diff
@@ -129,7 +129,7 @@
         <glassfish-management-api.version>3.2.3</glassfish-management-api.version>
         <pfl.version>4.1.2</pfl.version>
         <gmbal.version>4.0.3</gmbal.version>
-        <shoal.version>3.0.0-M1</shoal.version>
+        <shoal.version>3.0.0</shoal.version>
```
pzygielo, line 132: With https://ci.eclipse.org/glassfish/job/all-projects_1_build-and-stage/85/parameters/ - this was (again) not pushed to GitHub.
I'm :-1:

pzygielo: Previously:
- https://github.com/eclipse-ee4j/glassfish/pull/23874#issuecomment-1257173752

dmatej: Hmmmmm, aha! Something is missing here! https://github.com/eclipse-ee4j/glassfish-shoal/tags

arjantijms: > this was (again) not pushed to GitHub.

I was waiting for it to pass the tests on the build.

pzygielo: I see this: https://github.com/eclipse-ee4j/glassfish-shoal/releases/tag/3.0.0-RELEASE

This is not the tag created by the job. Compare tag for 2.0.0 which is annotated one.
This points to the code in 3.0.0-SNAPSHOT version.

There (https://github.com/eclipse-ee4j/glassfish-shoal/branches) is still no release branch created by the job.

I don't like it.

arjantijms: Maybe the real alternative is to see if we can remove Shoal completely from GlassFish? But this will take some time I think.

dmatej: > Maybe the real alternative is to see if we can remove Shoal completely from GlassFish? But this will take some time I think.

We can't, it is important for the cluster. We could use hazelcast, but that would need a lot of effort and currently I have doubts it would help anything.

pzygielo: > Maybe the real alternative is to see if we can remove Shoal completely from GlassFish? But this will take some time I think.

Until then I suggest
- [ ] drop staged shoal 3.0.0 (orgglassfish-1233)
- [ ] delete 3.0.0-RELEASE tag from GitHub
- [ ] build shoal again, but this time check push to github box
- [ ] restart this Jenkins perhaps

arjantijms: >  but this time check push to github box

I wonder, why do we have this box? Maybe because it once broke, and we had to stage anyway without the GitHub tags? 3.0.0 is already in central, so we would need to do a 3.0.1. But what changes would 3.0.1 need?

arjantijms: If the push to GitHub works (the comment says it doesn't), then we don't need it. I'll remove the box and try a 3.0.1 then.

pzygielo: > 3.0.0 is already in central

pzygielo: > If the push to GitHub works (the comment says it doesn't),


It does: https://ci.eclipse.org/glassfish/job/all-projects_1_build-and-stage/84/

arjantijms: 3.0.1 running: https://ci.eclipse.org/glassfish/view/All%20Projects/job/all-projects_1_build-and-stage/86/console

arjantijms: It looks like the box description was wrong and it does work, hence the box isn't needed really or should default to true. I'll update it.

pzygielo: I have added recently the comment for the field:
> Next SNAPSHOT version.
> Note: `-SNAPSHOT` will be added by script so do not include it here.

but observe: https://github.com/eclipse-ee4j/glassfish-shoal/commit/6dee54ee51839a664171133536d2cd821d391c48

arjantijms: Okay, 3.0.1 has been done. The SNAPSHOT-SHAPSHOT can be easily fixed in that branch. I'll close this PR.

PR #24151, opened 2022-11-04
Fix AdminGUI for GF7

avpinchuk:
Fixes an incompatibility administrator's console with GF7.

Related PR's
eclipse-ee4j/glassfish-jsftemplating#82
eclipse-ee4j/glassfish-woodstock#1421

Fixes eclipse-ee4j/glassfish#23782

Signed-off-by: Alexander Pinchuk <alexander.v.pinchuk@gmail.com>

PR #24191, opened 2022-11-20
Add javadoc jar for jackson repackaged

arjantijms:


dmatej (approved):
Hmmmm, wouldn't be more simple to download (dependency plugin) -> deploy with new name? How we do that with other repackaged modules?

dmatej:
Caused by #24151

PR #24198, opened 2022-11-28
Attempt to stabilize log rotation integration test

avpinchuk:
`AsadminLoggingITest.rotateLog()` sporadically fails (1 run of 10 in my case) due to `server.log` is not empty after rotation (may contains up to 10 lines).

Signed-off-by: Alexander Pinchuk <alexander.v.pinchuk@gmail.com>


dmatej (approved):
Except the order of imports I like it!
But while we don't have any checkstyle rule yet (order by alphabet, static first OR last) ... :-)

avpinchuk:
@dmatej, I have some questions about admin rest tests. Where can we discuss them?

PR #24200, opened 2022-11-28
Revisited stop/start of GlassFish servers

### LINKED ISSUE #24187: Make stop-domain more robust
In years we all experiences some issues with stop-domain command. They are not well reproducible, often depend on used OS or filesystem, speed of the CPU ... with weird consequences. The worst is that sometimes the command says the domain stopped, but it did not (yet).

Some examples I catched recently will follow here.

### PID.PREV could not be found

* Is this ok or not? Is the domain running?
* Catched in Eclipse's TCK run with GF7M9.
```
+ /root/vi/glassfish7/glassfish/bin/asadmin stop-domain --kill
Picked up JAVA_TOOL_OPTIONS: -Xmx6G
Can not find the process ID of the server.  It is supposed to be here: /root/vi/glassfish7/glassfish/domains/domain1/config/pid.prev.  Unable to kill the process.
Command stop-domain failed.
```

### Could not stop, inaccessible admin cache
* Timeout, why?
* What is admin cache and why it cannot be read?
* Catched in Eclipse's TCK run with GF7M9.
```
+ /root/ri/glassfish7/glassfish/bin/asadmin --user admin --passwordfile /root/admin-password.txt stop-domain
Picked up JAVA_TOOL_OPTIONS: -Xmx6G
Cannot read admin cache file for cache/localhost_4848/__locations
Cannot read admin cache file for cache/localhost_4848/stop-domain
Waiting for the domain to stop ...........................................................
Timed out (60 seconds) waiting for the domain to stop.
Command stop-domain failed.
```

### Domain stopped, but another start fails, port still used.

* Seen on private Jenkins
* Read the code and check the shutdown correctly releases all ports.

dmatej:
* Fixes https://github.com/eclipse-ee4j/glassfish/issues/24187

- Added host name validation with timeout
  - 5 seconds is an extreme value for the local server
 
- Honoring command-line arguments so they are not overwritten from domain.xml
  which could be updated since startup and doesn't contain correct values
  usable to stop the server.

- WARN: ServerDirs.serverName doesn't contain the server name, but instance
  name or domain name. Probably terminology yet from SunONE.
- Replaced JPS and JSTACK usages by ProcessHandle usages

- WARN: restart-domain leaves zombies on some systems
  - visible on Jenkins TCK
  - these processes are marked as <defunct> and operating system decides when
    they will be cleaned up. It probably relates to Jenkins which reads IO
    from the container, so stese streams remain always open.

- Fixed NPE in ClassPathUtils

- Added AfterAll to ClusterITest

- Added AS_TRACE support to admin-tests - bound to FINEST in tests

- UnknownHostException is now an error for asadmin, it doesn't mean that
  the server is stopped.

- Command line --host has higher priority than domain.xml and default value
  - However it is not always true, ie restart-domain must use the arg value
    for stop, but domain.xml value for start.

- Generalized waitFor, added time measurement.

- Using addSuppressed to keep all errors (visible with AS_TRACE)

- Fixed situation when there was no asadmin truststore - load null and use empty

- JavaClassRunner revisited
  - Avoid "oversanitization", use the same value as the last time

- ServerDirs revisited (quickly)
  - Added getter getLastPidFile and javadoc explaining the difference
  - pid file uses pid of running server, is deleted when server stops
  - pid.prev file uses pid of last started server, is not deleted, just overwritten

- Using Thread.yield when waiting rather than sleep
  - More assertive, even aggressive
  - Increases probablity of detection of misbehavior if there's any
  - Faster, we don't wait if not necessary

- Directory for GF client can be set by the GF_CLIENT_DIR env property
  - by default it is still in ${user.home}/.gfclient
  - Fixes issue with tests which used user's home, which affected results
  - Fixes TCK on build server which wasn't permitted to write to user's home (root)

- New ENV options
  - AS_STOP_TIMEOUT to override default timeout for the stop domain command.
  - AS_START_TIMEOUT to override default timeout for the start domain command.

- Fixed asadmin logging - ignored exceptions

- Fixed StatusGenerator + test
  - the fix is not complete, but at least it doesn't crash with NPE
  - Seen somewhere in logs, originally thought that I broke it, but it was broken for a long time

- Revisited all user.home usages


dmatej:
Status: I am yet fixing the failure of ejb_group_3, seems imq stops asynchronously to domain and can still run when we try to tar all files.

arjantijms (approved):

pzygielo:
- https://github.com/eclipse-ee4j/glassfish/issues/19371

PR #24201, opened 2022-11-29
Authorization Constraints Ignored When Using Path Traversal Penetration Using Default Virtual Module

Pandrex247:
Security fix, from [Payara](https://github.com/payara/Payara/pull/6079)

The original investigation was done in Payara code so some things may not line up _perfectly_, but should be _near enough_

It was possible to circumvent JACC authentication using a ./ traversal, since the request normalisation which normally stops these attacks actually happens after the JACC check has occurred. So while the path is normalised as expected ([here](https://github.com/eclipse-ee4j/glassfish/blob/7.0.0-M10/appserver/web/web-core/src/main/java/org/apache/catalina/core/StandardContextValve.java#L244)), the actual JACC check has already occurred and passed when it would fail against the normalised path ([here](https://github.com/eclipse-ee4j/glassfish/blob/7.0.0-M10/appserver/web/web-core/src/main/java/org/apache/catalina/authenticator/AuthenticatorBase.java#L514) - this kicks off a call to `AuthorizationService.isPermitted`).

This PR reactivates the old normalisation in the CoyoteAdapater which was commented out many many moons ago without explanation when integrating a Grizzly update with the forked Tomcat in the days of GlassFish 2 & 3.

What happens now is that if a request which would fail its normalisation checks occur, a HTTP 400 request is returned. This is, from a quick scan of their code, in line with how Tomcat / Catalina functions today.

The criteria for failing normalisation are:
* Request not starting with a "/"
* Request with backslashes in (if a property is not enabled: `org.glassfish.grizzly.tcp.tomcat5.CoyoteAdapter.ALLOW_BACKSLASH`)
* Request with null bytes in
* Requests which try to escape the context

Reproduction steps:
* Build attached application: [Helloworld.zip](https://github.com/eclipse-ee4j/glassfish/files/10114708/Helloworld.zip)
* Deploy application: `asadmin  --name Helloworld /path/to/Helloworld/target/Helloworld-1.0.war`
* Set as default module: `asadmin set configs.config.server-config.http-service.virtual-server.server.default-web-module=Helloworld`
* Poke endpoint with valid path, should receive 403: `echo -e "GET /test/index.html HTTP/1.1\nHost: localhost\r\n\r\n" | nc localhost 8080`
* Poke endpoint with naughty path, unpatched you're allowed in, patched you should get a "shoo": `echo -e "GET /test/index.html HTTP/1.1\nHost: localhost\r\n\r\n" | nc localhost 8080`

I don't seem to be able to build GlassFish locally atm, there seems to be some issue with the ant task which generates the Full Profile domain (seems `microprofile-jwt-auth-api.jar` is missing?), but I tested using the tried-and-true "build the jar and drop it in the modules dir" of GlassFish 7.0.0-M10 and it seems to fix the issue as it does in Payara.

### DIFF
--- a/appserver/web/web-core/src/main/java/org/apache/catalina/connector/CoyoteAdapter.java
+++ b/appserver/web/web-core/src/main/java/org/apache/catalina/connector/CoyoteAdapter.java
@@ -329,6 +329,13 @@ public class CoyoteAdapter extends HttpHandler {
             return false;
         }
 
+        // Normalize Decoded URI
+        if (!normalize(decodedURI)) {
+            res.setStatus(400);
+            res.setDetailMessage(("Invalid URI"));
+            return false;
+        }
+
         if (compatWithTomcat || !v3Enabled) {
 
             // Set the remote principal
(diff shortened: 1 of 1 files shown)


pzygielo:
I feel uncomfortable looking at exceptions used for flow control...

appserver/web/web-core/src/main/java/org/apache/catalina/connector/CoyoteAdapter.java
```diff
@@ -687,10 +696,10 @@ private static boolean normalizeChars(MessageBytes uriMessageBytes) {
             }
             // Prevent from going outside our context
             if (index == 0) {
-                return false;
+                throw new IOException("Request traversed outside of allowed context");
             }
             int index2 = -1;
-            for (pos = start + index - 1; (pos >= 0) && (index2 < 0); pos--) {
+            for (pos = start + index - 1; (pos >= 0) && (index2 < 0); pos --) {
```
pzygielo, line 702: I'd rather keep `--` operator not separated, like in new line 604 is.

Pandrex247: That will have been a cherry-pick reformatting which I missed, will fix.

Pandrex247:
Reworked to not use exceptions

smillidge (approved):

PR #24209, opened 2022-12-20
Complete 7.0.0 release

pzygielo:


pzygielo:
Interesting
```
08:35:13  [ERROR] Failed to execute goal on project common-util: Could not resolve dependencies for project org.glassfish.main.common:common-util:glassfish-jar:7.0.1-SNAPSHOT: Could not find artifact org.osgi:osgi.enterprise:jar:7.0.1-SNAPSHOT -> [Help 1]
```

pzygielo:
It seems that build/release job changes everything it likes to consider as version to be changed, even for external dependencies...

Haven't checked everything, but this one for now is restored in
- #24210

dmatej:
That's why I created updateVersion.sh, but it is not used on Jenkins. Last time I fixed these version changes and created the script. But maybe it would need to recheck if it still does everything correctly.

PR #24150, opened 2022-11-04
Fix null pointer in ComponentDefinition

OndroMih:
Fixes an exception in case of scanning nested classes and some other types of classes. 

Reproducible with a nested class in an app, deployed on GF embedded using ScatteredArchive pointed to target/classes. Here's an example app: https://github.com/OmniFish-EE/Presentation-Piranha-from-GlassFish/tree/main/embedded-glassfish-app

Here's an exception when running the reproducer - it's not clear from Javadoc but calling `contains` with a null argument on an unmodifiable Set results in an exception, at least on OpenJDK:

```
SEVERE: Exception during lifecycle processing
java.lang.NullPointerException
        at java.base/java.util.Objects.requireNonNull(Objects.java:208)
        at java.base/java.util.ImmutableCollections$SetN.contains(ImmutableCollections.java:937)
        at org.glassfish.apf.impl.ComponentDefinition.isExcludedFromAnnotationProcessing(ComponentDefinition.java:145)
        at org.glassfish.apf.impl.ComponentDefinition.constructClassList(ComponentDefinition.java:87)
        at org.glassfish.apf.impl.ComponentDefinition.<init>(ComponentDefinition.java:60)
        at org.glassfish.apf.impl.JavaEEScanner.getComponentInfo(JavaEEScanner.java:46)
        at org.glassfish.apf.impl.AnnotationProcessorImpl.process(AnnotationProcessorImpl.java:155)
        at org.glassfish.apf.impl.AnnotationProcessorImpl.process(AnnotationProcessorImpl.java:115)
        at com.sun.enterprise.deployment.archivist.Archivist.processAnnotations(Archivist.java:614)
        at com.sun.enterprise.deployment.archivist.Archivist.readAnnotations(Archivist.java:462)
        at com.sun.enterprise.deployment.archivist.Archivist.readAnnotations(Archivist.java:448)
        at com.sun.enterprise.deployment.archivist.Archivist.readRestDeploymentDescriptors(Archivist.java:421)
        at com.sun.enterprise.deployment.archivist.Archivist.readDeploymentDescriptors(Archivist.java:397)
        at com.sun.enterprise.deployment.archivist.Archivist.open(Archivist.java:274)
        at com.sun.enterprise.deployment.archivist.Archivist.open(Archivist.java:283)
        at com.sun.enterprise.deployment.archivist.Archivist.open(Archivist.java:244)
[...]

PR #24214, opened 2022-12-21
Add test for class with no canonical name

pzygielo:


appserver/common/annotation-framework/src/main/java/org/glassfish/apf/impl/ComponentDefinition.java
```diff
@@ -41,12 +43,12 @@
  */
 public class ComponentDefinition implements ComponentInfo {
 
-    private static final Set<String> EXCLUDED_FROM_ANNOTATION_PROCESSING = Set.of(
+    private static final Set<String> EXCLUDED_FROM_ANNOTATION_PROCESSING = Collections.unmodifiableSet(new HashSet<>(Set.of(
```
dmatej, line 46: Set.of already returns unmodifialble collection, see javadoc ;)

appserver/common/annotation-framework/src/main/java/org/glassfish/apf/impl/ComponentDefinition.java
```diff
@@ -142,11 +142,7 @@ static boolean isExcludedFromAnnotationProcessing(Class<?> clazz) {
         if (clazz.getPackage().getName().startsWith("java.lang")) {
             return true;
         }
-        if (clazz.getCanonicalName() == null) {
```
dmatej, line 145: Maybe it's just nanos or so, but I believe that the null check is faster than contains(null)...?

appserver/common/annotation-framework/src/main/java/org/glassfish/apf/impl/ComponentDefinition.java
dmatej: Aha, it is to allow contains(null), but ...

appserver/common/annotation-framework/src/main/java/org/glassfish/apf/impl/ComponentDefinition.java
pzygielo: `null`-check was introduced for other than performance reasons, and I'd like to get rid of it.

I like the pre-#24150 version better - so I'd like to restore it.
But if I do not share this liking with anyone, perhaps #24213 could be considered separately?

appserver/common/annotation-framework/src/main/java/org/glassfish/apf/impl/ComponentDefinition.java
pzygielo: [JDK-8265905: ImmutableCollections: Querying for 'null' should return 'false' instead of NPE](https://bugs.openjdk.org/browse/JDK-8265905) was closed as won't fix, so the only hope is for [JDK-8267823: Document NPE on immutable collections query methods](https://bugs.openjdk.org/browse/JDK-8267823). But it might take 10 LTSes to get there.

appserver/common/annotation-framework/src/main/java/org/glassfish/apf/impl/ComponentDefinition.java
dmatej: Ok, ok, it's just my first idea when I asked myself why are you doing that :-)
Tests are important and they explain reasons well, it's ok to have both here.

dmatej (approved):

pzygielo:
Thanks for checking.

PR #24216, opened 2022-12-22
Make evaluateNormalizedPathWithSinglePoint package-accessible and test it

pzygielo:


dmatej (approved):
One idea ... could we use Path.normalize for URL paths instead of own impl? I believe it should be possible, I think even used it on some project ...

arjantijms (approved):

Issue #24219, opened 2022-12-23
Some modules fail `command-security-maven-plugin:check`


pzygielo:
## Environment Details
* GlassFish Version (and build number): 7.0.1-SNAPSHOT (a364892e1)
* JDK version: 17
* OS: GNU/Linux
* Database: n/a

----------

## Problem Description
Some modules fail `command-security-maven-plugin:check`

Example:
```
[WARNING] Following command classes neither provide nor inherit authorization: [org.glassfish.jms.admin.cli.ChangeMasterBrokerCommand, org.glassfish.jms.admin.cli.ConfigureJMSCluster, org.glassfish.jms.admin.cli.CreateJMSDestination, org.glassfish.jms.admin.cli.CreateJMSHost, org.glassfish.jms.admin.cli.CreateJMSResource, org.glassfish.jms.admin.cli.DeleteJMSDestination, org.glassfish.jms.admin.cli.DeleteJMSHost, org.glassfish.jms.admin.cli.DeleteJMSResource, org.glassfish.jms.admin.cli.FlushJMSDestination, org.glassfish.jms.admin.cli.GetJmsPhysicalDestinationCommand, org.glassfish.jms.admin.cli.JMSPing, org.glassfish.jms.admin.cli.ListJMSDestinations, org.glassfish.jms.admin.cli.ListJMSHosts, org.glassfish.jms.admin.cli.ListJMSResources, org.glassfish.jms.admin.cli.UpdateJmsPhysicalDestinationCommand]
```

[//]: # "Describe the bug in detail highlighting current behavior vs expected behavior"
[//]: # "State if the problem is easily reproducible or happens intermittently"
[//]: # "Include stack traces or command outputs"

## Steps to reproduce
Build with `-Dcommand.security.maven.plugin.isFailureFatal=true`.

## Impact of Issue
IDK yet.

PR #24220, opened 2022-12-23
Fail on command-security failures, but excuse current violations

pzygielo:
- #24219


pzygielo:
At this moment I don't know if the violations should be inspected and fixed or the plugin execution to be dropped.

dmatej (approved):
The change: The configuration was pushed down, so current issues are tolerated, new issues are prevented. Good!

PR #24222, opened 2022-12-26
Move method to test

pzygielo:
1. This change moves method from production code to unit test. This saves about 2285 bytes in `.class` file.
1. I'm not sure if this (persistence/entitybean-container) is used in runtime, but it's part of distribution. Due to that:
   1. ~I'm not changing dependencies (e.g. to add JUnit), but am falling back to execute test by surefire [almost] directly.~
   1. I'm not changing test method (e.g. to replace comments with explicit assertions).
1. If the class can be turned to not be executable - removing that empty now `main` method could save about 118 additional bytes.

### DIFF
--- a/appserver/persistence/entitybean-container/pom.xml
+++ b/appserver/persistence/entitybean-container/pom.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
+    Copyright (c) 2022 Contributors to the Eclipse Foundation. All rights reserved.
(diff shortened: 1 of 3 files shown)


dmatej:
If you would continue more and change the test to be a real JUnit5 test, it would be perfect. This seems to me like some old "zombie test", maybe even useful, but not executed for years. Usages of the class (fulltext search):
image
Maybe we could add even some logs, so we could later see at least if it is touched by some TCK tests ...

dmatej (approved):

PR #24225, opened 2023-01-07
New XmlSec4 without mandatory SLF4J

### LINKED ISSUE #24138: slf4j-api not working with Glassfish 6
## Environment Details
* GlassFish Version (and build number): 6.2.5  (build 6.x-b66-g0159b68 2022-02-12T17:39:59+0000).

* JDK version:11
* OS: Linux Ubuntu
----------
## Problem Description
I am trying to use **slf4j-api** with Logback on Glassfish 6.2.5.

I copied logging  jars to `glassfish/lib` folder as I did with older versions (I come from Payara 4):

- slf4j-api-2.0.3.jar
- logback-classic-1.4.4.jar
- logback-core-1.4.4.jar

When I deploy the web app, the logging calls through slf4j logger calls are either:

- a) ignored, as if they are a NO-OP (ie.`  log.info("OK");`) .

- b) If I use newer methods from slf4j-api 2.x only (like ` log.atInfo().log("OK:");`) I get 

> jakarta.ejb.TransactionRolledbackLocalException: Exception thrown from bean: java.lang.NoSuchMethodError: 'org.slf4j.spi.LoggingEventBuilder org.slf4j.Logger.atInfo()'

Which indicates the jar is not even loaded. 

Is this expected and  how to ensure my slf4j-api jar is being loaded?

dmatej:
### Dependencies

* WSIT/Metro: https://github.com/eclipse-ee4j/metro-wsit/pull/421
* XmlSec: https://github.com/apache/santuario-xml-security-java/pull/192 (originally https://github.com/apache/santuario-xml-security-java/pull/101 , closed by my mistake)

### Consequences

* Fixes https://github.com/eclipse-ee4j/glassfish/issues/24138 , no dependency on SLF4J, no need to repackage dependencies

### Validation

* Passed GlassFish build: https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/job/PR-24225/
* Passed the TCK build: https://ci.eclipse.org/jakartaee-tck/job/jakartaee-tck/job/10.0.x/ - see GF-PR-24225 for builds


arjantijms (approved):

avpinchuk (approved):

appserver/extras/embedded/all/pom.xml
```diff
@@ -1382,6 +1382,12 @@
             <groupId>org.apache.santuario</groupId>
             <artifactId>xmlsec</artifactId>
         </dependency>
+        <!-- xmlsec depends on commons-codec -->
```
pzygielo, line 1385: Shouldn't it come from `xmlsec` without that declaration? Why do we need to have it explicitly?

dmatej: It is a workaround to avoid downloading all dependencies when they are repackaged inside. 
In the future we might consider also to avoid repackaging dependencies except glassfish.main, but everything has consequences.
Also Maven 4 will separate pom used for build and pom used for deployment. Originally we discussed to use some maven plugin for that with Maven 3 ... again ... has consequences.

dmatej: Yet one thing about Apache Commons - I always loved them, but it seems I need them less and less as I use newer JDK versions. We could consider if the dependency couldn't be removed from our dependencies.

arjantijms: > Originally we discussed to use some maven plugin for that with Maven 3 ... again ... has consequences.

The maven flatten plug-in, right? I started to use it in several projects.

PR #24226, opened 2023-01-08
GlassfishUrlClassloader and toString

dmatej:
Extracted classloader changes from https://github.com/eclipse-ee4j/glassfish/pull/24225

* ClassLoader changes - all GF classloaders now have toString; it helped me when I had issues with OSGI; URLClassLoader printed just a classname and hashcode, now it prints managed URLs.
* WebappClassLoader, StandardManager small revision - fixed some stacktraces in logs I have seen along the path, produced by ugly workarounds in the code.

### DIFF
--- a/nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/GlassfishBootstrapClassLoader.java
+++ b/nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/GlassfishBootstrapClassLoader.java
@@ -20,16 +20,19 @@ import java.io.File;
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.net.URLClassLoader;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
+import java.util.stream.Collectors;
 
 
 /**
  * This classloader is located between OSGI classloader and system JDK extensions classloader,
  * so it can provide early initialized libraries (before OSGI startup, so even OSGI can use them).
  * <p>
- * Examples
+ * List:
  * <ul>
  * <li>GlassFish Java Util Logging Extension
  * <li>Grizzly NPN API
@@ -37,7 +40,7 @@ import java.util.List;
  *
  * @author David Matejcek
  */
-public class GlassfishBootstrapClassLoader extends GlassfishUrlClassLoader {
+public class GlassfishBootstrapClassLoader extends URLClassLoader {
 
     /**
      * Initializes the classloader.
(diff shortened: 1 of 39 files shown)


nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/GlassfishBootstrapClassLoader.java
```diff
@@ -20,24 +20,27 @@
...
+ * List:
  * <ul>
  * <li>GlassFish Java Util Logging Extension
  * <li>Grizzly NPN API
  * </ul>
  *
  * @author David Matejcek
  */
-public class GlassfishBootstrapClassLoader extends GlassfishUrlClassLoader {
+public class GlassfishBootstrapClassLoader extends URLClassLoader {
```
pzygielo, line 43: Could this extend `o.g.c.util.GlassfishUrlClassLoader` to share `toString`?

dmatej: No, because this CL is specific (bootstrap) - it must not have dependencies from it's children classloaders.

pzygielo: I don't understand, so checking it with
- https://github.com/eclipse-ee4j/glassfish/pull/24227

Can you predict what will fail?

dmatej: Simply don't do that. You can't even add logging to the GFurlcl, because it would go against initialization. See GlassFishMain.

pzygielo: Could `o.g.c.util.GlassfishUrlClassLoader` have `public static String toString(URLClassLoader)` - to be used by both `GlassfishUrlClassLoader.toString()` and `GlassfishBootstrapClassLoader.toString()` to keep the format the same?

dmatej: That makes again another common dependency for both while we simply want to keep them isolated. The bootstrap part is at the beginning a "semiisolated" JVM starting another. That's the reason of all that reflection in GlassfishMain.
And it still causes some pain on some systems. I don't like code duplication too, but here it has some reason and this is really short piece of code.
Maybe it would be possible to make that separation more clear somehow, on the other hand ... I see 100+ high priority security issues in Snyk ... I also created several issues for WebappCL, seen in logs, that's my direction here.

hs536 (approved):
LGTM (seems helpful!)

btw, could you commit important changes and code cleaning/formatting separately from next time? Otherwise, reviewers have to look carefully at all changes to avoid fatal oversight...

PR #24233, opened 2023-01-11
Use archive suffix to determine the compression program

pzygielo:
Currently observed:
```
$ wget https://ci.eclipse.org/glassfish/view/GlassFish/job/glassfish_build-and-test-using-jenkinsfile/job/master/907/artifact/jdbc_all-results.tar.gz
$ gzip --decompress web_jsp-results.tar.gz 

gzip: web_jsp-results.tar.gz: not in gzip format
$ file web_jsp-results.tar.gz 
web_jsp-results.tar.gz: POSIX tar archive (GNU)
```

### DIFF
--- a/appserver/tests/common_test.sh
+++ b/appserver/tests/common_test.sh
@@ -1,5 +1,6 @@
 #!/bin/bash -ex
 #
+# Copyright (c) 2023 Contributors to the Eclipse Foundation. All rights reserved.
 # Copyright (c) 2017, 2020 Oracle and/or its affiliates. All rights reserved.
 #
 # This program and the accompanying materials are made available under the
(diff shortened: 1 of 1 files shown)


dmatej:
Wow! :-)
Or explicitly use -z instead of -a?

dmatej (approved):

pzygielo:
I added `-z` in first commit pushed: 54627c774293a11e768197d1368db151603610f4.
But with `a` - if we decide to use different algorithm/filter there would be only one change needed - in the extension.
It will also support plain `.tar`.
I didn't chose that though, as I don't want to track usage of those `*results` archives (now).

PR #24246, opened 2023-01-17
Issue #24228 Fixed shutdown monitoring

### LINKED ISSUE #24228: The tar command complains that osgi cache is still changing on stopped domain
Sometimes build fail when they are tarring domain directory after stop-domain. It maybe caused by a resource leak OR by simple fact that stop-domain ends too quickly after it loses all possible ways to check if the domain is still running - but the domain JVM is maybe still flushing streams to a file system.

I would prefer to ensure all streams close before this state. The least preferred option is to add pause after stop-domain command finished (how long? it depends on file system latency and speed of the whole system - no reliable answer, just probability).

```
tar: /home/jenkins/agent/workspace/nd-test-using-jenkinsfile_master/glassfish7/glassfish/domains/domain1/osgi-cache/felix: file changed as we read it
```

One example (deployment_all): https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/job/PR-24225/3/execution/node/413/log/

dmatej:
- The pid file was removed twice while it is used to check if the instance is still running.
- faster check to increase probability of issues if there are some - and also to be really faster :-)
- Local testing: start: 2200-2500 ms, stop: 30-1200 ms (the slower value after undeploy in tests).
- Added log for failures.
- The javadoc was not right, it is not possible in current Java versions with the current GlassFish implementation, but it is possible to avoid hooks by using SIGKILL signal. But that isn't an issue any more, because since GF7 we use ProcessHandles to check if the process is running, while old versions used commands of the operating system.
- However, I am still not sure if this is enough to resolve the issue, because Felix really writes to the osgi cache until the real end. But JVM hooks are executed AFTER parallely executed application hooks and the DeleteOnExitHook should be the last one. That should remove the pid file. Then asadmin command checks if it exists or not, if it exists, it loads it's content and checks handles if the process is alive or not. On startup it checks the same, so there is no problem.
- Tested CTRL+C, SIGKILL, pid, stop, --force, --kill, restart-domain, everything works now.

Copy-pasted comment from the DeleteOnExitHook in the JDK17 (same at least from JDK8, I don't see older in Eclipse):
```
        // DeleteOnExitHook must be the last shutdown hook to be invoked.
        // Application shutdown hooks may add the first file to the
        // delete on exit list and cause the DeleteOnExitHook to be
        // registered during shutdown in progress. So set the
        // registerShutdownInProgress parameter to true.
```

The last commit fixed the issue (tested locally, let's see what Jenkins says).
- For local domains: domain stopped if (it is not listening on port AND (the pid file doesn't exist OR the process with the pid is not alive).
- For remote domains: domain stopped if it is not listening. Therefore when we stop remote domain, it stops listening, but then still does a cleanup for some time. But we cannot see it, for us it stopped.

This PR finishes what started here: https://github.com/eclipse-ee4j/glassfish/pull/24200 (I hope I did not miss any use case).

### DIFF
--- a/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/StopDomainCommand.java
+++ b/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/StopDomainCommand.java
@@ -167,9 +167,10 @@ public class StopDomainCommand extends LocalDomainCommand {
     protected void doCommand() throws CommandException {
         // run the remote stop-domain command and throw away the output
         RemoteCLICommand cmd = new RemoteCLICommand(getName(), programOpts, env);
+        File watchedPid = isLocal() ? getServerDirs().getPidFile() : null;
         try {
             cmd.executeAndReturnOutput("stop-domain", "--force", force.toString());
-            waitForDeath();
+            waitForDeath(watchedPid);
         } catch (Exception e) {
             // The domain server may have died so fast we didn't have time to
             // get the (always successful!!) return data.  This is NOT AN ERROR!
@@ -177,7 +178,6 @@ public class StopDomainCommand extends LocalDomainCommand {
             if (kill && isLocal()) {
                 try {
                     File prevPid = getServerDirs().getLastPidFile();
-                    File watchedPid = getServerDirs().getPidFile();
                     ProcessUtils.kill(prevPid, watchedPid, Duration.ofMillis(DEATH_TIMEOUT_MS), !programOpts.isTerse());
                     return;
                 } catch (Exception ex) {
@@ -190,14 +190,21 @@ public class StopDomainCommand extends LocalDomainCommand {
 
     /**
      * Wait for the server to die.
+     *
+     * @param watchedPid
      */
-    private void waitForDeath() throws CommandException {
+    private void waitForDeath(File watchedPid) throws CommandException {
         if (!programOpts.isTerse()) {
             // use stdout because logger always appends a newline
             System.out.print(Strings.get("StopDomain.WaitDASDeath") + " ");
         }
         final Duration timeout = Duration.ofMillis(DEATH_TIMEOUT_MS);
-        final Supplier<Boolean> deathSign = () -> !ProcessUtils.isListening(addr);
+        final Supplier<Boolean> deathSign;
+        if (isLocal()) {
+            deathSign = () -> !ProcessUtils.isListening(addr) && !ProcessUtils.isAlive(watchedPid);
+        } else {
+            deathSign = () -> !ProcessUtils.isListening(addr);
+        }
         final boolean dead = ProcessUtils.waitFor(deathSign, timeout, !programOpts.isTerse());
         if (!dead) {
             throw new CommandException(Strings.get("StopDomain.DASNotDead", timeout.toSeconds()));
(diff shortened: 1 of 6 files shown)


nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/StopDomainCommand.java
```diff
@@ -167,17 +167,17 @@ protected int dasNotRunning() throws CommandException {
     protected void doCommand() throws CommandException {
         // run the remote stop-domain command and throw away the output
         RemoteCLICommand cmd = new RemoteCLICommand(getName(), programOpts, env);
+        File watchedPid = isLocal() ? getServerDirs().getPidFile() : null;
```
pzygielo, line 170: First look - if this becomes `null`, won't `ProcessUtils.isAlive(watchedPid)` throw `NPEx` on `pidFile.exists()`?

nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/StopDomainCommand.java
```diff
@@ -190,14 +190,21 @@ protected void doCommand() throws CommandException {
...
-    private void waitForDeath() throws CommandException {
+    private void waitForDeath(File watchedPid) throws CommandException {
         if (!programOpts.isTerse()) {
             // use stdout because logger always appends a newline
             System.out.print(Strings.get("StopDomain.WaitDASDeath") + " ");
         }
         final Duration timeout = Duration.ofMillis(DEATH_TIMEOUT_MS);
-        final Supplier<Boolean> deathSign = () -> !ProcessUtils.isListening(addr);
+        final Supplier<Boolean> deathSign;
+        if (isLocal()) {
```
pzygielo, line 203: To avoid https://github.com/eclipse-ee4j/glassfish/pull/24246/files#r1072881827 - could `watchedPid` be checked here against `null` instead?

pzygielo: On the other hand `isLocal` has more meaning than `watchedPid != null`.

dmatej:
Admin Tests passed locally and I did not notice anything suspicious ... rerunning Jenkins ...

dmatej:
HA! Seems we should limit maximal number of concurrent builds - now if we push too often, Jenkins build crash, first due to timeouts, then OOME, my last build crashed even before git clone ... https://ci.eclipse.org/glassfish/computer/ ... seems 4 should be enough. That's why they used the k8s settings, it used hard coded limits, when they were not available, job was forced to wait. Now it starts because there's no limit, and crashes.

Sixth parallel job crashed.
image

pzygielo:
So the sponsoring of resource packs is just a waste of money?

arjantijms:
> So the sponsoring of resource packs is just a waste of money?

It obviously should not be a waste, as resource packs are essentially extra nodes  (virtual servers) right?

If there's something still limiting parallel builds to a single node, leaving the available nodes unused, we should take that up with Eclipse, specifically with Frederic Gurr I think.

arjantijms (approved):

dmatej:
Merged as a priority, it should help at least some of failures (jdbc_all, persistence_all, deployment_all sometimes failed).

PR #24243, opened 2023-01-15
Classloader updates

dmatej:
### WHY

- GlassFish classloaders were quite obsoleted, still based on Java 1.4 features.
- Known bugs
- It was hard to read the code.
- The implementation was slow.
- Logging did not tell much.
- Tomcat WebappClassLoader sources diverged, however are somehow old too. Meanwhile GF implemented other things ....

### Fixed bugs along the path

* #23769 Obsoleted code, current JDK provides solution. JdbcLeakPrevention replaced.
* ACCAgentClassLoader implemented GlassfishUrlClassLoader which is not possible as it is used as a bootstrap classloader. Introduced in 15332ce2e69b14a910 .
* ReferenceCleaner (originally inside WebappClassLoader) did not work, now does.
  * Tried to reset enums (forbidden)
  * Tried to access fields without permission (forbidden)
  * Noisy logging when it ignored excluded fields. (removed, not interesting)

### Changes

* WebappClassLoader is now "parallel-capable". Means basically declaring that and using `getClassLoadingLock(name)` for locking.
* WebappClassLoader is stateful with an attempt to strictly control what can be called
  * URLClassLoaders are Closeable
  * originally was just started and not started, while it wasn't checked.
  * now can be NEW/RUNNING/CLOSED, however some methods can be called both when NEW and RUNNING, which is not good, but it would need much more work to simplify it (give all parameters to constructor and use the CL until it is closed).
* Using System.Logger instead of JUL.
* Removed redundant logs, added new logs, set these and watch:
    ```
    org.glassfish.web.loader.WebappClassLoader.level=FINEST
    org.glassfish.web.loader.JarFileManager.level=FINEST
    org.glassfish.web.loader.ReferenceCleaner.level=FINEST
    ```
* Several new package-visible classes split from WebappClassLoader, each responsible for some segment.
  * JarFileManager manages JAR files under classloader's control
      * synchronized blocks replaced by ReentrantReadWriteLock
      * JarFile management moved inside to a special scheduled task which closes files when they are not used. This was done also in original implementation, but managed by the StandardContext and WebappLoader.
      * JarFile instances not closed on startup "to avoid peaks" as they were immediately open again.     
      * Both period and the timeout are configurable by the JVM option. Original timeout was 90 seconds, now period is 15 , timeout 60 seconds.
  * ReferenceCleaner serves to resolve possible CL leaks. If there are ThreadLocal values forgotten, it should set them to null, static fields, etc. The implementation was already there, but did not work. 
[...]

PR #24247, opened 2023-01-17
Classloader security - the fixed package.policy parameter

### LINKED ISSUE #24245: The package.definition property cannot work
When @hs536 noticed my dumb mistake in #24243 , we have found that the property is set via Security.setProperty, but CL parses it from System.getProperty.
As @pzygielo  correctly encouraged, I am creating this issue to fix it.

Usages (based on the mentioned PR)
image

dmatej:
* The setter - Catalina parses and sets the property to the Security.setProperty, then originally WebappCl parsed this property (from wrong class, System). I have connected them together this way; I expect that Catalina configures the property (via it's constructor) before any creation of the WebappCL.
* Fixes #24245 
* Depends on #24243 
* I tried to compare files in history and it seems to me that it was always broken ... but I may be wrong. 
* Despite the first commit fixed the setting in WebappClassLoader, the feature still did not work.
* The second commit added some logs and then I was able to follow what is happening and fix the behavior.
* Then failed six QuickLook tests, because the server chose a secured path which was broken (see reflections in the `ApplicationContextFacade`)

### DIFF
--- a/appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
+++ b/appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
@@ -39,7 +39,6 @@ import java.net.URL;
 import java.net.URLClassLoader;
 import java.security.AccessControlException;
 import java.security.AccessController;
-import java.security.AllPermission;
 import java.security.CodeSource;
 import java.security.Permission;
 import java.security.PermissionCollection;
@@ -52,7 +51,6 @@ import java.util.Collections;
 import java.util.Enumeration;
 import java.util.Hashtable;
 import java.util.List;
-import java.util.ResourceBundle;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
@@ -137,7 +135,6 @@ public final class WebappClassLoader extends GlassfishUrlClassLoader
     public static final boolean DELEGATE_DEFAULT = true;
 
     private static final Logger LOG = LogFacade.getSysLogger(WebappClassLoader.class);
-    private static final ResourceBundle rb = LogFacade.getLogger().getResourceBundle();
 
     private static final Function<String, String> PACKAGE_TO_PATH = pkg -> pkg.replace('.', '/');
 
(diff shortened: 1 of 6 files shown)


dmatej:
It is more complicated:

- Catalina loads catalina.properties which contains just an empty string
- Copies all content to the System.properties 
- CatalinaProperties is still in a state rather for Tomcat, should be updated for GlassFish, ie. use domain.xml, not catalina.properties which do not exist, therefore can be loaded from user's home.
  -  catalina.config sysprop is not set in GF.
  - catalina.home is set by Embedded.initDirs, but who knows how and if it is called on GF ...
  - if it is not set, it can use even catalina.properties from user's home
  - Perhaps to be refactored in some future PR, including removal of unused code.
- The issue is that part of the code is static, so when I added logs now, the package protection was ALWAYS disabled. Seems the static initialization in SecurityUtil was too fast.

Therefore at this moment this PR did not change anything. Locally, after some changes, I was able to enable it. Surprisingly the jacc tests passed again.

dmatej:
Ok, I did it, these are defaults on JDK17 :-)
```
[2023-01-20T19:06:46.932525+01:00] [GlassFish 7.0] [FINE] [] [jdk.event.security] [tid: _ThreadID=119 _ThreadName=AutoDeployer] [levelValue: 500] [[
SecurityPropertyModification: key:package.definition, value:sun.misc.,sun.reflect.,]]

[2023-01-20T19:06:46.932614+01:00] [GlassFish 7.0] [CONFIG] [] [jakarta.enterprise.web.core] [tid: _ThreadID=119 _ThreadName=AutoDeployer] [levelValue: 700] [[
appendSecurityProperty(property=package.access, packageList=)]]

[2023-01-20T19:06:46.932750+01:00] [GlassFish 7.0] [FINE] [] [jdk.event.security] [tid: _ThreadID=119 _ThreadName=AutoDeployer] [levelValue: 500] [[
SecurityPropertyModification: key:package.access, value:sun.misc.,sun.reflect.,]]
```

dmatej:
Not sure if this issue is the same in Payara, but I would recommend to check your code too, @smillidge . 

Basically just do a fulltext seach of the `package.definition` and `package.access` strings and look where and when it is set to/from Security.setProperty/getProperty and System.setProperty/getProperty. The result depends on the order of actions which was wrong in GlassFish. When you add and enable logs like we have in GF, run QuickLook tests and look into the server.log, it should be visible.

appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
```diff
@@ -341,6 +333,20 @@ public void setAntiJARLocking(boolean enable) {
     }
 
 
+    /**
+     * Enables checks for the package definition permissions.
+     *
+     * @param enable
+     */
+    public void setPackageDefinitionSecurityEnabled(boolean enable) {
+        if (enable && SECURITY_MANAGER == null) {
```
arjantijms, line 342: Note that eventually we'd have to remove all uses of the security manager

arjantijms (approved):

appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
dmatej: Correct, but while it is there, it should work :-)

PR #24248, opened 2023-01-18
Security fix: regexp catastrophic backtracking fix

dmatej:
- Based on SonarQube report: https://sonarcloud.io/project/security_hotspots?id=dmatej_glassfish&sinceLeakPeriod=true&tab=code
- Also fixed OneLine detection, there was one block more than required
- https://rules.sonarsource.com/java/RSPEC-5852

### DIFF
--- a/nucleus/glassfish-jul-extension/src/main/java/org/glassfish/main/jul/cfg/ConfigurationHelper.java
+++ b/nucleus/glassfish-jul-extension/src/main/java/org/glassfish/main/jul/cfg/ConfigurationHelper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 Eclipse Foundation and/or its affiliates. All rights reserved.
+ * Copyright (c) 2022, 2023 Eclipse Foundation and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -27,6 +27,7 @@ import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.logging.Level;
 import java.util.logging.LogManager;
+import java.util.stream.Collectors;
 
 import org.glassfish.main.jul.tracing.GlassFishLoggingTracer;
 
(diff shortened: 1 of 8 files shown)


dmatej:
Seems there are some issues with Jenkins - once OOME in compilation, now timeouts ... this branch is based on master, not on changes in classloaders and not on changes in stop-domain, so the issue must be really with Jenkins.

pzygielo:
> * Based on SonarQube report

As this is probably not published (I can't find the violations in https://sonarcloud.io/project/issues?resolved=false&id=dmatej_glassfish) - is it possible to share the link to rule triggered - https://rules.sonarsource.com? I'd like to read about that, please.

nucleus/glassfish-jul-extension/src/main/java/org/glassfish/main/jul/cfg/ConfigurationHelper.java
```diff
@@ -85,7 +88,7 @@ public class ConfigurationHelper {
         if (v == null || v.isEmpty()) {
             return Collections.emptyList();
         }
-        return Arrays.asList(v.split("[\\s]*,[\\s]*"));
+        return Arrays.asList(PATTERN_STR_TO_LIST.split(v, 0));
```
hs536, line 91: If the pattern matching only means split and trim, I thought it would be better to execute it clearly.
such as...
```java
return Arrays.stream(v.split(",")).map(String::trim).collect(Collectors.toList());
```

dmatej: I think that this would be much slower - it creates a stream object, parses the pattern for every value, then calls trim function for every result, and creates an array list.
While my version uses pregenerated Pattern and creates an ArrayList directly backed by the result array.

dmatej: But the truth is that for several values from a property file it doesn't matter.

dmatej: I have updated just the commit with your recommendation, thank you!

arjantijms (approved):

PR #24250, opened 2023-01-18
Use new pod template

pzygielo:


dmatej:
You are faster than my thoughts! :-)

pzygielo:
But - will it work? Frederic created this pod template for testing, not sure if it's even still there.

Maybe some kubernetes configuration removed in
- https://github.com/eclipse-ee4j/glassfish/pull/24144

should be restored, after all.

dmatej:
Maybe we should create an issue on gitlab and ask for help, first I would like to know what changed, why is it happening.

pzygielo:
> 16:20:38  Failure executing: POST at: https://10.30.0.1/api/v1/namespaces/glassfish/pods. Message: Forbidden!Configured service account doesn't have access. Service account may have been revoked. pods "centos-8-4cpu-452kr-z5ck3" is forbidden: minimum cpu usage per Pod is 200m, but request is 100m.

Jenkinsfile
```diff
@@ -129,7 +129,11 @@ pipeline {
   stages {
 
     stage('build') {
-      agent any
+      agent {
+        kubernetes {
+          inheritFrom 'centos-8-6gb'
```
pzygielo, line 134: This might be xtext-specific template, as it seems to be defined exclusively in https://github.com/eclipse-cbi/jiro/blob/49a71c613efd7cfe86ddefe34bbd72bedc603c31/instances/modeling.tmf.xtext/config.jsonnet#L43

dmatej: Btw, there was a note in our old k8s, I don't remember who added it, maybe me, maybe it was there even before. Labels are deprecated, we should use inheritFrom too. These settings were for the "root" machine, build machines had higher requests. I tested it to find out where is the maximum for limis, and then tried to find optimal shares for jobs. With additional resource packs perhaps those limits moved too.

```
      # fixes random failure: minimum cpu usage per Pod is 200m, but request is 100m.
      # affects performance on large repositories
      limits:
        memory: "1200Mi"
        cpu: "300m"
      requests:
        memory: "1200Mi"
        cpu: "300m"
```

pzygielo:
I hope, with https://www.eclipsestatus.io/incidents/j724xpdbz3sb resolved, this is no longer needed.

PR #24252, opened 2023-01-21
Jenkinsfile maintenance

dmatej:
Digging some info for https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2544

Based on https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2102 there should be 17 resource packs assigned to GF, but I have no idea where to check the current state.

What I see is that when we did [this](https://github.com/eclipse-ee4j/glassfish/commit/1baa948c13b968269f725a100f774eba897fd534) we requested higher limits for CPU+MEM, so now we can run more jobs in parallel with less memory, but for some unknown reason the performance decreased much later, and not just in tests:
```
... BASIC CFG (see an interesting comment):
      limits:
        memory: "12Gi"
        cpu: "8000m"
      requests:
        memory: "7Gi"
        cpu: "4000m"

... FOR MVN tests:
          resourceRequestMemory: "7Gi",
          resourceRequestCpu: "2650m"

... FOR ANT tests:
          resourceRequestMemory: "4Gi",
          resourceRequestCpu: "2650m"
```

Now we have (found in logs):
```
       limits:
         cpu: "2000m"
         memory: "4096Mi"
       requests:
         cpu: "1000m"
         memory: "4096Mi"
```

I remember I was tuning those values and reading documentation, but finally set those values after some experiments. Then, later, @pzygielo replaced that by more simple configuration, which originally had much better results by the end of the year 2022. Now it doesn't, but why?! 

For example these builds are nearly 1 month old: https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/view/change-requests/job/PR-24220/
K8S settings seems the same to me. Results in the table are taken from the popup when you click on the build phase - just the critical part without waiting and preparations.

```
| phase        |   old |   new |
| build        |  9:14 |  5:23 |
| verify       | 17:45 | 20:01 |
| cdi          |  6:59 | 16:52 |
| QL_full      |  3:18 |  4:13 |
...
```

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 2018-2021 Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022, 2023 Contributors to the Eclipse Foundation
+ * Copyright (c) 2018, 2021 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -14,6 +14,21 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  */
+
+
+def dumpSysInfo() {
+  sh """
+    id || true
+    uname -a || true
+    env | sort || true
+    df -h || true
+    \${JAVA_HOME}/bin/jcmd || true
+    mvn -version || true
+    ant -version || true
+    cat /proc/cpuinfo || true
+    cat /proc/meminfo || true
+  """
+}
 
 def jobs = [
   "verifyPhase",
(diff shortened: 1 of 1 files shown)


pzygielo:
:crossed_fingers: to get some values to reason from!

pzygielo:
I'm not very familiar with k8s, but as I see it now - it's possible that with my change, referenced above (which seemed fine at that moment) we switched from requesting whole pack (2(4*)CPU, 8GB) for build. One pack is now split into two agents 1CPU/4GB (I guess).

\* https://wiki.eclipse.org/CBI#What.27s_provided.3F

Perhaps we should consider to restore pre- 1baa948c13b968269f725a100f774eba897fd534 config, but use CBI container and tools only.

I'll try [centos-8-6gb](https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/1163#note_679775) in the meantime.

dmatej:
I have create

> I'm not very familiar with k8s, but as I see it now - it's possible that with my change, referenced above (which seemed fine at that moment) we switched from requesting whole pack (2(4*)CPU, 8GB) for build. One pack is now split into two agents 1CPU/4GB (I guess).
> 
> * https://wiki.eclipse.org/CBI#What.27s_provided.3F
> 
> Perhaps we should consider to restore pre- [1baa948](https://github.com/eclipse-ee4j/glassfish/commit/1baa948c13b968269f725a100f774eba897fd534) config, but use CBI container and tools only.
> 
> I'll try [centos-8-6gb](https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/1163#note_679775) in the meantime.

It is still weird, then it shouldn't be possible to be so fast in any way (but was!). But especially wrong is that it crashes without any reason and even error message, see build `#13` - jdbc simply died before it started, and verify simply vanished when running. https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/view/change-requests/job/PR-24252/

dmatej:
I am not yet closing this, some parts seems useful to me.
* Collecting sever.log files
* Dumping informations about current system

But ... it is well visible that each step has different requirements and we should reflect that in Jenkinsfile.

arjantijms (approved):

PR #24257, opened 2023-01-24
Switch compiler to use `--release` flag instead of `source`/`target`

pzygielo:


appserver/microprofile/config/pom.xml
```diff
@@ -63,6 +63,13 @@
 
     <build>
         <plugins>
+            <plugin>
```
dmatej, line 66: Why is it required here? Also it is more code than before ...

pzygielo: > Helidon 3:
> Java 17 minimum JDK. Java 11 no longer supported.

Needed, because otherwise - with inherited `release=11` - `java.lang.Record` is unavailable due to compiler checking what is and what isn't in given release.

Maybe it would be better to compile this module for release 17?

PR #23770, opened 2022-01-23
Fixed ejb tests + Fixed generating EJB classes

dmatej:
This PR is a bit nonstandard, because it fixes several broken things which fixed work together. The process was long and iterative as I learned to understand the causes.

1. Fixed Glassfish's EJB tests (ejb_all, ejb_group_1, ejb_group_2, ejb_group_3 + new ejb_group_embedded)

  - tested with JDK11 and JDK17 locally
  
image

 
2. Fixed generating classes and proxies for EJB (PFL is broken) in JDK17
3. Replaced most of deprecated calls to ClassLoader.defineClass - left just where there is no other way
4. Some issues were resolved by refactoring
5. Removed j2ee management-api (already unused)
6. Optimized Jenkinsfile and related files to get what we can from the build server, however there is still work to do.

TCK test result:
image

Issue #22638, opened 2018-11-28
Upgrade antlr to version 4.x


m0mus:

bshannon:
We got a special exemption to allow us to use Antlr 2.x for Eclipse GlassFish 5.1,
but we need to upgrade to Antlr 3.x or newer for the subsequent release.
Antlr is used by the old EJB CMP support.

oasuncion:
Hello,
I can see that this ticket was flagged 'help wanted' last November, let me know if I can help on this.
BR

bshannon:
As far as I know, no one is working on this, so yes, we still need help.  Thanks!

oasuncion:
Ok, in this case I will have a look.
I will try to update to the last antlr available version (4.7.2), identify and update the impacted areas and perform some non-regression testings. 
Any insight is appreciated (I am new to GlassFish and Antlr).
BR

oasuncion:
Hi,
I did a test, upgrading the antlr dependency to 4.7.2. 
I could update the grammar files taking into account the new syntax for parser/lexer definitions in Antlr 4, this part is fine.
The problem I am facing now is related to the Java code written in these files, intended to be part of the generated code. Changes between Antlr 2.x and 4.x are huge, most of this code doesn't compile anymore and it is often hard for me to see how to properly upgrade it as I am missing the initial intend of the code.
It would be better to have this ticket handled by someone with a knowledge of the initial work and of Antlr 4, so he may have a better idea of which part of code is not relevant anymore, which is still relevant, and in this case how to upgrade it.

bshannon:
If only we had such a person!

All the people who did this work initially are gone.
We no longer have anyone who has a deep understanding of this code.
Someone is going to need to learn enough about this code to understand how to change it.

smillidge:
Most of the antlr usage is in old cmp code.  Do we still want that code? Is there anybody still using it?

bshannon:
It's still required so that we have a complete implementation of all required
and optional Java EE / Jakarta EE APIs.

pzygielo:
It's still at 2x:
https://github.com/eclipse-ee4j/glassfish/blob/e0a2a1f1fb67178d17b336aca2683706ad56d889/nucleus/parent/pom.xml#L144

arjantijms:
We probably want to make sure CMP gets actually pruned/removed instead of just made optional.

pzygielo:
It's still at 2x:
[...]

PR #24258, opened 2023-01-24
Drop mvn's `-T` option

pzygielo:
Parallel build uses goals not marked as `@threadSafe` by plugins:
```
 [WARNING] org.glassfish.build:glassfishbuild-maven-plugin:3.2.27
 [WARNING] org.glassfish.build:spec-version-maven-plugin:2.1
 [WARNING] org.glassfish.hk2:config-generator:2.5.0-b53
 [WARNING] org.glassfish.hk2:consolidatedbundle-maven-plugin:3.0.3
 [WARNING] org.glassfish.hk2:hk2-inhabitant-generator:3.0.3
 [WARNING] org.omnifaces:antlr-maven-plugin:2.4
```

Thus I propose to consider **not using** `-T` option, until above are upgraded.

`-T`s seem to be introduced in
- https://github.com/eclipse-ee4j/glassfish/pull/23770

but in that PR the usage of non-threadSafe goals and warnings from maven are not discussed.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -147,7 +147,7 @@ pipeline {
           uname -a
 
           # Until we fix ANTLR in cmp-support-sqlstore, broken in parallel builds. Just -Pfast after the fix.
-          mvn clean install -Pfastest,staging -T4C
+          mvn clean install -Pfastest,staging
           ./gfbuild.sh archive_bundles
           mvn clean
           tar -c -C ${WORKSPACE}/appserver/tests common_test.sh gftest.sh appserv-tests quicklook | gzip --fast > ${WORKSPACE}/bundles/appserv_tests.tar.gz
--- a/gfbuild.sh
+++ b/gfbuild.sh
@@ -67,7 +67,7 @@ archive_bundles(){
 }
 
 dev_build(){
-  mvn -U clean install -Pstaging,fastest -T2C ${MVN_EXTRA}
+  mvn -U clean install -Pstaging,fastest ${MVN_EXTRA}
 }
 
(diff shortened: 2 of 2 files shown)


dmatej:
It is just a declaration, ANTLR plugin has much more serious issues, other work well. I don't see real reason for this ...?

pzygielo:
> I don't see real reason for this?

Feel free to close this then as the confirmation that the goals work correctly, please.

dmatej:
> > I don't see real reason for this?
> 
> Feel free to close this then as the confirmation that the goals work correctly, please.

I don't like closing work of someone else, honestly - I forgot on these warnings for a while. 
For ANTLR we have this one: #22638. 
I believe other plugins should be just revisited, so perhaps us two could do that in next few days (weeks) ... If we would find that build cannot be executed in parallel, we would have to really merge this, but for 2 years I did not have any issue (always running -T4C if I am not running tests and if my Eclipse IDE doesn't start compilation in parallel to Maven, which still cannot be resolved by this change).

arjantijms:
p.s. We had the `@Treadsafe` marking of all these plug-ins (after testing, and manual analysis if possible/where needed) on an internal TODO many years ago.

pzygielo:
Thanks for the feedback.

pzygielo:
> p.s. We had the `@Treadsafe` marking of all these plug-ins (after testing, and manual analysis if possible/where needed) on an internal TODO many years ago.

I took the liberty to made few of them public - they are referenced above. Not sure if this will make the maven warnings go away.

PR #24264, opened 2023-01-24
Integrate Soteria 3.0.2

arjantijms:
Changes: https://github.com/eclipse-ee4j/soteria/compare/3.0.1-RELEASE...3.0.2-RELEASE

Signed-off-by: Arjan Tijms <arjan.tijms@omnifish.ee>

### DIFF
--- a/appserver/pom.xml
+++ b/appserver/pom.xml
@@ -153,7 +153,7 @@
 
         <!-- MicroProfile Config -->
         <microprofile.config-api.version>3.0.2</microprofile.config-api.version>
-        <helidon-config.version>3.1.0</helidon-config.version>
+        <helidon-config.version>3.0.2</helidon-config.version>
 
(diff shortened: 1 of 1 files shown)


appserver/pom.xml
```diff
@@ -153,7 +153,7 @@
 
         <!-- MicroProfile Config -->
         <microprofile.config-api.version>3.0.2</microprofile.config-api.version>
-        <helidon-config.version>3.0.2</helidon-config.version>
+        <helidon-config.version>3.1.0</helidon-config.version>
```
hs536, line 156: helidon upgrades that may not be necessary is mixed in.

dmatej: From my perspective it would be alright just to rename the PR including descriptions.

arjantijms: We could do both, but I'll revert the Helidon change then. Thx!

dmatej (approved):

PR #24281, opened 2023-01-29
Remove temporary directory on exit

pzygielo:


dmatej (approved):

dmatej:
Does it work even when the directory has some content?

pzygielo:
> Does it work even when the directory has some content?

No, it doesn't - directory has to be empty. But it works in our case, as we do `file.deleteOnExit();`. So after tests - a lot of empty `gf_uploads*` were left in temporary directory.

PR #24283, opened 2023-01-31
fix: update update_version per 7.0.1 release

chenrui333:
Looks like the glassfish 7.0.1 release artifact contains the 7.0.0 version info, raise this PR to patch that version mismatch

relates to https://github.com/Homebrew/homebrew-core/pull/121872

### DIFF
--- a/appserver/pom.xml
+++ b/appserver/pom.xml
@@ -191,7 +191,7 @@
         <version_suffix />
         <major_version>7</major_version>
         <minor_version>0</minor_version>
-        <update_version>0</update_version>
+        <update_version>1</update_version>
         <install.dir.name>glassfish7</install.dir.name>
 
         <test.logManager>org.glassfish.main.jul.GlassFishLogManager</test.logManager>


appserver/pom.xml
```diff
@@ -191,7 +191,7 @@
         <version_suffix />
         <major_version>7</major_version>
         <minor_version>0</minor_version>
-        <update_version>0</update_version>
+        <update_version>1</update_version>
```
pzygielo, line 194: 1. for the master it should be 2 already
2. I think those should be calculated, maybe with [build-helper:parse-version](https://www.mojohaus.org/build-helper-maven-plugin/parse-version-mojo.html) or something else - to never need to update this manually again

chenrui333: > I think those should be calculated, maybe with [build-helper:parse-version](https://www.mojohaus.org/build-helper-maven-plugin/parse-version-mojo.html) or something else - to never need to update this manually again

yeah, that would be better solution.

chenrui333: > for the master it should be 2 already

I am aware of the master status, but want to call this out.

pzygielo: > I am aware of the master status, but want to call this out.

I see. So new issue would be better choice to report this with. (And it does not require ECA to sign.)

chenrui333: yeah, that is true. But for me to proceed on the homebrew PR, I should propose a known fix one (that is why this PR got created)

chenrui333: And for some projects (not saying this one :) ), the feedback on issues are not faster than PRs. So I always prefer creating the PRs (I actually already signed the ECA XD )

chenrui333: not sure why ECA still failing :(
image

arjantijms: I revalidated the ECA, but unfortunately it keeps failing.

pzygielo: ECA status might take some time to propagate to service.

This however is wrong change. There will be no 7.0.1 release, as it's already in central, and this will be out of date immediately. This will need manual intervention again, and I will not remember that in few hours.

chenrui333: @pzygielo agree with you, closing this PR for now (as we have shipped the homebrew PR). :) 

Thanks @arjantijms for the revalidation, I will see if I can contribute a better patch.

chenrui333: Appreciate the comments in here.

dmatej:
Arrrgh, and I reviewed it in Eclipse manually, but forgot to check this property.

PR #24300, opened 2023-02-20
Fixed #24297: getResource for a directory

### LINKED ISSUE #24297: GlassFish 7.0.1 does not return directories any more
There is a regression from GlassFish 7.0.0 into GlassFish 7.0.1
If you have code that requests a directory from the ClassLoader it fails now on exploded deploy

## Environment Details
* GlassFish Version (and build number):
* 7.0.1 release
* JDK version:
* JDK 17
* OS:
* Windows/unix

----------

## Problem Description
public class MyServletextends HttpServlet {
  protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws ServletException {
    System.out.println("found url " + getClass().getClassLoader().getResource("adirectoryname"));
  }
}
The classloader returned there is WebappClassLoader that was refactored.
It works on packaged deploy and all older GF versions.

Another place we got the same issue was from Spring:
    appContext = new GenericXmlApplicationContext();
    appContext.load("conf/*-config.xml"); -- now fails because folder conf is not found from the classpath on exploded deploy.

## Steps to reproduce
Get an instance of WebappClassLoader and try to load a directory URL from there:
getResource("somerandomdirectory")

## Impact of Issue
The biggest one would be spring that tries to load all xml-s from a directory on exploded deploy.

dmatej:
* Fixes #24297 
* New tests
  * Failed with 7.0.1
  * Passed with 7.0.0
  * Passed with 7.0.2-SNAPSHOT

### DIFF
--- a/appserver/web/war-util/src/main/java/org/glassfish/web/loader/RepositoryManager.java
+++ b/appserver/web/war-util/src/main/java/org/glassfish/web/loader/RepositoryManager.java
@@ -46,6 +46,9 @@ class RepositoryManager implements Closeable {
         LOG.log(DEBUG, "addRepository(repository={0}, file={1})", repository, directory);
         Objects.requireNonNull(repository);
         Objects.requireNonNull(directory);
+        if (!repository.endsWith("/")) {
+            throw new IllegalArgumentException("The repository must end with the '/' character!");
+        }
         repositories.add(new Repository(repository, directory));
     }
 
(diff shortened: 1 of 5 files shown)


arjantijms:
The tests keep failing on this one. I tried to start the job again, but it failed again :(

dmatej:
And passed locally, weird. Issues with Jenkins again?

arjantijms (approved):

PR #24309, opened 2023-02-22
GlassFishLogManagerLifeCycleTest stabilized

dmatej:
- all logs are asynchronous here, so we have to respect that we don't have control over those unsynchronized parallel executions. Same as we don't have that control when it is running in multithreaded appserver.
- as an alternative we use the Thread.sleep, however even this might not be enough on some systems (virtual systems with cpu sharing limits, ie.).


arjantijms (approved):

pzygielo:
Thanks!

Issue #24284, opened 2023-01-31
Looks like the glassfish 7.0.1 release artifact contains the 7.0.0 version info


pzygielo:
_Originally posted by @chenrui333 in https://github.com/eclipse-ee4j/glassfish/pull/24283_

pzygielo:
Perhaps
```xml
        <major_version>7</major_version>
        <minor_version>0</minor_version>
        <update_version>0</update_version>
```
could be determined automatically in the build time.

dmatej:
Arrrgh, and I reviewed it in Eclipse manually, but forgot to check this property. 
There's also ./updateVersion.sh script, it could be included into that. Just simple regexp.
Or some maven plugin, btw there's yet another bug with generating build id - it generates wrong date part (2020 in my builds).

szarza:
The "about" button in admin web application of 7.0.1 shows :
Eclipse GlassFish 7.0.0 (build master-b160-g0b1e109 2020-12-19T17:24:00+0000)

Could it be for the same reason?

dmatej:
> The "about" button in admin web application of 7.0.1 shows : Eclipse GlassFish 7.0.0 (build master-b160-g0b1e109 2020-12-19T17:24:00+0000)
> 
> Could it be for the same reason?

Yes, this has the same reason.

PR #24316, opened 2023-02-24
Issue #24284 Fixed git.build.time

dmatej:
- Fixes comment in #24284
- was overriden by a value from eclipse project 1.0.8

### DIFF
--- a/nucleus/parent/pom.xml
+++ b/nucleus/parent/pom.xml
@@ -174,6 +174,8 @@
         <nucleus.install.dir.name>nucleus</nucleus.install.dir.name>
         <javadoc.skip>false</javadoc.skip>
         <deploy.skip>false</deploy.skip>
+        <!-- Overrides the value from parent and enables the one we generate. -->
+        <project.build.outputTimestamp></project.build.outputTimestamp>
 
         <maven.test.jvmoptions.add-opens>--add-opens java.base/java.lang=ALL-UNNAMED --add-opens=java.naming/javax.naming.spi=ALL-UNNAMED</maven.test.jvmoptions.add-opens>
         <maven.test.jvmoptions.memory.sizes>-Xss512k -Xms256m -Xmx1g -XX:MaxDirectMemorySize=512m</maven.test.jvmoptions.memory.sizes>


pzygielo:
I'd rather have ever-changing `git.build.time` removed from `build.id` than move away from reproducible builds goal (I understand it's centuries of work to do to have it here, and such goal is not even established yet...).

IMO - the timestamp in signature file is enough.

- https://github.com/eclipse-ee4j/ee4j/pull/71 - @hboutemy - what do you think?

dmatej:
> I'd rather have ever-changing `git.build.time` removed from `build.id` than move away from reproducible builds goal (I understand it's centuries of work to do to have it here, and such goal is not even established yet...).
> 
> IMO - the timestamp in signature file is enough.
> 
> * [prepare Reproducible Builds ee4j#71](https://github.com/eclipse-ee4j/ee4j/pull/71) - @hboutemy - what do you think?

I have yet another changes prepared, but I wanted to push this first before somebody starts screaming that I did too many changes again :D (revisited branding.properties, related documentation and implementation)

Also - the commit is useful when you start arguing what was the source commit of the build. That's what I did in my next PR (yet after one with plugin updates).

I don't understand the reason for https://github.com/eclipse-ee4j/ee4j/pull/71 - it doesn't make builds reproducible at all. I think it should work exactly opposite - to be able to run the build with -Dproject.build.outputTimestamp with one concrete time stamp for several executions, but this change made it mandatory.

dmatej:
But maybe we could even abandon the build id and rather replace it with more readable information instead of generating "magical" id: 
* the source commit
* the timestamp

Then in the log output instead of this:
`Running GlassFish Version: Eclipse GlassFish  7.0.2  (build updates-b43-g6276a24 2020-12-19T18:24:00+0100)]]`
would be this:
`Running GlassFish Version: Eclipse GlassFish  7.0.2-SNAPSHOT  (commit ec1ce24934f0bb174333a9886b58a0d2f9e52b44, timestamp 2023-02-25T18:24:00+0100)]]`

(My usual issue - do I work with snapshot or with a released version? How old is it?)

pzygielo:
> I'd rather have ever-changing `git.build.time` removed from `build.id`.

This would leave other `build.id`' components like branch, commits since last tag and commit to give: `updates-b43-g6276a24`. Which by the way - `b{xx}` is questionable.

I'm all for only-commit (full hash) (and no timestamp).

pzygielo:
> I have yet another changes prepared, but I wanted to push this first
:+1:

pzygielo (approved):

PR #24320, opened 2023-02-25
Reorganized jenkinsfile

dmatej:
- verify will follow the compilation, ant tests will be executed as last in parallel jobs with lower resources
- more resources for compilation and maven integration tests
- if maven fails, ant will not run now
- Step to future: my intention is to start migrating the Zombie Army (Ant tests) to Maven. At some point I will start splitting the maven part to parallel tests too, but they will be much faster than now with Ant.

Note - this change is not going back before 1baa948c13b968269f725a100f774eba897fd534 : 
* We are using standard images now (maven+temurin+ubuntu docker image and basic Jenkins nodes)
* I just merged recommendations of https://wiki.eclipse.org/Jenkins with recently abandoned k8s settings, because state before this commit was quite sensitive to the hardware chosen by Jenkins ... sometimes. I have noticed at least two different CPUs, one 2,1GHz, second much more modern 4GHz. Another factor were k8s settings. Especially tests running more instances of GlassFish then had issues with timeouts.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -30,8 +30,7 @@ def dumpSysInfo() {
   """
 }
 
-def jobs = [
-  "verifyPhase",
+def antjobs = [
   "cdi_all",
   "ql_gf_full_profile_all",
   "ql_gf_web_profile_all",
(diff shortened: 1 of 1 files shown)


dmatej:
Note - this change is not going back before 1baa948c13b968269f725a100f774eba897fd534 : 
* We are using standard images now (maven+temurin+ubuntu docker image and basic Jenkins nodes)
* I just merged recommendations of https://wiki.eclipse.org/Jenkins with recently abandoned k8s settings, because state before this commit was quite sensitive to the hardware chosen by Jenkins ... sometimes. I have noticed at least two different CPUs, one 2,1GHz, second much more modern 4GHz. Another factor were k8s settings. Especially tests running more instances of GlassFish then had issues with timeouts.

hs536 (approved):

PR #24337, opened 2023-03-10
Refactoring around WebBundleDescriptor and EjbBundleDescriptor

dmatej:
Context:
* CargoTracker touched several issues.
* Some were already resolved and fixes were merged in #24335 
* Some I have "fixed" locally, but I did't believe the fix is alright. So I started writing unit tests but immediately collided with the fact that these two classes are not mockable and I can't even create some trivial fake implementations. So we have just integration tests, slow and covering just one of scenarios while I need a code coverage going through many possibilities. Even then there could be a mistake, but I still need to be able to just extend tests in 5 minutes.

So, here we go.

* new javadocs (and comments for methods used by reflection)
* new tests
* removed obsoleted methods with alternatives + updated usages
* renamed "hacks"
* methods with just a single usage moved to where they are used, simplified.
* some bugs were fixed by refactoring (not in between, but directly "by")

Fixed issues:
* Validation of resource bundles in deployment - spammed logs without a reason (CargoTracker did it)
* Several NullPointerExceptions when deployment failed (broken CargoTracker state)
* Now it should be possible to write more unit tests for DOL, open door for future iterations of refactoring.

Tests:
 
* TCK: https://ci.eclipse.org/jakartaee-tck/job/jakartaee-tck/job/10.0.x/50/ (all passed)
* CargoTacker on JEE10 - in progress, there's probably yet one bug, but I will create another PR for it as it seems not related to descriptors.

Breaking changes:

* Changed API of several classes in deployment, it may affect connectors despite I tried to keep some consistency. 
* However I doubt someone outside GF would be able to extend/implement those APIs to work well with GlassFish as the API is something horrible; too many dependencies between classes, circular relations, etc.


smillidge:
This is a large broad set of changes touching a number of subsystems and 212 files. Not sure I can review it tbh. I'd have to assume it is correct.

appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/connectors/deployment/annotation/handlers/ConnectionFactoryDefinitionHandler.java
```diff
@@ -268,7 +160,7 @@ private ConnectionFactoryDefinitionDescriptor createDescriptor(ConnectionFactory
         desc.setMaxPoolSize(defn.maxPoolSize());
         desc.setMinPoolSize(defn.minPoolSize());
 
-        if (defn.description() != null && !defn.description().equals("")) {
+        if (defn.description() != null && !defn.description().isEmpty()) {
```
arjantijms, line 163: Could also use the GF utility method for this; 

com.sun.enterprise.util.Utility.isEmpty(String)

arjantijms (approved):

appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/connectors/deployment/annotation/handlers/ConnectionFactoryDefinitionHandler.java
dmatej: > This is a large broad set of changes touching a number of subsystems and 212 files. Not sure I can review it tbh. I'd have to assume it is correct.

I completely agree that it is horrible, but I have no idea how to move from "swamps of DOL" by a different way. I could make a PR after every safe point, but even then it would not help too much. Everything is fragile, in previous iterations I learned some "rules" how to avoid breaking things. Also in the first iteration took whole day to find out where I did the bad step, this time it was much faster.

There are overlapping interfaces and methods, removing one can break deployment or related things despite it compiles. Many reflections used to set attributes or just call methods; when they are just setters or "adders", it is easy, but sometimes they depend on the order of execution, which is fragile, fragile, fragile ...

That leads to some "stack" and "routine" in refactoring. And also thinking about next iterations.

Reviewing it by reading code is nearly worthless (except well visible mistakes I really could make), that's why I always run the TCK and why we were helping the TCK (so we can run it against GF PR), in last two years we also (partially) fixed the "Zombie Army" (Ant tests, I'm often checking if the count of passing tests did not decrease), now I'm trying to make friends of GF and CargoTracker (the last issue: derby pool). 

Good thing is that in the future it should converge to some more ideal state and those PRs should be smaller and smaller. I believe that creating them is still more painful than reading them.

hs536 (approved):
it passed the TCK, so we could say it meets the minimum quality requirements. although incompatibilities should be carefully checked afterwards, I'd welcome modifications that would make GlassFish more testable.

PR #24360, opened 2023-03-27
Fix sporadic error in view log REST integration test

avpinchuk:
This is a bug fix.

Under some circumstances the test may sporadically fails. This occurs when we expect an empty log, but it actually may contains several log records. I'm assuming this is due to the deferred write.


dmatej:
Nope, this is not how we should resolve it. Finally we would not test anything.

avpinchuk:
Because query param of request formally is an offset from the beggining of the log file, we have two cases:

* when second get returns the emty string then an offset remains same as in previous query,
* when is not empty then offset also greater.

Maybe try to use this conditions?

avpinchuk:
Also, when second  response is empty then url's are equals. And non-equals in opposite case

avpinchuk:
I'll reopen the PR. Which option is preferred - parse offsets or just compare url's?

dmatej (approved):

PR #24361, opened 2023-03-28
Cleanups in connectors logging

dmatej:
- some messages were broken
- some causes were lost
- several logs were redundant
- fixed some javadocs
- these changes were done locally when I was working on #24335 and #24337 
- 08d7fc5493a81505ab9b307aa42b72158333a3fb is rather trivial, I wasn't sure if it cannot lead to negative pool size or size higher than max. Now it is covered by a unit test.


appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/allocator/LocalTxConnectorAllocator.java
```diff
@@ -81,17 +80,6 @@ public ResourceHandle createResource()
 
             return resource;
         } catch (ResourceException ex) {
-            Object[] params = new Object[]{spec.getPoolInfo(), ex.toString()};
-            _logger.log(Level.WARNING, "poolmgr.create_resource_error", params);
```
hs536, line 85: Is the deletion of this log as intended? (delegate log to the caller who catches the PoolingException?

dmatej: Yes, because the exception is wrapped and thrown again several times. In effect when something broke, I had 5 times the same error in the server.log file, yet once with a broken message.

But based on your comment it seems to me that some paths may result in hiding the exception, so for sure I will revisit all paths and ensure that the exception is logged exactly once including stacktrace.

appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/allocator/ConnectorAllocator.java
```diff
@@ -143,30 +140,14 @@ public ConnectorAllocator(PoolManager poolMgr,
...
             mc.addConnectionEventListener(l);
             return resource;
         } catch (ResourceException ex) {
-            Object[] params = new Object[]{spec.getPoolInfo(), ex.toString()};
-            _logger.log(Level.WARNING,"poolmgr.create_resource_error",params);
-            if(_logger.isLoggable(Level.FINE)) {
-                _logger.log(Level.FINE,"Resource Exception while creating resource",ex);
-            }
-
-            if (ex.getLinkedException() != null) {
```
dmatej, line 165: This is obsoleted deprecated getter, generally not supported any more, and also ignored on other places. 
TODO in next commits of this PR: Remove all calls of the setter.

appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/allocator/LocalTxConnectorAllocator.java
dmatej: TODO for next commits of this PR: catch and process the exception, don't repeat logging and rethrowing.

dmatej: I have revisited that one more. It is finally called just from one place in ConnectionPool, which throws the exception including the cause OR sleeps and gives it another try. Now, before the sleep, it should log it as a warning including some message what will happen next.

hs536 (approved):

PR #24366, opened 2023-04-01
enable reproducible builds

hboutemy:
just applied fixes proposed by `mvn artifact:check-buildplan`, because release 7.0.3 is not reproducible at all (outputTimestamp fully dropped instead of setting an initial value)

see https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/glassfish/main/README.md

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -43,6 +43,10 @@
       <url>https://github.com/eclipse-ee4j/glassfish</url>
     </scm>
 
+    <properties>
+        <project.build.outputTimestamp>2023-04-02T15:44:03Z</project.build.outputTimestamp>
+    </properties>
+
     <build>
         <defaultGoal>install</defaultGoal>
         <pluginManagement>
--- a/appserver/pom.xml
+++ b/appserver/pom.xml
@@ -172,7 +172,7 @@
         <schema2beans.version>6.7</schema2beans.version>
         <derby.version>10.15.2.0</derby.version>
         <wsdl4j.version>1.6.3</wsdl4j.version>
-        <maven-rar-plugin.version>2.4</maven-rar-plugin.version>
+        <maven-rar-plugin.version>3.0.0</maven-rar-plugin.version>
 
         <product.name>Eclipse GlassFish</product.name>
         <product.name.abbreviation>GF</product.name.abbreviation>
--- a/nucleus/parent/pom.xml
+++ b/nucleus/parent/pom.xml
@@ -174,7 +174,7 @@
         <javadoc.skip>false</javadoc.skip>
         <deploy.skip>false</deploy.skip>
         <!-- Overrides the value from parent and enables the one we generate. -->
-        <project.build.outputTimestamp></project.build.outputTimestamp>
+        <project.build.outputTimestamp>2023-04-02T15:44:03Z</project.build.outputTimestamp>
 
         <maven.test.jvmoptions.add-opens>--add-opens java.base/java.lang=ALL-UNNAMED --add-opens=java.naming/javax.naming.spi=ALL-UNNAMED</maven.test.jvmoptions.add-opens>
         <maven.test.jvmoptions.memory.sizes>-Xss512k -Xms256m -Xmx1g -XX:MaxDirectMemorySize=512m</maven.test.jvmoptions.memory.sizes>


pom.xml
```diff
@@ -43,6 +43,10 @@
       <url>https://github.com/eclipse-ee4j/glassfish</url>
     </scm>
 
+    <properties>
+        <project.build.outputTimestamp>2020-12-19T17:24:00Z</project.build.outputTimestamp>
```
dmatej, line 47: We want to know when the artifact was built and I am sure it wasn't in 2020.

pzygielo: > We want to know when 

- I don't. https://github.com/eclipse-ee4j/glassfish/pull/24316#issuecomment-1445037472

hboutemy: it will be updated when you use `mvn versions:set`

dmatej: I opened the link from the description and I think I understand the issue. It is a timestamp, TIMESTAMP. It doesn't matter if it is in the signature file, pom or properties, Maven Central just wants that it doesn't change in the release.

I will try something on monday. 

Settting that to 2020 or any other value for snapshots is clearly a nonsense.

hboutemy: you can have a look at https://maven.apache.org/guides/mini/guide-reproducible-builds.html#faq

your reaction is classical. you're discovering that a timestamp is a timestamp that does not always represent "now" from the current build (which is by definition not reproducible: "now" from 2 builds on the same code is not the same): a timestamp is a timestamp of "something"

> or any other value for snapshots is clearly a nonsense

I worked on it for sufficient time to tell you that it is NOT a nonsense, but a change from previous habits.
For some people who feel unhappy, they set this timestamp to the last Git commit: makes them less unhappy from a logical perspective.
From a build cache perspective in a context like yours of a massive multi-module build, it's not the best choice: many modules will have their output checksum changed even if no code changed.
FYI, on that question, Gradle did a straightforward choice: timestamp is a fixed value = start of Unix EPOCH...

At least, here, the timestamp is updated on release: releases give you an expected semantics on the timestamp
And you'll have to accept that the timestamp of the start of a new spring is the best option to void rebuilding content when it's not necessary

dmatej: So now the definition of the "repeatable build" is just having a constant timestamp in pom.xml, right? It doesn't matter if jar files have different content (and they do). It is not just illogical, it is simply wrong.

> your reaction is classical. you're discovering that a timestamp is a timestamp that does not always represent "now" from the current build (which is by definition not reproducible: "now" from 2 builds on the same code is not the same): a timestamp is a timestamp of "something"

If it is classical, it is probably a signal that your idea may be wrong. How setting a timestamp to a constant makes builds repeatable? It doesn't, it is just a fake date.

> many modules will have their output checksum changed even if no code changed.

Yes! Exactly - when checksums changed, we don't have 100% reproducible build. And then I want a human readable timestamp from which I can quickly see when I or someone else built the artifact; then I can guess a direction of what happened. 

Maven Central prohibits reuploading artifacts if I remember well, but that's not true for other repositories. Then is useful to find who, where, when and how someone created the duplicit release. And then we have snapshots, where you do that nearly on a daily basis.

Your change affects what we see this in logs; now there would be year 2020:
```
Running GlassFish Version: Eclipse GlassFish 7.0.3-SNAPSHOT (branch: master, commit: 1cdea232a94fadebf51523440c31afa74b29c38d, timestamp: 2023-03-30T11:16:14Z)]]
```

> Gradle did 

I don't know even how Gradle uses timestamps, because it is always a timestamp of some event. As you mentioned - it depends on the usage. In our case it is a build timestamp. And as builds are never 100% repeatable, which is proven by changing checksums of artifacts, it should change too. But ok, current hype is that build timestamp is not a timestamp.

> the timestamp is updated on release

In fact it is NOT updated. You just set the really old date here. We release a version nearly every month and the timestamp was this old date for several versions.

> it will be updated when you use mvn versions:set

1. MojoHaus versions plugin is not a standard plugin. It shouldn't be used in "perfect" projects, it is rather an additional tool. GlassFish is not a perfect project yet, so sometimes we use it.
2. Maven release plugin is the right plugin. Here it writes about our issue: https://maven.apache.org/guides/mini/guide-reproducible-builds.html#faq

-------
So let's get to something, please tell me if you agree:

1. For snapshots the value will be generated (as it is until now).
2. For a releases since 7.0.4 we will update the project and releasing process to ensure that in the released pom the value in the pom.xml is set. My current idea is this:
    1. Update the release process to use the maven release plugin as much as possible.
    2. Add your plugin to the build, so it would fail the release build if the value wouldn't be set.
    3. Update the "release howto"

I believe this would resolve the problem for both of us, because then
1. The release will be repeatable in your terminology.
2. The snapshot would change the timestamp with every build (which may contain even uncommitted changes, so using git timestamp is a mistake and using a constant means resignation).

hboutemy:
PR updated to have a current timestamp: reflects more reality of the value that you'd have had when switching to 7.0.4-SNAPSHOT (FYI I did it simply by running `mvn versions:set`)

this PR is about Reproducible Builds: https://reproducible-builds.org/ = the objective is to have same binary as output

having a reproducible timestamp is a necessity to get the same output binary from 2 builds
And from that reproducible timestamp, many recent plugins do their job to ensure reproducible binary output (

you can get the result with versions:set, maven-release-plugin or by using git commit timestamp, as you wish (no resignation in any strategy): you don't need to change your current release practice

dmatej:
>  the objective is to have same binary as output

Hmmm, originally I wanted to say that binaries are still not the same and the date doesn't ensure that, but to my surprise the simplest GJULE.jar has always the same sha after your change and differs without that. But I think I understand you. Is it related to something new in Maven 4?

So I tried that also with the distributed glassfish.zip:
First run: `119157795 B`, sha512: `2551d2e01864e12c0eaf0748de3c95cb945705c49abd25c6a5de4695790172f7a687d9f1c892e6edb9761bb737352ccabacf42e045639f0ada61b9a8beef8f2e`
Second run: `119157586 B`, sha512: `344d81b146dcc8787a447c9b069101dd2b6cbd2e3ed9add01c30d6ff3c75e63a723fdb382dbe53d214ec7faa7f65b0408e812a5dba7b670bb62cf1a9c0a54afa`
Third run: `119157663 B`, sha512:
`1809afa52dfd4f2ec3e38a02dde8840b2e7dfc6696384e754b59c8896abbca88b8d55a7a01dab3eb52aa1d488167319cbc978e4c4df010576b57b2124ed4d4b2`

So no, we still don't have the same binary output. And this artifact is the most important.

I still have doubts about this idea, but I understand now why @pzygielo wrote that he doesn't need the timestamp at all. After we merge this PR, we simply have to remove also all usages from properties and the code, because it provides false information and is worthless.

As a result we lose some information just to pass `mvn artifact:check-buildplan`, which in fact doesn't verify that the artifact will be the same in every build of the same commit - it passed on this PR, but I proved that the zip file is different in every build.

So then we have to find out why the zip changes between builds. Alright, we can merge this, temporarily as this is not the final fix. I don't like that I am losing the simplest possible way to check if I started the right artifact, but what can I do ...

dmatej:
@hboutemy Thank you for your patience, btw. ;-)

dmatej (approved):

hboutemy:
> But I think I understand you. Is it related to something new in Maven 4?

everything happens at plugins level, nothing at Maven level itself: if you want to see more details, the hard core ones are here https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=74682318

`artifact:check-buildplan` is a quick check of basics, with help for quick fixes. You're right, it's not a real check.
Really checking is covered in https://maven.apache.org/guides/mini/guide-reproducible-builds.html#how-to-test-my-maven-build-reproducibility
And for studying differences between different outputs, `diffoscope` is our friend https://diffoscope.org/

I have reproduced more than 1200 releases until now https://github.com/jvm-repo-rebuild/reproducible-central , from smallest projects to biggest ones
Big projects usually require a few releases to fix issues step by step: see for example https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/nifi/nifi/README.md

hboutemy:
i ran a real test with this PR applied: ` mvn clean install -Pfastest && mvn clean verify -Pfastest artifact:compare`

many remaining issues are just 
```
diffoscope target/reference/console-jca-plugin-7.0.4-SNAPSHOT.jar appserver/admingui/jca/target/console-jca-plugin.jar
--- target/reference/console-jca-plugin-7.0.4-SNAPSHOT.jar
+++ appserver/admingui/jca/target/console-jca-plugin.jar
├── META-INF/hk2-locator/default
│ @@ -1,7 +1,7 @@
│  #
│ -# Generated on Mon Apr 03 00:02:42 CEST 2023 by hk2-inhabitant-generator
│ +# Generated on Mon Apr 03 00:05:12 CEST 2023 by hk2-inhabitant-generator
│  #
```

easy to solve: i can provide more PRs after the current first one to make sure release 7.0.4 will have those simple issues fixed

but as you mentioned in #24367, you have a choice to do: do you want Reproducible Builds or not?

hs536:
IMHO, setting the value of `project.build.outputTimestamp` at the time of release will solve both problems.

from(current release commit proc):
```
master
+-@someone: Something latest commit
+-@glassfish-bot: Prepare release 7.0.4
+-@glassfish-bot: Prepare next development cycle for 7.0.5-SNAPSHOT
```
to:
```
master
+-@someone: Something latest commit
|  +-@glassfish-bot: Prepare release 7.0.4 (* set timestamp and create tag here)
+-@glassfish-bot: Prepare next development cycle for 7.0.5-SNAPSHOT
```

What do you think?

hs536:
> IMHO, setting the value of `project.build.outputTimestamp` at the time of release on a release branch will solve both problems.

This can be done with a few changes to Jenkins' settings. (This does not solve the problem of the zip's sha changing for each build, though)

dmatej:
> > IMHO, setting the value of `project.build.outputTimestamp` at the time of release on a release branch will solve both problems.
> 
> This can be done with a few changes to Jenkins' settings. (This does not solve the problem of the zip's sha changing for each build, though)

Better would be to use maven release plugin, because it can do that automatically.  But if I am the only person which uses this "feature" in logs, it is not worth of the effort. 
I would not like to implement it in Jenkins, I would like to make it more simple, so use standard release:prepare and release:perform.
Another argument is that some tools may be affected, as Herve wrote; see #24367 where I tried to simplify description of the issue. Basically some tools check that built artifacts did not change. But it the value is generated, they do change -> ie. editor will rebuild the whole workspace every time. Then it would be an issue even for snapshots, @hboutemy is right.

hboutemy:
I suppose the current release process uses `mvn versions:set`: it is perfectly supported and gives same result as maven-release-plugin, you are not forced to change

hs536:
My intention may not have been clear. I propose that we do not explicitly set a value for "project.build.outputTimestamp" during development, and only set it for release. This way, SNAPSHOT versions can still get the usual timestamp.

hboutemy:
oh, now I better understand: that would mean non-Reproducible Builds during SNAPSHOT, and Reproducible Builds only on releases. This is not a scenario I expected, it will require adding and removing configuration forth and back: looks complex.
If you absolutely want to have a moving timestamp timestamp even during development, I suppose the best option is then using last commit timestamp

dmatej:
Yes, that was my idea too, but if tools like Eclipse IDE check checksums of generated artifacts, it would be still wrong, unreproducible build. For example if m2e would use this instead of current monitoring of classes for quick rebuilds of the workspace:

* old way, builds not reproducible: I change something in one dependency, used by many other -> direct dependencies are rebuilt too -> their dependencies -> ... ... -> everything was rebuilt.
* reproducible builds: I change something in one dependency, used by many other -> they will rebuild too -> but their checksum did not change -> no need to rebuild transitive dependencies, we are done.

I believe this is maybe more important that reproducible builds of releases, which are always built just once, uploaded, and not changed forever, that is ensured by rules of Maven Central. If it works how I described here :-)

hs536 (approved):

PR #24370, opened 2023-04-03
API usage improvement to JDK11 in ASURLClassLoader

ctabin:
Just some API improvements to use up-to-date collections and try-with-resources with the JDK11.

### DIFF
--- a/nucleus/common/common-util/src/main/java/com/sun/enterprise/loader/ASURLClassLoader.java
+++ b/nucleus/common/common-util/src/main/java/com/sun/enterprise/loader/ASURLClassLoader.java
@@ -51,15 +51,15 @@ import java.security.cert.Certificate;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Enumeration;
-import java.util.Hashtable;
+import java.util.HashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.StringTokenizer;
-import java.util.Vector;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.jar.Attributes;
 import java.util.jar.JarEntry;
@@ -89,13 +89,13 @@ public class ASURLClassLoader extends GlassfishUrlClassLoader
     implements JasperAdapter, InstrumentableClassLoader, PreDestroy, DDPermissionsLoader {
 
     /** logger for this class */
-    private static final Logger _logger=CULoggerInfo.getLogger();
+    private static final Logger _logger = CULoggerInfo.getLogger();
 
     /**
      * list of url entries of this class loader. Using LinkedHashSet instead of original ArrayList
      * for faster search.
      */
-    private final Set<URLEntry> urlSet = Collections.synchronizedSet(new LinkedHashSet<URLEntry>());
+    private final Set<URLEntry> urlSet = Collections.synchronizedSet(new LinkedHashSet<>());
 
     /** cache of not found resources */
     private final Map<String, String> notFoundResources = new ConcurrentHashMap<>();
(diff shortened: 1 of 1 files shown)


nucleus/common/common-util/src/main/java/com/sun/enterprise/loader/ASURLClassLoader.java
```diff
@@ -118,7 +118,7 @@ public class ASURLClassLoader extends GlassfishUrlClassLoader
     private volatile String doneSnapshot;
 
     /** streams opened by this loader */
-    private final Vector<SentinelInputStream> streams = new Vector<>();
+    private final List<SentinelInputStream> streams = new ArrayList<>(16);
```
avpinchuk, line 121: Potentially unsynchronized access in `SentinelInputStream` constructor or I wrong?

ctabin: @avpinchuk Yep you're right, should be OK now.

nucleus/common/common-util/src/main/java/com/sun/enterprise/loader/ASURLClassLoader.java
```diff
@@ -887,11 +879,7 @@ private Vector<SentinelInputStream> getStreams() {
      * Must be synchnronized to (a) avoid race condition checking 'streams'.
      */
     private synchronized void closeOpenStreams() {
-        if (streams == null) {
-            return;
-        }
-        SentinelInputStream[] toClose = streams.toArray(new SentinelInputStream[streams.size()]);
-        for (SentinelInputStream s : toClose) {
+        for (SentinelInputStream s : streams) {
```
avpinchuk, line 882: I think we can get here `ConcurrentModificationException` unlike original version

dmatej: True.
Pffff, I opened the class in Eclipse now and it has much more issues ... 
* done() method should be a close() since 1.7. The closed classloader should not be used any more.
   * Such change may result in exceptions in incorrect implementations using this classloader, so we have to fix them.
* All those finalizers should be removed. Java [plans to remove them too](https://openjdk.org/jeps/421).
* Maybe we should continue step by step. I can't see all consequences yet.

nucleus/common/common-util/src/main/java/com/sun/enterprise/loader/ASURLClassLoader.java
```diff
@@ -795,21 +792,16 @@ protected synchronized ClassData findClassData(String name) throws ClassNotFound
...
                 bout.write(buf, 0, num);
             }
-        } finally {
-            try {
-                bstream.close();
-            } catch (IOException closeIOE) {
-                ASURLClassLoader._logger.log(INFO,
-                        CULoggerInfo.exceptionInASURLClassLoader, closeIOE);
-            }
+        } catch (IOException closeIOE) {
```
dmatej, line 802: This doesn't catch just close IOE, but all IOExceptions in the try. Perhaps just don't catch it, it is always a failure, same if bstream.read fails and if we cannot close it.

nucleus/common/common-util/src/main/java/com/sun/enterprise/loader/ASURLClassLoader.java
ctabin: @dmatej Yep, I saw that too, but I wanted to change that step by step to avoid regression. Also finalizers have [been removed](https://github.com/payara/Payara/pull/6179) in Payara, so... I guess the same changes should be included in GlassFish.

ctabin: Fixed.

nucleus/common/common-util/src/main/java/com/sun/enterprise/loader/ASURLClassLoader.java
ctabin: Fixed.

avpinchuk (approved):
LGTM

dmatej (approved):

PR #24371, opened 2023-04-05
Refactor around REST resources related to server logs

### LINKED ISSUE #24068: StructuredLogViewerResource randomly produces corrupted output
I am not quite sure about the cause of the error as LoggingRestITest mostly passes, but sometimes fails because log records/filenames in the response are escaped. That is probably caused by these circumstances:

JSON methods of both StructuredLogViewerResource and LogNamesResource provide both JSON and PLAIN response types. Despite client (the test) asked for JSON, sometimes (only on Jenkins and still not always, "makes me maaaad!!!") produces incorrectly escaped JSON. 

The JSON response (and XML too) is produced manually in the Java code - it uses JSONObject to convert the record to a JSON string and then concatenates these strings to create JSON. Same does it with XML, but there's just single XML response type declaration.

So ... it seems to me that when the client asks for JSON, Jersey takes the string produced by the endpoint and escapes it as a JSON response (escapes all "\n\t..." characters by a backslash), because it expects that the string is an entity to be wrapped as json. Then the response cannot be parsed for some reason as a JsonObject. Example from [Jenkins](https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/job/PR-24067/4/execution/node/365/log/):
```
LoggingRestITest.logFileNames:53 ? JSON A JSONObject text must begin with '{' at character 1 of "{\"InstanceLogFileNames\": [\"server.log\"]}\n"
```
The scary thing is that it is not reproducible. It just sometimes happens.

```
    private String convertQueryResult(List<String> files, String type) {
        String sep = "";
        if (type.equals("json")) {
            sb.append("{\"InstanceLogFileNames\": [");
        } else {
            sb.append("<InstanceLogFileNames>\n");
        }

        // extract every record
        for (String name : files) {
            if (type.equals("json")) {
                sb.append(sep);
                sb.append(quote(name));
                sep = ",";
            } else {
                sb.append("<" + name + "/>");
            }
        }
        if (type.equals("json")) {
            sb.append("]}\n");
        } else {
            sb.append("\n</InstanceLogFileNames>\n");

        }
        return sb.toString();
    }

    private String quote(String s) {
        return "\"" + s + "\"";
    }
```

How to fix it? What I would do is ...

- split PLAIN and JSON response types - it is not the same.
- maybe the PLAIN text would be useful too, no JSON, no XML, just text.
- return just list of LogRecord instances (in JSONObject under the "records" key to keep API ... or in a Map), let Jersey to format it. Same for XML.

However it seems to me quite suspicious that it changes the behavior and I don't know what exactly causes it. 
Any ideas?

avpinchuk:
This PR should resolve #24068.

What's done:

* Refactored Log Names REST endpoint.
* Refactored View Log Details endpoint.
* Used JSON as a default output format when client has not defined a media type.
* Changed output format for media type `text/plain` from JSON-like string to [CSV](https://www.ietf.org/rfc/rfc4180.txt) (based on advice from @dmatej).
* Raised priority of the JSON entity provider.

Signed-off-by: Alexander Pinčuk <alexander.v.pinchuk@gmail.com>

### DIFF
--- a/nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/logviewer/LogRecord.java
+++ b/nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/logviewer/LogRecord.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022, 2023 Contributors to the Eclipse Foundation
  * Copyright (c) 2010, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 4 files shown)


avpinchuk:
@dmatej, what You think about an output format for `text/plain` media type? Now I left it untouched.

dmatej:
> @dmatej, what You think about an output format for `text/plain` media type? Now I left it untouched.

Good question. Any idea? What I would expect with "naive request" ... maybe just comma-separated list of names?

nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/resources/custom/LogNamesResource.java
```diff
@@ -66,35 +83,46 @@ private String getLogNames(String instanceName, String type) throws IOException
...
+                Writer xml = new StringWriter();
+
+                XMLStreamWriter writer = newDefaultFactory().createXMLStreamWriter(xml);
+                writer.writeStartElement("InstanceLogFileNames");
+                for (String file : files) {
+                    writer.writeEmptyElement(file);
+                }
+                writer.writeEndElement();
+                writer.flush();
+                writer.close();
```
dmatej, line 103: It would be better to move that to finally (it is not AutoCloseable, but has the close method); flush is called in close too.

avpinchuk:
> Good question. Any idea? What I would expect with "naive request" ... maybe just comma-separated list of names?

CSV seems good. But what CSV? ;-)

`InstnanceLogFileNames, server.log, server1.log, ... `

or

`InstanceLogFileNames, "[server.log, server1.log, ... ]"`

or just

`server.log, server1.log, ...`

And the second question is what to do with a backward compatibility - we break output format completely. Or it does not matter?

dmatej:
It depends ...
* if it randomly crashed, we don't break anything :D 
* I would keep the KISS principle - just `server.log, server1.log, ...` 
* If user specifies the type, he will receive what he expected, JSON/XML ... and now new plain text. Can we use JSON as a default?

I

avpinchuk:
> * Can we use JSON as a default?

Yes, according to REST spec, we can do it with the `qs` parameter of media type on the server side.

avpinchuk:
CSV not implemented yet

nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/logviewer/LogRecord.java
```diff
@@ -17,44 +17,48 @@
...
-import org.w3c.dom.Node;
 
 /**
- * internal REST wrapper for a log record will be used to emit JSON easily with Jackson framework
+ * Internal REST wrapper for a log record will be used to emit JSON easily
+ * with Jackson framework.
  *
  * @author ludo
  */
+@SuppressWarnings("unused")
```
dmatej, line 36: I don't like these annotations - why it is here?
If there is some unused field, it should be removed OR at least documented why is it here.

nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/resources/custom/StructuredLogViewerResource.java
```diff
@@ -58,145 +69,204 @@ public class StructuredLogViewerResource {
...
         }
 
-        if (type.equals("json")) {
-            sb.append("]}\n");
-        } else {
-            sb.append("\n</records>\n");
+        return Response.ok(entity.toString(), type).build();
+    }
+
+    @SuppressWarnings("unused")
```
dmatej, line 182: I see it is used ;-)

avpinchuk: Bad habit ;-)
Fixed.

dmatej (approved):

Issue #24367, opened 2023-04-02
Make GlassFish build reproducible by Maven rules


dmatej:
Based on the discussion with @hboutemy in #24366 we have to reach the state when two consecutive builds of the same codebase will have binary equal results.

Simple test - the following command should produce the same artifact size and the same checksum when you execute it twice: 
```
mvn clean install -q -T4C -Pfastest && ll ~/.m2/repository/org/glassfish/main/distributions/glassfish/7.0.4-SNAPSHOT/glassfish-7.0.4-SNAPSHOT.zip && sha512sum ~/.m2/repository/org/glassfish/main/distributions/glassfish/7.0.4-SNAPSHOT/glassfish-7.0.4-SNAPSHOT.zip
```

### What does it mean

1. Remove all usages of `project.build.outputTimestamp`  property including it's definitions (it is inherited too with a nonsense value of 2020-...)
2. Remove all usages of `git.build.time` property including it's definitions
3. Remove the timestamp related configuration added in 537f247aefd0ed19a040181135fcfe107906b388
4. Remove the code related to timestamp
5. Find out why the glassfish.zip differs and fix it (currently I don't have any idea why is it happening)

Relates also to #24284 and #24316

### See also

- https://maven.apache.org/guides/mini/guide-reproducible-builds.html#faq
- https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/glassfish/main/README.md
- https://reproducible-builds.org/

### Additional context

There are two conflicting requirements:
* To have unique identification of the build in the released artifact.
* To have repeatable build, so the same codebase is transformed to the same binary.

We can't have both and current state of things is that the repeatability is preferred. As famous czech nonexisting hero Jara Cimrman told, "You can disapprove, but that's all you can do against it." ;-)

pzygielo:
https://github.com/eclipse-ee4j/glassfish/pull/24462

pzygielo:
https://github.com/eclipse-ee4j/glassfish/pull/24373

pzygielo:
https://github.com/eclipse-ee4j/glassfish/pull/24372

pzygielo:
https://github.com/eclipse-ee4j/glassfish/pull/24463

dmatej:
Seems all related PRs are merged, so I am closing the issues as fixed.

PR #24372, opened 2023-04-09
disable date in hk2 generated files

hboutemy:
fixes 133 non-reproducible jars after #24366 has been merged
(calculated by running ` mvn clean source:jar install -Pfastest && mvn clean source:jar package -Pfastest artifact:compare`)

after this PR, 710 are ok, 33 remain to be fixed, I'm continuing...

notice that hk2 release is necessary, because there is a bug that ignores the plugin config in 3.0.4: see https://github.com/eclipse-ee4j/glassfish-hk2/pull/821


pzygielo:

dmatej (approved):
I would agree to make it a default value.

PR #24384, opened 2023-04-22
Fix redirections in Admin Console

OndroMih:
Fix redirections when:
* application deployed, a blank page was displayed
* application undeployed, a blank page was displayed

Extract links to static resources to a string resource so that the links are defined in a single place. It's then easier to change the link, e.g. to increase the version number query param.

Added version number query param, so that the script adminjsf.js is loaded from the server and not from cache. An old version could be cached for about a month and it contains a bug. Without this users would have to wait to benefit from these bug fixes one month after they last used the previous version of Admin Console.

### DIFF
--- a/appserver/admingui/common/src/main/resources/org/glassfish/common/admingui/Strings.properties
+++ b/appserver/admingui/common/src/main/resources/org/glassfish/common/admingui/Strings.properties
@@ -15,6 +15,11 @@
 # SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
 #
 
+## URLs to static resources. 
+# The query parameter 'v' should be increased 
+# every time the resource was updated to bypass browser caching
+adminjsf.url=/resource/common/js/adminjsf.js?v=2
+help.css.url=/resource/common/help/help.css?v=1
 
 ## 'server' is the name of the server instance. No Not Translate it.
 tree.adminServer=server (Admin Server)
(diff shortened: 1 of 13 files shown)


dmatej (approved):

appserver/admingui/common/src/main/resources/org/glassfish/common/admingui/Strings.properties
```diff
@@ -15,6 +15,11 @@
 # SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
 #
 
+## URLs to static resources.
```
dmatej, line 18: I don't like it as it is rather a hack than a fix. But while there is still the wrong "pragma" in AdminUI working with caches this way ...
Perhaps we should set something like this instead: https://stackoverflow.com/questions/6208715/pragma-no-cache-and-javascript

OndroMih: Yes, it makes sense. But it's not the point of this PR and could be done later, what do you think? Once we resolve it in a better way, we can remove the additional query parameter and we will need to do it only in a single place.

arjantijms (approved):

appserver/admingui/common/src/main/resources/org/glassfish/common/admingui/Strings.properties
dmatej: That's why I approved it. Generally we should do fixes correctly and just once, not hack and postpone to forever. But with the code in current state it might be more difficult, so lets do this step.

Issue #24262, opened 2023-01-24
Check goals for thread safety


pzygielo:
Following goals of `config-generator`:
- `:generate-injectors`
- `:generate-test-injectors`

are used by the build and cause maven to issue warning during parallel build.

Check, and if possible mark them as being `@threadSafe`.

-----

- #24258

dmatej:
Seems this was resolved in 7.0.5 by transformation of the maven plugin to annotation processor, see https://github.com/eclipse-ee4j/glassfish/pull/24399

PR #24393, opened 2023-04-25
Fix ConfigInjector metadata generation in config-generator maven plugin

avpinchuk:
Generated `ConfigInjector`'s metadata does not include elements and attributes from an indirectly implemented interfaces.

This fixes it.

**Note**: we should build the GlassFish with this fixed plugin, not with old 2.5.0-b53.

### DIFF
--- a/nucleus/hk2/config-generator/src/main/java/org/jvnet/hk2/config/generator/ConfigInjectorGenerator.java
+++ b/nucleus/hk2/config-generator/src/main/java/org/jvnet/hk2/config/generator/ConfigInjectorGenerator.java
@@ -297,7 +297,7 @@ public class ConfigInjectorGenerator extends AbstractProcessor {
                     }
                 }
 
-                for (TypeMirror it : clz.getInterfaces())
+                for (TypeMirror it : t.getInterfaces())
                     q.add((TypeElement) ((DeclaredType)it).asElement());
 
                 if (ElementKind.CLASS.equals(t.getKind())) {


dmatej:
@arjantijms We need to release the new version of the plugin too.

dmatej (approved):

arjantijms (approved):

pzygielo:
> @arjantijms We need to release the new version of the plugin too.

Maybe
- https://github.com/eclipse-ee4j/glassfish/issues/24262

this could be considered as well?

avpinchuk:
@pzygielo, I thought about this. But it will take some time

dmatej:
@arjantijms did not release the plugin for 7.0.4 -> this fix is not inside 7.0.4. What are the consequences for this GlassFish version?

avpinchuk:
No consequences now.

I found a bug while refactoring the config API. Some interfaces annotated `@Configured` implements both `Resource` and `BindableResource`, whiсh implements `Resource` too. I we left only `BindableResource` in the list of implemented interfaces, then we get an error.

To reproduce, remove `Resource` from implemented interfaces in `org.glassfish.connectors.config.AdminObjectResource`. Left only `BindableResource`, then `mvn clean install` the GF. You get an error in unit tests.

avpinchuk:
> @arjantijms did not release the plugin for 7.0.4 -> this fix is not inside 7.0.4. What are the consequences for this GlassFish version?

I want to update the plugin, replace old code, check for the thread-safety etc. What do you think about it?

dmatej:
> > @arjantijms did not release the plugin for 7.0.4 -> this fix is not inside 7.0.4. What are the consequences for this GlassFish version?
> 
> I want to update the plugin, replace old code, check for the thread-safety etc. What do you think about it?

Can you wait for a moment? Just tomorrow. I would try to migrate it to the glassfish build maven plugin. It has a dependency on GF module, but it seems it is just for it's transitive dependencies, so I will try to make it narrow. 
Then next week we could release a next generation of the plugin.

avpinchuk:
> Can you wait for a moment?

No problem. I have a few more bugs in my stash :-)

dmatej:
So, I was wrong. The plugin depends on GlassFish's hk-config classes and GlassFish depends on the plugin. That also explains why we still use the old version of the plugin. I am thinking what we can do with that ...

- Move hk2-config to the refactored gf plugin as it's jar file. It doesn't have any GF dependencies, so it would be safe and correct.
- Move just the plugin goal and make it dependant on older GF version. Bad design.
- Keep it here. Unusable wrong design with recursion.

Basically I can do the first option until the end of this week and we can try it before merging PRs.

avpinchuk:
> So, I was wrong. The plugin depends on GlassFish's hk-config classes and GlassFish depends on the plugin. That also explains why we still use the old version of the plugin. I am thinking what we can do with that ...

At a first glance, we could try to replace all `Class` occurrences with FQN. But it is not exactly. Need further investigation.

PR #24395, opened 2023-04-27
Allow ALL value for loggers in Admin Console.

OndroMih:
A fix of a serious bug in Admin Console:
* When Module Log Levels modified (e.g. a new logger is added), `GlassFishLogHandler` is set to OFF, which means that logging to `server.log` stops.

Before this, Admin Console was showing OFF level instead of ALL level, and saved the OFF level even  if a user didn't change this intentionally.
ALL is used for GlassFishLogHandler by default, and, according to Javadoc, it's also a valid value for logger levels, meaning that all messages are logged.

Keep the OFF default level for new loggers as before. 

This is what Admin Console was showing when `org.glassfish.main.jul.handler.GlassFishLogHandler.level=ALL` in `logging.properties` (the default value):

image

Now it's shows:
image

When a new logger is added, the default level is OFF as before. I just made this configurable and not dependant on the order of the values in the drop down:
image

### DIFF
--- a/appserver/admingui/common/src/main/resources/configuration/loggerLevels.jsf
+++ b/appserver/admingui/common/src/main/resources/configuration/loggerLevels.jsf
@@ -44,7 +44,8 @@
             loggerLevels="#{requestScope.loggerLevels.data.extraProperties.logLevels}"
             loggerList="#{requestScope.tableList}");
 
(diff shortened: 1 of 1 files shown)


dmatej (approved):
Maybe we could also kick out EMERGENCY and ALERT, but that is not target of this fix. Thanks!

PR #24398, opened 2023-04-28
Fix collected log files retrieval

### LINKED ISSUE #24071: Asadmin collect-log-files --retrieve doesn't work
TODO: reproducer including creating instances

```
/app/appservers/glassfish7/bin/asadmin collect-log-files --target Cluster1 --retrieve true /tmp/space/output
Log files are downloaded for test-instance.
Log files are downloaded for ClusterServer1.
Log files are downloaded for ClusterServer2.
Created Zip file under /tmp/space/output/log_2022-08-06_14-57-53.zip.

```

- all files are empty
- unzips the zip file - I don't want it

avpinchuk:
This is a bug fix.

This should close #24071.

Signed-off-by: Alexander Pinčuk <alexander.v.pinchuk@gmail.com>

### DIFF
--- a/appserver/tests/admin/tests/src/test/java/org/glassfish/main/admin/test/ClusterITest.java
+++ b/appserver/tests/admin/tests/src/test/java/org/glassfish/main/admin/test/ClusterITest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022, 2023 Contributors to the Eclipse Foundation
  * Copyright (c) 2012, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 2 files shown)


avpinchuk:
@dmatej, what you think about that?

dmatej:
Wow, thanks! Can you add also a test to ClusterITest (the asadmin one?)

dmatej (approved):
Manually verified that
* it can download domain's server.log file
* it can download domain's server.log file and cluster's log files (perhaps it should not download the domain's file, but we would have to change `--target` semantics and that wasn't part of this issue; then we would need something like "targets" with regexp ...?)
* it fails when I ask just for instance's server.log file if it is a part of a cluster

avpinchuk:
> * it can download domain's server.log file and cluster's log files

Yes, as exactly stated in the Reference Manual

PR #24399, opened 2023-04-30
HK2 Config plugin removed, we use just the annotation processor

dmatej:
This PR was squashed as originally I tried to move the plugin to the glassfishbuild maven plugin, and when I did so, I have found that basically the work it does can be done much faster and easier by the maven compiler plugin. So I moved it back again, redesigned its dependencies, and used it.

Finally those intermediate steps are just a noise, so I will probably squash it myself in few days.

Effects:

* Faster build
* More readable build
* Errors in annotation processing would be thrown, they would not be swallowed, and the build would fail
   * Originally there were several places which crashed, but the original plugin ignored those failures. 
   * Fortunately those failures (except one) did not have any effect, because the module did not use the feature. 
   * ConcurrencyResource was missing an interface (the only one case)
* Tiger types were reduced to just several classes, we should consider their replacement later
* Usage of the glassfishbuild-maven-plugin:echo was removed, because it just logged what was already logged by maven. This thing is a legacy helper, I think that Maven2 did not print these paths.

Next generation, what we could do with this:
* Instead of generating these injector classes, we could add some runtime annotation processing which could do that automatically, without any need of compilation steps. But I don't know if it is possible, at this point it is just an idea.


arjantijms:
@russgold WDYT?

dmatej:
I think that originally the plugin was required because maven compiler plugin did not support this "layered" processing. But now it does.

arjantijms (approved):

arjantijms:
LGTM, but still wonder what @russgold thinks.

PR #24405, opened 2023-05-06
Trivial fixes of logging

dmatej:
- 7.0.4 prints some errors

### DIFF
--- a/nucleus/security/core/src/main/java/com/sun/enterprise/security/admin/cli/SecureAdminConfigMonitor.java
+++ b/nucleus/security/core/src/main/java/com/sun/enterprise/security/admin/cli/SecureAdminConfigMonitor.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2023 Contributors to the Eclipse Foundation
  * Copyright (c) 2010, 2021 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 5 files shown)


nucleus/security/core/src/main/java/com/sun/enterprise/security/admin/cli/SecureAdminConfigMonitor.java
```diff
@@ -47,15 +47,14 @@ public class SecureAdminConfigMonitor implements ConfigListener {
 
     private static final String restartRequiredMsg = Strings.get("secure.admin.change.requires.restart");
 
-    /*
-     * Must inject Domain to get notifications of SecureAdmin changes.  We
-     * cannot inject SecureAdmin itself because it might be null and that
+    /**
```
pzygielo, line 50: Why promote it to javadoc?

pzygielo (approved):

nucleus/security/core/src/main/java/com/sun/enterprise/security/admin/cli/SecureAdminConfigMonitor.java
dmatej: To make the popup visible in Eclipse. Otherwise I would use `//` instead of the "block comment".
Another question is if the content is still true...?

arjantijms (approved):

PR #24420, opened 2023-05-19
Add several recommended guidelines to the security guide

hs536:
add guidelines in response to a report of an security issue when using old JDKs and when unsecured listener ports are exposed.


hs536:
@smillidge Could you check the contents before review?

dmatej (approved):

smillidge (approved):

PR #24423, opened 2023-05-20
Fix resetting security context if principal is the same

### LINKED ISSUE #24196: securityContext.isCallerInRole() incorrectly returning false after authenticating until next page reload

## Environment Details
* GlassFish 7 M10
* JDK version: OpenJDK Runtime Environment Zulu17.36+13-CA (build 17.0.4+8-LTS)
* OS:Ubuntu 22.04.1 LTS
* Database: N/A

----------

## Problem Description
The same issue that was reported for Payara: https://github.com/payara/Payara/issues/4734

> Checking with `securityContext.isCallerInRole("admin")` incorrectly returns false whereas `request.isUserInRole("admin")` returns true. Refreshing the page results in `securityContext.isCallerInRole("admin")` then correctly returning true. Note, the EJB container is affected by this bug, if a Singleton bean requires the admin role, after logging in the principal (which was created) is "missing" the role until you refresh the page.

## Steps to reproduce

The same as in https://github.com/payara/Payara/issues/4734 with this reproducer app (modified for Jakarta EE 10): 
[security-bug.zip](https://github.com/eclipse-ee4j/glassfish/files/10095860/security-bug.zip)

## Impact of Issue
Type: Bug
Component: Security
Priority: high

arjantijms:
If the principal in the request is equal to the principal in the current security context, there is no need to reset that context. Doing this can even lead to losing the roles.

Has also fixed #24196

### DIFF
--- a/appserver/web/web-glue/src/main/java/com/sun/web/server/J2EEInstanceListener.java
+++ b/appserver/web/web-glue/src/main/java/com/sun/web/server/J2EEInstanceListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, 2022 Contributors to Eclipse Foundation.
+ * Copyright (c) 2021, 2023 Contributors to Eclipse Foundation.
  * Copyright (c) 1997, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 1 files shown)


appserver/web/web-glue/src/main/java/com/sun/web/server/J2EEInstanceListener.java
```diff
@@ -203,7 +203,7 @@ else if (baseHttpServletRequest instanceof RequestFacade) {
 
                 if (principal != null && principal == basePrincipal && principal.getClass().getName().equals(WEB_PRINCIPAL_CLASS)) {
                     securityContext.setSecurityContextWithPrincipal(principal);
-                } else if (principal != basePrincipal) {
+                } else if (principal != basePrincipal && principal != getCurrentCallerPrincipal()) {
```
avpinchuk, line 206: Caller principal shouldn't be same as a principal in the current security context or souldn't be equals?

arjantijms: I think it's safe. We're looking at the exact same principal instance, which indicates that the security context has already been setup.

After authentication (the Jakarta Authentication path) both the request and the security context are set using the exact same principal (the one that was set in the `Subject` by the `ServerAuthModule`. This is exactly the case where we don't want to setup the security context again.

avpinchuk (approved):

PR #24424, opened 2023-05-21
Integration tests for cluster deployment

OndroMih:
Covers also the fix in https://github.com/eclipse-ee4j/glassfish/pull/24421


appserver/tests/admin/tests/pom.xml
```diff
@@ -141,6 +141,28 @@
...
+                    <execution>
+                        <id>test-app-for-ClusterITest</id>
+                        <phase>process-test-classes</phase>
+                        <goals>
+                            <goal>war</goal>
+                        </goals>
+                        <configuration>
+                            <primaryArtifact>false</primaryArtifact>
+                            <classifier>clusterit</classifier>
+                            <failOnMissingWebXml>false</failOnMissingWebXml>
```
pzygielo, line 157: https://maven.apache.org/plugins/maven-war-plugin/war-mojo.html#failOnMissingWebXml
> Starting with 3.1.0, this property defaults to false if the project depends on the Servlet 3.0 API or newer.

Since this module uses `3.3.2`, could we get rid of this explicit element then?

OndroMih: The catch is that the dependency on Servlet API is in `test` scope. We could remove `<failOnMissingWebXml>` if we change the scope to `compile`. But I prefer keeping it this way, because it's explicit. If we change servlet scope to compile while all other dependencies are in `test`, it's not clear why it is like that.

avpinchuk:
@OndroMih, maybe better get rid of `maven-war-plugin` and simply use programmatic `war` creation as in `appserver/tests/admin/tests/src/test/java/org/glassfish/main/admin/test/rest/ApplicationITest.java`? Maybe we should move `getWar` and `getEar` methods from `RestTestBase` class to a separate utility class and also reuse `HelloServlet`.  Content of the test `war` does not matters, right?

OndroMih:
Yes, @avpinchuk , it's better to create the WAR programmatically. I didn't realize that ShrinkWrap is already used in the project. I can even just copy the getWar method to keep things simple, I only need a simple empty WAR with a simple index.html file, which can be added as a text resource.

Thanks for the hint.

avpinchuk (approved):

dmatej (approved):

PR #24429, opened 2023-05-22
Fixes #24223 Make PrincipalCallback Serializable & prevent JMX exception

### LINKED ISSUE #24223: PrincipalCallback Not Serializable
## Environment Details
* GlassFish Version (and build number):  7.0.0  (build master-b1292-g9052d5b 2020-12-19T17:24:00+0000)
* JDK version: java-11-openjdk
* OS: Linux
* Database: None

----------

## Problem Description

The `PrincipalCallback` class within the `AdminLoginModule` is not serializable. In general, this is not a problem in itself. However, the `AuthenticationCallbackHandler` in GlassFish can raise an `UnsupportedCallbackException` which contains the unsupported callback within its properties. The following snipped shows the relevant code from the `AuthenticationServiceImpl` class:

```java
private static class AuthenticationCallbackHandler implements CallbackHandler {
        private final String user;
        private final char[] pass;

        public AuthenticationCallbackHandler(String username, char[] password) {
            user = username;
            pass = password;
        }

        protected String getUsername() { return user; }
        protected char[] getPassword() { return pass; }

        @Override
        public void handle(Callback[] callbacks) throws UnsupportedCallbackException {
            for (Callback callback : callbacks) {
                if (callback instanceof NameCallback) {
                    ((NameCallback) callback).setName(user);
                } else if (callback instanceof PasswordCallback) {
                    ((PasswordCallback) callback).setPassword(pass);
                } else {
                    // TODO - Have configuration setting for throwing exception
                    throw new UnsupportedCallbackException(callback,
                        "AuthenticationCallbackHandler: Unrecognized Callback "
                            + callback.getClass().getName());
                }
            }
        }
    }
```

When connecting via *RMI* (*JMX*) and triggering the `UnsupportedCallbackException`, a `NotSerializableException` is raised. This is of course because the `UnsupportedCallbackException` is written to the *RMI* stream, but its callback property is not serializable. This is confusing and hard to debug, because the actual issue of the exception gets concealed.

## Steps to reproduce

1. Download [GlassFish server](https://projects.eclipse.org/projects/ee4j.glassfish/downloads)
2. Start the server:  `./glassfish-7/bin/asadmin start-domain`
3. Set an admin password `./glassfish-7/bin/asadmin change-admin-password`
4. Connect via `jconsole` without specifying credentials. Make sure to use the `-debug` flag to see the resulting stack trace

image

## Impact of Issue

As mentioned above, there is no real impact. When connecting with credentials everything works fine. When connecting without credentials to a non credential protected instance, everything works fine too. However, as this issue demonstrates, there are edge cases that lead to unexpected behavior.

It is also worth noting that all default callbacks from the `javax.security.auth.callback` package are serializable too. From my point of view `PrincipalCallback` should be made serializable. But this requires also all relevant `Principal` implementations to be serializable. Should workout for most classes, but there might be edge cases too :man_shrugging:

arjantijms:
Fixes #24223

### DIFF
--- a/nucleus/admin/util/src/main/java/com/sun/enterprise/admin/util/AdminLoginModule.java
+++ b/nucleus/admin/util/src/main/java/com/sun/enterprise/admin/util/AdminLoginModule.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2023, 2023 Contributors to the Eclipse Foundation. All rights reserved.
  * Copyright (c) 2012, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -19,6 +20,9 @@ package com.sun.enterprise.admin.util;
 import com.sun.enterprise.config.serverbeans.Domain;
 import com.sun.enterprise.config.serverbeans.SecureAdmin;
 import com.sun.enterprise.config.serverbeans.SecureAdminPrincipal;
+
+import java.io.Serial;
+import java.io.Serializable;
 import java.security.Principal;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


avpinchuk:
@arjantijms, `@Serial` annotation requires Java 14, but GF 7 should work with Java 11

arjantijms:
> @arjantijms, `@Serial` annotation requires Java 14, but GF 7 should be able to compile with Java 11

You are right, and I was wrong. Thank you!

avpinchuk (approved):

senivam (approved):

dmatej (approved):

PR #24432, opened 2023-05-25
Admin Console: Update instance status asynchronously

OndroMih:
Pages for list of standalone instance, clusters and nodes show information about instance status. In real life deployments, due to network issues, it can take a long while to retrieve the status of the instances.

This change defers checking the statuses to an ajax call, one ajax call per each instance. This could be optimized to retrieve statuses of all instances in a single ajax call to save admin threads but this solution is faster because it polls the instances in parallel. And it's also easier to write.

What it does:
* in all cases, the page is displayed almost immediately, with statuses of all instances displayed as UNKNOWN
* AJAX calls then update the statuses after the page is loaded

When everything is fast, all works as before - correct statuses are displayed. If retrieving statuses is slow, before this change it took a long time to display the whole page until statuses of all instances were retrieved. After this change, the page is displayed instantly with UKNOWN statuses, and statuses are updated with correct values after a while. Each status can be updated independently, so some instances can show Running or Stopped, while some other will keep showing UNKNOWN for a longer while.

Examples:

Standalone instance when statuses take longer time to retrieve, before they are retrieved:

image

After they are retrieved:

image

Cluster instances (some are retrieved, some show UNKNOWN because they take a bit longer to retrieve the status):

image

And nodes, with all statuses retrieved:

image

### DIFF
--- a/appserver/admingui/cluster/src/main/resources/cluster/clusterInstances.jsf
+++ b/appserver/admingui/cluster/src/main/resources/cluster/clusterInstances.jsf
@@ -49,6 +49,7 @@
         createMap(result="#{requestScope.listInstanceAttrMap}");
         setPageSessionAttribute(key="clusterInstancesPage" value="#{true}" )
         mapPut(map="#{requestScope.listInstanceAttrMap}" key="whichtarget" value="#{pageSession.encodedClusterName}")
+        mapPut(map="#{requestScope.listInstanceAttrMap}" key="nostatus" value="true")
         gfr.getInstancesStatus();
         gf.getMapKeys(Map="#{pageSession.instanceStatusMap}" Keys="#{pageSession.instancesName}");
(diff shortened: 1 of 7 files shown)


dmatej (approved):
Wow, awesome!

avpinchuk (approved):
Looks great!

hs536 (approved):
LGTM.
I'm a little worried that users might think there was something wrong with getting the status because the first information they see is "Unknown". but I have no other suitable word...

OndroMih:
I changed the temporary status indication, so that it doesn't show Unknown status, but Retrieving status... instead, with a progress bar icon. This depends on https://github.com/eclipse-ee4j/glassfish-woodstock/pull/1439, otherwise the progress bar icon isn't displayed and alternative text is shown instead.

Here's how it looks like with the progress icon:

image

PR #24439, opened 2023-05-30
Merge 7.0.5 branch into master

arjantijms:


avpinchuk (approved):
@arjantijms, felix goal failed, it's ok?

nucleus/parent/pom.xml
```diff
@@ -499,12 +499,12 @@
             <dependency>
                 <groupId>org.apache.felix</groupId>
                 <artifactId>org.apache.felix.main</artifactId>
-                <version>7.0.5</version>
+                <version>7.0.6-SNAPSHOT</version>
```
hs536, line 502: Since jenkins uses sed to rewrite the version, it seems to have accidentally rewritten the version of an unrelated artifact. (Maybe we should update the jenkins job to use maven release plugin or maven version plugin...

nucleus/parent/pom.xml
```diff
@@ -499,12 +499,12 @@
...
             </dependency>
             <dependency>
                 <groupId>org.apache.felix</groupId>
                 <artifactId>org.apache.felix.framework</artifactId>
-                <version>7.0.5</version>
+                <version>7.0.6-SNAPSHOT</version>
```
hs536, line 507: also here

nucleus/parent/pom.xml
pzygielo: - I remember this #24210 (with very similar 1d3c2a7e51b10a5792a11101b90094722e6c21c3)

dmatej: There's already a bash script updateVersion.sh, but it seems Jenkins doesn't use it.

PR #24447, opened 2023-06-03
Update latest docs to 7.0.5 on the website

OndroMih:
Until we release the 7.0.5 artifact of docs, we're using the 7.0.5-SNAPSHOT artifact.
Also updated RELEASE.md to include the step to update the docs version

### DIFF
--- a/docs/publish/pom.xml
+++ b/docs/publish/pom.xml
@@ -130,7 +130,7 @@
                                         <artifactItem>
                                             <groupId>org.glassfish.docs</groupId>
                                             <artifactId>distribution</artifactId>
-                                            <version>${glassfish.version.7x}</version>
+                                            <version>${glassfish.version.7x.artifact}</version>
                                             <outputDirectory>
                                                 ${docs.7x.dir}
                                             </outputDirectory>
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -42,7 +42,10 @@ In this example we assume 7.0.3.
     - click [Build] button 
 23. Create the release on Github: https://github.com/eclipse-ee4j/glassfish/releases click "draft a new release"
 24. Create the release on Eclipse: https://projects.eclipse.org/projects/ee4j.glassfish click "create a new release"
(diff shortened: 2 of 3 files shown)


avpinchuk:
@OndroMih, when we'll release the new version, we will need change `docs/publish/pom.xml` again? Or am I misuderstanding your PR?

docs/publish/pom.xml
```diff
@@ -130,7 +130,7 @@
                                         <artifactItem>
                                             <groupId>org.glassfish.docs</groupId>
                                             <artifactId>distribution</artifactId>
-                                            <version>${glassfish.version.7x}</version>
+                                            <version>${glassfish.version.7x.artifact}</version>
```
hs536, line 133: Does this change affect the release process?

OndroMih: No, this change is only to reflect what is happening right now. Currently, we don't publish the docs artifact as part of the release process, the docs are only published daily for the snapshot version:
* There's only 7.0.0 version in Maven Central: https://search.maven.org/search?q=g:org.glassfish.docs%20AND%20a:distribution
* There's only 7.0.0 version and snapshot versions in Jakarta Maven repo: https://jakarta.oss.sonatype.org/content/groups/staging/org/glassfish/docs/distribution/

So we need to download the snapshot artifact instead of the final artifact. The rest is the same as before.

If we published the docs artifact as part of the release process, this wouldn't be necessary and we could publish the docs from the ${glassfish.version.7x} version.

OndroMih:
> @OndroMih, when we'll release the new version, we will need change `docs/publish/pom.xml` again? Or am I misuderstanding your PR?

Yes, we'll need to update the version of the docs together with other updates on the website (e.g. updating the downloads and news sections). We didn't do it before, therefore the docs for the "Current release" always point to 7.0.0 version: https://glassfish.org/docs/. We need to update the `glassfish.version.7x` to update "Current release" docs to point to the new version of the docs.

arjantijms (approved):

PR #24448, opened 2023-06-04
Fix bytecode transformation in the ASURLClassLoader

avpinchuk:
Fix `ClassCircularityError ` in the `ASURLClassLoader`.

This is looks like [#24451](https://github.com/eclipse-ee4j/glassfish/pull/24451).

### DIFF
--- a/nucleus/common/common-util/src/main/java/com/sun/enterprise/loader/ASURLClassLoader.java
+++ b/nucleus/common/common-util/src/main/java/com/sun/enterprise/loader/ASURLClassLoader.java
@@ -50,7 +50,6 @@ import java.security.cert.Certificate;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashMap;
@@ -76,9 +75,9 @@ import static java.lang.ThreadLocal.withInitial;
 import static java.util.logging.Level.INFO;
 
 /**
- * Class loader used by the ejbs of an application or stand alone module.
+ * Class loader used by the ejbs of an application or stand-alone module.
  *
- * This class loader also keeps cache of not found classes and resources.
+ * <p>This class loader also keeps cache of not found classes and resources.
  *
  * @author Nazrul Islam
  * @author Kenneth Saks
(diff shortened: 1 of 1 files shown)


senivam (approved):
generally, it's recommended in Javadoc to close any HTML tags (like `<p>`) but as soon as GF already has a number of not closed HTML tags in Javadoc and somehow deals with it, this is just a notice.

avpinchuk:
@senivam, thank you for your remark!

dmatej:
> generally, it's recommended in Javadoc to close any HTML tags (like `<p>`) but as soon as GF already has a number of not closed HTML tags in Javadoc and somehow deals with it, this is just a notice.

I hear for the first time about such recommendation - Javadoc is based on HTML3.
See https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#format

I tried to google it a bit, I found some issue for Idea to disable enforcing closing the `<p>` element. Perhaps that bug was the source?

nucleus/common/common-util/src/main/java/com/sun/enterprise/loader/ASURLClassLoader.java
```diff
@@ -118,6 +118,9 @@ public class ASURLClassLoader extends GlassfishUrlClassLoader
 
     private final ArrayList<ClassFileTransformer> transformers = new ArrayList<>(1);
 
+    /** Bytecode transformation flag */
+    private static final ThreadLocal<Boolean> BYTECODE_TRANSFORMATION = withInitial(() -> false);
```
dmatej, line 122: I think that thread locals should be removed at some point, at least when the classloader is closed. Then they shouldn't be static as classes are defined by the classloader. SonarQube told me :-)
Btw why do we use ThreadLocal here? findClass is called by ClassLoader.loadClass in a synchronized block (synced via getClassLoadingLock).
It seems to me that now it causes just a side effect with multiple threads ... but I can't imagine all situations, because the getClassLoadingLock logic is a bit "magic" for me.

avpinchuk: @dmatej, yo're absolute right! Because this CL is not registered as a parallel-capable, they obtain itself as lock in the `getClassLoadingLock`. And I think it's enough a simple non-volatile boolean flag.

senivam (approved):

dmatej (approved):

Issue #23565, opened 2021-08-24
Remove com.sun.enterprise.util.ProcessExecutor


arjantijms:
`com.sun.enterprise.util.ProcessExecutor` is set to be deprecated and to be replaced by `com.sun.enterprise.universal.process.ProcessManager`.

It's however still in use.

avpinchuk:
Competed with #24446, #24453, #24458, #24467 and #24468.

PR #24453, opened 2023-06-08
Remove com.sun.enterprise.util.ProcessExecutor

mz1999:
Replaced `ProcessExecutor` with `ProcessManager` and removed `ProcessExecutor` as suggested in the [issue](https://github.com/eclipse-ee4j/glassfish/issues/23565).

`ProcessExecutor` redirects standard output and standard error to files, while `ProcessManager` stores them in memory, so `ProcessManager.getStdout()` and `ProcessExecutor.getLastExecutionOutput` are equivalent, and `ProcessManager.getStderr()` and `ProcessExecutor.getLastExecutionError` are equivalent.

`ProcessExecutor` has a default timeout, so when replacing with `ProcessManager`, you need to explicitly set the timeout.


dmatej (approved):

avpinchuk (approved):

arjantijms (approved):
Looks great! wow!

PR #24455, opened 2023-06-09
Switch to a fork of jsch (com.github.mwiede) and upgrade to 0.2.9

OndroMih:
The 0.2.9 version of jsch fixes a defect when creating an SSH node with a private key instead of password.

The repackaged artifact is no longer needed, the upstream artifact includes a correct OSGi manifest. 

I closed the previous PR for the repackaged repository: https://github.com/eclipse-ee4j/glassfish-repackaged/pull/21

I also added a logger for Jsch. The logger shifts log level of Jsch, which logs a lot of INFO messages that look more like DEBUG messages to me, and WARNING messages that look like INFO to me. The default level in logging .properties is INFO, which now logs only a message about adding a host to the list of known hosts on every SSH login, which is not very relevant but reasonable.

### DIFF
--- a/nucleus/admin/template/src/main/resources/config/logging.properties
+++ b/nucleus/admin/template/src/main/resources/config/logging.properties
@@ -139,6 +139,7 @@ jakarta.enterprise.system.core.naming.level=INFO
 jakarta.enterprise.system.core.security.level=INFO
 jakarta.enterprise.system.core.security.web.level=INFO
 jakarta.enterprise.system.core.selfmanagement.level=INFO
+jakarta.enterprise.system.core.ssh.level=INFO
 jakarta.enterprise.system.core.transaction.level=INFO
 jakarta.enterprise.system.jmx.level=INFO
 jakarta.enterprise.system.security.ssl.level=INFO
--- a/nucleus/cluster/ssh/src/main/java/org/glassfish/cluster/ssh/launcher/JavaSystemJschLogger.java
+++ b/nucleus/cluster/ssh/src/main/java/org/glassfish/cluster/ssh/launcher/JavaSystemJschLogger.java
@@ -19,6 +19,19 @@ import com.jcraft.jsch.JSch;
 import java.lang.System.Logger;
 import java.lang.System.Logger.Level;
 
+/* JSCH triggers a lot of unimportant messages, which have a lower level than issued by JSCH
+  For example, for a single SSH connection, it triggers 1 WARNING about adding a host to the list of known hosts,
+  47 INFO messages, and 6 DEBUG messages. Therefore we decrease their level to match their real severity.
+
+  The log levels from JSCH are mapped to these levels in GlassFish:
+
+  JSCH DEBUG level -> TRACE level in GlassFish
+  JSCH INFO level -> DEBUG level in GlassFish
+  JSCH WARN level -> INFO level in GlassFish
+  JSCH ERROR level -> WARNING level in GlassFish
+  JSCH FATAL level -> ERROR/SEVERE level in GlassFish
+
+ */
 public class JavaSystemJschLogger implements com.jcraft.jsch.Logger {
 
     private final Logger logger;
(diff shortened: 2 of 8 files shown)


nucleus/cluster/ssh/src/main/java/org/glassfish/cluster/ssh/launcher/JulJschLogger.java
```diff
@@ -0,0 +1,69 @@
...
+import com.jcraft.jsch.JSch;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class JulJschLogger implements com.jcraft.jsch.Logger {
+
+    private final Logger logger;
+
+    public JulJschLogger() {
+        this.logger = Logger.getLogger(JSch.class.getName());
```
dmatej, line 27: It would be better to use System.Logger instead of direct JUL usage. We are slowly moving in this direction.

nucleus/admin/template/src/main/resources/config/logging.properties
```diff
@@ -140,6 +140,7 @@ jakarta.enterprise.system.core.security.level=INFO
 jakarta.enterprise.system.core.security.web.level=INFO
 jakarta.enterprise.system.core.selfmanagement.level=INFO
 jakarta.enterprise.system.core.transaction.level=INFO
+jakarta.enterprise.system.core.jsch.level=SEVERE
```
dmatej, line 143: Ordering: alphabet ;-)

nucleus/cluster/ssh/src/main/java/org/glassfish/cluster/ssh/launcher/JulJschLogger.java
OndroMih: Good catch. I forgot about this. I switched to System.Logger for Jsch. But I kept the JUL logger in the SSHLauncher which was already there.

nucleus/admin/template/src/main/resources/config/logging.properties
OndroMih: I renamed the logger to `jakarta.enterprise.system.core.ssh` and fixed the ordering ;-)

hs536 (approved):

nucleus/cluster/ssh/src/main/java/org/glassfish/cluster/ssh/launcher/JavaSystemJschLogger.java
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    static Level getSystemLoggerLevel(int jschLevel) {
+        switch (jschLevel) {
+            case DEBUG:
+                return Level.TRACE;
+            case INFO:
+                return Level.DEBUG;
+            case WARN:
```
dmatej, line 59: WARN to INFO and INFO to DEBUG sounds suspicious ... I don't know how those logs look like, but I would expect that JSCH is based on some old presumptions about FATAL in LOG4J or Commons Logging or where I have seen it for the last time some 15 years ago ...

OndroMih: Yes, it's suspicious, but reasonable. Have a look at this log file: [gf-jsch-ping-node.log](https://github.com/eclipse-ee4j/glassfish/files/11750634/gf-jsch-ping-node.log), which contains messages from jsch just for a single ping action to an SSH node:

* 1 INFO message
* 47 FINE messages
* 6 FINER messages

If we mapped WARN to WARNING, INFO to INFO, DEBUG to DEBUG, it would be 1 WARNING, 47 INFO messages and 6 FINE messages. Too many INFO messages for a single SSH connection, IMHO. And the warning message just says that the host was added to the list of known hosts, which, according to me, doesn't qualify as a WARNING but rather as an INFO message. It's logged for every connection, because the list of known hosts isn't preserved between sessions.

Does it now make sense why I made this shift of log levels? :smiley:

OndroMih: To clarify this in the code, I added a comment in this logger class to clarify how the JSCH log levels are mapped to levels in GlassFish and why.

avpinchuk: Seems reasonable, imo

dmatej: Ok, that's why I asked, thanks.

dmatej (approved):

PR #24456, opened 2023-06-09
Add docs distribution artifact to mvn deploy plugin for Jenkins.

OndroMih:
So that it's deployed by Jenkins during a GF release.

I tested locally by running `mvn -Prelease-phase2 deploy` and it looks like it works as it should - runs the deploy plugin on modules `qa`, `nucleus`, and `appserver` as before, and then it runs the deploy plugin on the docs/distribution module, and skips the deploy plugin for all other modules in the docs directory. All the dependencies of the docs `distribution` artifact are optional, therefore they don't have to be available in a maven repository.

### DIFF
--- a/docs/distribution/pom.xml
+++ b/docs/distribution/pom.xml
@@ -36,6 +36,15 @@
         <maven.site.skip>true</maven.site.skip>
     </properties>
 
+    <profiles>
+        <profile>
+            <id>release-phase2</id>
+            <properties>
+                <maven.deploy.skip>false</maven.deploy.skip>
+            </properties>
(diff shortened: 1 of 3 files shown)


dmatej:
Yet context - originally, before upgrade, the docs build was depending on Java8 while GlassFish already used JDK11/17. So now this should be really possible, I will take a look bit later (perhaps someone else can be faster).

hs536:
+1 It would be nice if the version of the latest GlassFish matched the version of the latest documentation.

arjantijms (approved):

PR #24446, opened 2023-06-02
Refactor ProcessManager to Use ProcessBuilder

mz1999:
This PR aims to refactor the `ProcessManager`, transitioning from the usage of `Runtime.exec()` to `ProcessBuilder`. The main objective is to provide better control and configurability over the process execution. 

I have also added some tests for `ProcessManager`.

Based on the suggestions in [issue 23565](https://github.com/eclipse-ee4j/glassfish/issues/23565), I had planned to use the refactored `ProcessManager` instead of `ProcessExecutor`, but I found that I could not simply replace it. The `ProcessExecutor` redirects the standard output to a file, which is different from the mechanism of `ProcessManager` and requires more careful consideration and testing, so I still finished the refactoring of `ProcessManager` first.

PR #24458, opened 2023-06-10
Fixed race condition in ProcessManager

dmatej:
- reader.readLine can return null when there was nothing to read.
- escallate errors to logs, don't hide them.
- fixes issue discovered in https://github.com/eclipse-ee4j/glassfish/pull/24446#issuecomment-1585533239

### DIFF
--- a/nucleus/common/common-util/src/main/java/com/sun/enterprise/universal/process/ProcessManager.java
+++ b/nucleus/common/common-util/src/main/java/com/sun/enterprise/universal/process/ProcessManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022, 2023 Contributors to the Eclipse Foundation
  * Copyright (c) 2009, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -27,7 +27,6 @@ import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
 import java.lang.System.Logger;
 import java.lang.System.Logger.Level;
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
@@ -46,25 +45,20 @@ public class ProcessManager {
     private static final Logger LOG = System.getLogger(ProcessManager.class.getName());
 
     protected final ProcessBuilder builder;
-    private final StringBuffer sb_out;
-    private final StringBuffer sb_err;
+    private String stdout;
+    private String stderr;
     private int timeout;
     private boolean echo = true;
     private String[] stdinLines;
-    private final List<Thread> threads = new ArrayList<>(2);
     private boolean waitForReaderThreads = true;
 
     public ProcessManager(String... cmds) {
         builder = new ProcessBuilder(cmds);
-        sb_out = new StringBuffer();
-        sb_err = new StringBuffer();
     }
 
 
     public ProcessManager(List<String> cmdline) {
         builder = new ProcessBuilder(cmdline);
-        sb_out = new StringBuffer();
-        sb_err = new StringBuffer();
     }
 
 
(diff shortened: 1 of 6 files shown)


nucleus/common/common-util/src/main/java/com/sun/enterprise/universal/process/ProcessManager.java
```diff
@@ -222,51 +211,66 @@ private void waitAwhile(Process process) throws InterruptedException, ProcessMan
...
-    }
-
-
-    static class ReaderThread implements Runnable {
+    static class ReaderThread extends Thread {
         private final BufferedReader reader;
-        private final StringBuffer sb;
+        private final StringBuilder sb;
         private final boolean echo;
+        private boolean stop;
```
avpinchuk, line 218: Hmm, non-volatile flag enough here?

dmatej: I believe it is not required here. If it would be several nanoseconds later it doesn't matter. Perhaps we should make some experiments if it is better to use volatile and slow down some context switching or not to use volatile and slow down the "shutdown" (it can make few iterations more).

avpinchuk: The `stop` flag set to true in the thread running `execute` (in `finish` method), but check in another (reader) thread. I think, visibility problem is possible here. We can get an infinite wait loop.

dmatej: Volatile is not a solution for thread safety, just an instruction to avoid optimization. The value change has just one direction, from false to true, and then with volatile it must be checked in memory before every use, without volatile it can be used from cpu cache between syncing the value state in memory, which is done by the cpu.
Or am I wrong and cpu can use the old value forever?
Can you prove that behavior on some cpu?

avpinchuk: Volatile write give us StoreLoad barrier to flush CPU write buffers. Without it cpu may use value from its local wtite buffer potentially indefinitely, i.e., second thread, running on another core/cpu may never see new value.

I have no cpu other than x86 (amd and intel) ;-(

dmatej: Hmm, I know about related issues if we would have two depending values, but in this case it seems contraproductive to me. But we don't need to hunt nanos here, so for simplicity I can just add it.

I am just curious if we are not overengineering that. Just to remember ie if arms would have such issue... and how to reproduce it.

avpinchuk: In case of two variables even volatile may not be sufficient ;-)

avpinchuk: As a microoptimization we can use AtomicBoolen lazySet/get here

dmatej: I had to read more ... now I am thinking about some tests outside GF related to this problematic :-)
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/invoke/VarHandle.html
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/atomic/AtomicBoolean.html

I used acquire-release, finally.
I tried to google where we could reproduce some issues like never ending loop, somebody mentioned Apple and ARM cpus.

avpinchuk: Great! I think, release/acquire enough here.

avpinchuk:
I clone your branch and still get same error https://github.com/eclipse-ee4j/glassfish/pull/24446#issuecomment-1585533239

dmatej:
Thanks for the navigation, @avpinchuk! What we discovered yet in private discussion - the manager in some parts ignored returned values, in another tried to kill dead process and THEN process the output, in another - for some reason the input values were not processed, but the error was reproducible just in some 0.1% on my machine.
I hope we resolved all of those issues now ...

avpinchuk:
@dmatej, happy to help ;-)

avpinchuk (approved):

arjantijms (approved):

PR #24462, opened 2023-06-11
improve reproducible builds

hboutemy:
continue work started in #24366

one key missing part is glassfish-hk2 3.0.5 to benefit from https://github.com/eclipse-ee4j/glassfish-hk2/pull/821

with this PR merged and glassfich-hk2 3.0.5, reproducibility score should improve a lot: we'll sort out next issues after
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/glassfish/main/README.md


nucleus/parent/pom.xml
```diff
@@ -1074,7 +1074,7 @@
                         <dateFormatTimeZone>UTC</dateFormatTimeZone>
                         <commitIdGenerationMode>full</commitIdGenerationMode>
                         <includeOnlyProperties>
-                            <includeOnlyProperty>git.branch</includeOnlyProperty>
+                            <!--includeOnlyProperty>git.branch</includeOnlyProperty--><!-- not reproducible -->
```
pzygielo, line 1077: If this isn't working for reproducibility - is the `git.closest.tag.commit.count` below?

pzygielo, line 1077: If this is not set - I suppose the https://github.com/eclipse-ee4j/glassfish/blob/f985ae7390941832b0f2f7c6a2ae651723598561/appserver/distributions/glassfish-common/src/main/resources/glassfish/config/branding/glassfish-version.properties#L21 and its usages should be updated.

pzygielo:
With this we're getting unresolved property in output:
```
$ ./glassfish7/glassfish/bin/asadmin version
Version string could not be obtained from Server [localhost:4848].
(Turn debugging on e.g. by setting AS_DEBUG=true in your environment, to see the details.)
Using locally retrieved version string from version class.
Version = Eclipse GlassFish 7.0.6-SNAPSHOT (branch: ${git.branch}, commit: ef2d471b4b11339b8c67f373d60f2fbc8ad5441c, timestamp: 2023-04-02T15:44:03Z)
Command version executed successfully.
```

hboutemy:
oh, I see. Here are the details: this impacts 3 files only, then we can postpone this small issue, there are more important ones remaining (I'll ask about it in another comment):

```
198 / 210 target/reference/glassfish-common-7.0.5.zip appserver/distributions/glassfish-common/target/glassfish-common.zip
├── glassfish7/glassfish/config/branding/glassfish-version.properties
│ @@ -14,15 +14,15 @@
│  #
│  # SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
│  #
│  
│  product.name=Eclipse GlassFish
│  product.name.abbreviation=GF
│  product.version=7.0.5
│ -product.build.git.branch=master
│ +product.build.git.branch=7b49f4eaf669e9a3b4f4648254f9016872455236
│  product.build.git.commit=7b49f4eaf669e9a3b4f4648254f9016872455236
│  product.build.timestamp=2023-04-02T15:44:03Z
│  
│  admin.command.name=asadmin
│  
│  domain.admin.groups=asadmin
│  domain.template.defaultJarFileName=appserver-domain.jar


199 / 210 target/reference/glassfish-7.0.5.zip appserver/distributions/glassfish/target/glassfish.zip
207 / 210 target/reference/web-7.0.5.zip appserver/distributions/web/target/web.zip
```

as you can see, it's just that when you `git checkout`, there is de-facto branch any more: the branch is there only when doing the release. Or if you find a way to git checkout keeping the branch...

perhaps injecting `-Dgit.branch=master` when building can override, I'll try

on `git.closest.tag.commit.count`, I did not find any diff caused by it

hboutemy:
the root cause of a vast majority of issues is definitely fixed by glassfish-hk2 3.0.5 

one issue that I'd like to discuss is that I find many places where I get:
```
├── META-INF/MANIFEST.MF
│ -Originally-Created-By: Apache Maven Bundle Plugin 5.1.8
```
which means that it seems there has been 2 runs of maven-bundle-plugin

and 
```
├── META-INF/maven/org.glassfish.main/hk2-config-generator/pom.properties
│ @@ -1,3 +1,3 @@
│  groupId=org.glassfish.main
│  artifactId=hk2-config-generator
│ -version=7.0.5-SNAPSHOT
│ +version=7.0.5
```
which means that some parts of the reference build we not done on the git tag

Is the precise release process written somewhere, so I can try to rebuild better than the current buildspec?

dmatej:
Ok, I will go for glassfishbuild maven plugin in few days.

pzygielo:

arjantijms (approved):

PR #23433, opened 2021-03-25
#23401 Improve build.id information

hussainnm:
Add git-commit-id-plugin to generate commit info

build.id property will be `BRANCHNAME`-b`COMMITCOUNT`-g`GITCOMMITID` `BUILDTIMESTAMP`

Fixes #23401

PR #24463, opened 2023-06-11
Drop branch from version string/product id info

### LINKED ISSUE #23521: Current usage of `git.closest.tag.commit.count` property in `build.id`
Current usage of `git.closest.tag.commit.count` property:
- if there is no tag in repository, as in [example from Jenkins](https://github.com/eclipse-ee4j/glassfish/pull/23519#issuecomment-877871161) from build 6, `build.id` has the part `*-b-*` (with  _b_ alone only)
- if there is any other ("private") tag in repository - it might be used to determine `git.closest.tag.commit.count`, but it is not present in `build.id`; so if I have the tag on 6eead71, it yields `revid-b0-g6eead71 2021-07-12T08:33:20+0200`; but as tag itself is not included - what is the meaning of `b0`? IMO - it's meaningless. It's different from `git describe` where tag and its distance from described commit IS present: `TEST_TAG-0-g6eead711e`.

I think #23401 could be revisited.

pzygielo:


hs536:
Could you add a little more explanation?
Certainly the branch info seems redundant, but what kind of effect can we get by deleting it?

pzygielo:
I hoped the linked issue would help, especially the comment:
> Yes, the timestamp and probably branch too should be removed.

with which I totally agree.

As this PR is only about branch - here is more closely why:
- https://github.com/eclipse-ee4j/glassfish/pull/24462 - https://github.com/eclipse-ee4j/glassfish/blob/ef2d471b4b11339b8c67f373d60f2fbc8ad5441c/nucleus/parent/pom.xml#L1077 as I would like to have reproducible GF build as well!

As for timestamp, I hope for it to be the next one to go away, for the same reason but also:
- https://github.com/eclipse-ee4j/glassfish/pull/24366#discussion_r1155091652
- https://github.com/eclipse-ee4j/glassfish/pull/24322
- https://github.com/eclipse-ee4j/glassfish/pull/24316

It seems we have a lot of work done for this (IMO not very crucial, but I'm open to be proved otherwise) version string feature. I'd rather see active development in other areas...

Unfortunately I did not object to
- https://github.com/eclipse-ee4j/glassfish/pull/23433
But I committed to myself to repay for this eventually...

And finally - there is
- https://github.com/eclipse-ee4j/glassfish/issues/23521
with no impact since 537f247aefd0ed19a040181135fcfe107906b388, only unused property left.


-----

> what kind of effect can we get by deleting it?

Why to keep it?

dmatej (approved):

hs536:
got it.  so the branch property is inconvenient to make the GlassFish build reproducible.

hs536 (approved):

pzygielo:

PR #24471, opened 2023-06-16
Startserv script starts GlassFish in foreground in a single JVM process

OndroMih:
Instead of starting a launcher with `start-domain --verbose`, which starts GF and then waits for it to finish 
(2 JVM processes running), runs GF in `dry-run` mode and then executes GF in foreground directly, 
using the command line retrieved from dry-run mode.

We need to remove the `-read-stdin` option from the command in the `dry-run` output, otherwise the command waits for input and that's not needed.

This is a preparation for Docker images that will run GlassFish in foreground in a single JVM process to save memory consumed by Docker containers.

I also copied the startserv and stopserv scripts to the top-level bin directory, to make them easier to use and accessible to users.

### DIFF
--- a/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/StartDomainCommand.java
+++ b/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/StartDomainCommand.java
@@ -143,7 +143,18 @@ public class StartDomainCommand extends LocalDomainCommand implements StartServe
                 logger.fine(Strings.get("dry_run_msg"));
                 List<String> cmd = glassFishLauncher.getCommandLine();
                 StringBuilder sb = new StringBuilder();
+                boolean skipLine = false;
                 for (String s : cmd) {
+                    if (s.equals("-read-stdin")) {
+                        // Don't print this option as it's not needed to run the server
+                        // Also skip the next line with "true", which is related to this option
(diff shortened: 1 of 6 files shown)


nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/StartDomainCommand.java
```diff
@@ -143,7 +143,18 @@ protected int executeCommand() throws CommandException {
                 logger.fine(Strings.get("dry_run_msg"));
                 List<String> cmd = glassFishLauncher.getCommandLine();
                 StringBuilder sb = new StringBuilder();
+                boolean skipLine = false;
```
hs536, line 146: just an idea that does not use the skip flag :)
```java
        int indexOfReadStdin = cmd.indexOf("-read-stdin");
        List cmdForRun = IntStream.range(0, cmd.size())
             .filter(index -> index < indexOfReadStdin || index > indexOfReadStdin + 1)
             .mapToObj(cmd::get).collect(Collectors.toList());
```

OndroMih: Thanks for your suggestion. I refactored the code as you suggested.

hs536 (approved):
LGTM

PR #24472, opened 2023-06-20
In fastest mode (-Pfastest) we don't need to build the docs

arjantijms:


senivam (approved):

avpinchuk (approved):

dmatej:
This is wrong, because aggregator pom is not inherited, so now the fastest properties are not applied to individual modules.

arjantijms:
> This is wrong, because aggregator pom is not inherited, so now the fastest properties are not applied to individual modules.

Ai, silly me. Let's correct it.

PR #24475, opened 2023-06-22
OSGI vs Java 11+

dmatej:
- Updated `osgi.properties`
- Updated system packages and separated extra packages for osgi
- Using `project.osgi.version` in `osgi.properties` instead of hardcoded `3.1` and `3.1.1`
- Reverted jaxb-osgi.jar hack; based on help from @lukasj in https://github.com/eclipse-ee4j/jaxb-ri/pull/1722


nucleus/distributions/nucleus-common/src/main/resources/config/osgi.properties
```diff
@@ -220,12 +217,102 @@ felix.shutdown.hook=false
...
 
-#
-# Java SE 6 platform packages. Taken from: http://java.sun.com/javase/6/docs/api/
-# Don't add/remove any new package unless you find bugs.
-# Since we don't know the actual package versions,
-# we export most of them with default version.
-jre-1.6=\
+# List of packages available in every Java SE
+# See also: https://docs.oracle.com/en/java/javase/11/docs/api/element-list
+java.packages=\
```
lukasj, line 222: an alternative solution may be to calculate exported packages dynamically - see the patch reachable from https://bugs.eclipse.org/bugs/show_bug.cgi?id=527353 - maybe similar programatic way could be employed instead of having to maintain a properties file?

(was looking for full list of packages in frameworks and somehow reached that bug and in the end noticed obvious `ModuleLayer.boot().getModules().forEach ... getDescriptor().exports().forEach ... process package`)

dmatej: Hmmm, perhaps I would merge this first and then I will take a look what is possible. I already tried to do something similar last night, but I was too tired and did some "tragic mistake", so GF did not even start.

avpinchuk (approved):
Looks great!

hs536 (approved):

PR #24477, opened 2023-06-23
Integrate CDI TCK 4.0.10 in TCK test

arjantijms:
Added "Only normal scopes can be passivating" to detect a deployment error (exception transformers for the TCK are allowed to use such specific scanning), and seed a new exception from the transformer with `root.getMessage()` instead of `root` to prevent a stackoverflow (the code calling the transformer will use `getCause()` and then transform again, leading to an infinite loop.


dmatej (approved):

dmatej:
Note: I did not execute it yet, but it looks well!

avpinchuk (approved):

gurunrao (approved):

PR #24478, opened 2023-06-24
Autodetection of packages for OSGI bootstrap

dmatej:
- Implemented the autodetection of packages for OSGI bootstrap
- JDK packages are based on JPMS
- Bootstrap classpath packages are based on manifest files
- Based on ideas from #24475 

Interesting additional changes and notes:
- The `LegacyJmsRecoveryResourceHandler` was removed. It fixed failure after I removed one explicit exclusion of the import package `!com.sun.messaging.jmq.admin.jmsspi`, while its reason wasn't simply visible in the code. It is a support for the IMQ version used by GF5 and I doubt it still worked.
- Some resource bundles of Connectors/JMS moved to modules where they were used. It shall fix some unresolvable messages and also simplify refactoring as it reduced dependencies on resource bundles in other modules.
- If all packages in the module were in Exported-Package, the manifest was generated correctly even with defaults, so I didn't need to customized it. Except if there were `impl` or `internal`, then I had to do that. See https://bnd.bndtools.org/tools/felix-maven.html for more.
- Migrated one of web_jsp Ant tests to application-tests as it reproduced one error I made and running Ant provided too slow iterations. There should be done much more soon in this direction ...
- jta module had package protected by a password "glassfish".  I removed the protection - passwords readable in sources  or simply guessable are not protecting anything.
- jts module had optional corba imports. To my surprise no customization was required, because felix plugin respected optional dependencies declared in pom.xml. Nice and smooth.
- war-util had optional dependency on `sun.rmi.transport` package, but it doesn't use it any more, so I removed it.
- web-core contains copy pasted Catalina code, which refers to Xerces. It is a dead code for GlassFish, but osgi doesn't know that. After some refactoring we can remove the explicit exclusion.
- Some imports were based on reflections, but current Felix maven plugin can detect even them, which sometimes creates new issues.
- The --exportcontents was bit overused. See https://felix.apache.org/documentation/subprojects/apache-felix-maven-bundle-plugin-bnd.html#_embed_dependency_and_export_package

Review:
- I made it using `git diff -w -M eclipse/master` now as github doesn't provide useful UI for such cases (moving content of files to other files and removing sources).

Testing:
   - https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/job/PR-24478/
   - Build 67: https://ci.eclipse.org/jakartaee-tck/job/jakartaee-tck/job/10.0.x/


dmatej:
Status: experimental. Usual issues are caused by `osgi.bundle` files, which were written manually. Current Felix plugin does quite good job, mostly, just in rare cases needs some additional instruction (dynamic packages, probes). I will try to remove most of them, but I need to do that step by step, otherwise I would get crazy.

dmatej:
Ok, all my builds passed, time for reviews!

pzygielo (approved):

avpinchuk (approved):

senivam (approved):

arjantijms (approved):

PR #24509, opened 2023-07-17
Update Maven plugins

arjantijms:


nucleus/parent/pom.xml
```diff
@@ -155,6 +155,7 @@
         <replacer.plugin.version>1.5.3</replacer.plugin.version>
 
         <easymock.version>4.3</easymock.version>
+        <jacoco.version>0.8.10</jacoco.version>
```
avpinchuk, line 158: Property `jacoco.version` already defined at line 1545 for `jacoco` profile. Is that how it should be?

arjantijms: It was exposed outside the `jacoco` profile as well. At least the maven versions plug-in mentioned it.

avpinchuk: Then the definition in line 1545 is not redundant?

arjantijms: I have no idea to be honest. @dmatej ?

arjantijms: I've removed the property again. Let's look at this in another PR.

avpinchuk (approved):

PR #24516, opened 2023-07-21
Security logging - another step of cleanup

dmatej:
- split from #24478 
- target: remove dependencies on resource bundles across modules - it is hard to follow changes, it causes issues after every refactoring.
- removed dtrace check (not reachable any more) 
- as I was quickly reading the code, I did some trivial changes, basically using Eclipse IDE tools


avpinchuk (approved):
Looks good

PR #24530, opened 2023-08-02
Fix webapp Multi-Release JAR processing

### LINKED ISSUE #24526: Glassfish server does not support multi-release JARs in WARs (if any)
## Environment Details
* GlassFish Version (and build number): 7.0.6
* JDK version: 17
* OS: Suse Linux 12.5
* Database: Not Applicable

----------

## Problem Description
Since Java 9 there are Multi-Release JARs ([MRJARS](https://openjdk.java.net/jeps/238)) that allow different classes for different Java versions to be included in one JAR file. I was surprised when a Glassfish 7.0.6 running on Java 17 executed Java-8-code in a JAR included in a WAR although the JAR was a MRJAR with code for Java 17.

## State if the problem is easily reproducible or happens intermittently
The problem is easily reproducible when a war is created having below dependency.
<dependency>
    <groupId>org.glassfish.jersey.ext</groupId>
    <artifactId>jersey-spring6</artifactId>
    <version>3.1.3</version>
</dependency>

## Impact of Issue
Our applications are not working after migrating to Jakarta EE 10. Dependencies mapped with @Autowired annotation failed to inject.

avpinchuk:
This is a bug fix.

Fixes #24526


arjantijms (dismissed):

dmatej (approved):

dmatej (approved):

appserver/tests/application/src/test/java/org/glassfish/main/test/app/web/mrjar/MultiReleaseServletContainerInitializerTest.java
```diff
@@ -0,0 +1,135 @@
...
+    @BeforeAll
+    public static void deploy() throws IOException {
+        File warFile = createDeployment();
+        try {
+            assertThat(ASADMIN.exec("deploy", warFile.getAbsolutePath()), asadminOK());
+        } finally {
+            try {
+                Files.deleteIfExists(warFile.toPath());
+            } catch (IOException e) {
+                LOG.log(WARNING, "An error occurred while delete file {0}", warFile.getAbsolutePath());
```
pzygielo, line 70: Guessing this would not happen very often.
But when it does and we log then, perhaps anything about the `e` could be logged as well? (Similar in other tests.)

avpinchuk: @pzygielo, you're right. Fixed.

Thank you for feedback!

hs536 (approved):

PR #24533, opened 2023-08-05
Optimize CDI extensions to reduce unnecessary PAT handler calls

OndroMih:
Follows the advice from WELD in the log:
WELD-000411: Observer method [BackedAnnotatedMethod] public org.glassfish.jms.injection.JMSCDIExtension.processAnnotatedType(@Observes ProcessAnnotatedType<T>) receives events for all annotated types. Consider restricting events using @WithAnnotations or a generic type with bounds.|#]

Tested manually that the RestExtension and ServerSentEventCdiExtension processesAnnotatedType mathods are executed many times on GF 7.0.7 even if there's only a single matching bean in the app. With this PR, I observed that those methods are triggered only once if there's an Application bean in the app or a bean with the ServerSentEvent annotation (the latter is an annotation from internal GF API, not from Jakarta API).


dmatej:
Wow, I was always wondering what exactly we should do with that! Good! :-)

dmatej (approved):

avpinchuk (approved):

PR #24534, opened 2023-08-05
Ondromih profiles to speedup dev build

OndroMih:
As described in the README:

* `noextras` profile - skips building extra artifacts, e.g. GlassFish Embedded artifacts
* `noweb` profile - skips building GlassFish Server Web distribution

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,13 @@ The Zip distributions can be found on following paths:
 * `mvn clean install -Pfast` - Building all distribution artifacts, running just unit tests, QA and integration tests excluded. Typical time: 3 minutes.
 * `mvn clean install -Pfastest` - Building all distribution artifacts, excluded all QA and testing. Typical time: 1.5 minutes.
 
+You can also skip building specific distributions with the following profiles:
+
+* `noextras` profile - skips building extra artifacts, e.g. GlassFish Embedded artifacts
+* `noweb` profile - skips building GlassFish Server Web distribution
+
+For example, the following command will build only the GlassFish main distribution without QA and testing, with no Web and extra (GlassFish Embedded) artifacts: `mvn clean install -Pfastest,noweb,noextras`
+
 You can use also some maven optimizations, ie. using `-T4C` to allow parallel build.
 
 If you want to see more logs you can use the `-Dtest.logLevel=FINEST` option set to an appropriate log level.
--- a/appserver/distributions/web/pom.xml
+++ b/appserver/distributions/web/pom.xml
@@ -32,108 +32,120 @@
 
     <name>Glassfish Web Profile Distribution</name>
 
-    <properties>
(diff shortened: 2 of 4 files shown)


dmatej:
Nice, for the first time you reproduced my local issue on Jenkins :D
I will push the fix now ...

dmatej:
Btw, what if we would keep a conventional `-Pskip*` prefix? `-PskipExtras` sounds natural to me. What other people think about it? Also it should be documented in README.md. 
It is getting more and more complicated. Maybe one could just use `-pl :!extras`
Or we could add these exclusions to `-Pfastest`.

OndroMih:
Unfortunately, `-pl !:extras` doesn't work well - it only excludes the `extras` module but still builds all its child modules.

Maybe it's better to use a completely different approach with `-am` to build a specific distribution artifact and all its dependencies?

E.g., to build GlassFish full only: `mvn -T4C clean install -Pfastest -am -pl appserver/distributions/glassfish`

dmatej:
Sad, you are right with those exclusions: https://issues.apache.org/jira/browse/MNG-5230

Wow, to my surprise this works, is it what gives the result you needed?
`mvn clean install -Pfastest -pl 'org.glassfish.main.distributions:glassfish' -am -T4C`

I will try yet with an empty .m2/repository ...

hs536:
IMHO, `-Pskip*` prefix sounds natural if we skip a process, but, in this case, `-Pno*` prefix also sounds natural to me because no archive is generated.

dmatej (approved):

OndroMih:
> Wow, to my surprise this works, is it what gives the result you needed? `mvn clean install -Pfastest -pl 'org.glassfish.main.distributions:glassfish' -am -T4C`
> 

@dmatej, yes, this is exactly what I would like to achieve, but with a simple command line. Unfortunately, maven doesn't support specifying command line options in a profile or in a file. I would like to have a maven plugin that can modify the default maven command line options so that we could enable the plugin with a profile like "dist-glassfish" and run the same as the above command just with `mvn clean install -Pfastest,dist-glassfish`. Or at least put the arguments `-pl 'org.glassfish.main.distributions:glassfish' -am -T4C` into a file called `dist-glassfish.config` and then run the same just with `mavn clean install -Pfastest --args dist-glassfish.config`

None of these options are supported by maven currently. And the closest to that is these `noweb` and `noextras` profiles. Or we can just document how to use the `-pl` and `-am` arguments to build specific distributions and use that instead of these profiles.

What do you think? Is it worth adding these profiles or just document `-pl` and `-am` arguments? Since Maven 3.9.0, , the maven arguments can be specified in the MAVEN_ARGS environment variable (https://maven.apache.org/configure.html#maven_args-environment-variable), so it's maybe enough to use them instead of profiles.

PR #24536, opened 2023-08-08
Add MicroProfile Rest Client TCK

avpinchuk:
*MicroProfile Rest Client TCK.*

Run:
```bash
mvn clean verify -Pstaging,tck -pl :glassfish-external-tck-microprofile-rest-client
```

### DIFF
--- a/appserver/tests/tck/microprofile/rest-client/src/main/java/org/glassfish/microprofile/rest/client/tck/client/BeansXmlTransformer.java
+++ b/appserver/tests/tck/microprofile/rest-client/src/main/java/org/glassfish/microprofile/rest/client/tck/client/BeansXmlTransformer.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright (c) 2023 Contributors to Eclipse Foundation.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+
+package org.glassfish.microprofile.rest.client.tck.client;
+
+import java.net.URL;
+import java.util.Optional;
+import java.util.function.Consumer;
+
+import org.jboss.arquillian.container.spi.event.container.BeforeDeploy;
+import org.jboss.arquillian.container.test.spi.client.deployment.ApplicationArchiveProcessor;
+import org.jboss.arquillian.core.api.annotation.Observes;
+import org.jboss.arquillian.test.spi.TestClass;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.ArchivePath;
+import org.jboss.shrinkwrap.api.Node;
+import org.jboss.shrinkwrap.api.asset.ArchiveAsset;
+import org.jboss.shrinkwrap.api.asset.UrlAsset;
+
+import static java.lang.String.format;
+import static java.lang.System.Logger.Level.INFO;
+
+/**
+ * This extension replaces beans.xml files with ones declaring the 'all' bean discovery type.
+ * This is because version 3.0.1 of the TCK still deploys an empty beans.xml due to a faulty
+ * assumption that CDI < 4 is still defaulting to the 'all' type.
+ */
+public class BeansXmlTransformer implements ApplicationArchiveProcessor {
+
+    private static final System.Logger LOG = System.getLogger(BeansXmlTransformer.class.getName());
+
+    private static final String LIB_DIR_PATH = format("WEB-INF%slib", ArchivePath.SEPARATOR);
+
+    private static final String[] BEANS_XML_PATHS = {
+        format("/META-INF%sbeans.xml", ArchivePath.SEPARATOR),
+        format("/WEB-INF%sbeans.xml", ArchivePath.SEPARATOR),
+    };
+
+    private final URL beansXmlResource;
+
+    public BeansXmlTransformer() {
+        this.beansXmlResource = getClass().getClassLoader().getResource("beans.xml");
+        if (beansXmlResource == null) {
+            throw new IllegalStateException("Unable to find beans.xml resource in test dir");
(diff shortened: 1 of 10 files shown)


appserver/tests/tck/microprofile/rest-client/tck-suite.xml
```diff
@@ -0,0 +1,35 @@
...
+<suite name="MicroProfile Rest Client TCK" verbose="2" configfailurepolicy="continue" >
+
+    <test name="MicroProfile Rest Client TCK">
+        <packages>
+            <package name="org.eclipse.microprofile.rest.client.tck.*" />
+        </packages>
+        <classes>
+            <class name="org.eclipse.microprofile.rest.client.tck.cditests.ConfigKeyForMultipleInterfacesTest">
+                <methods>
+                    <exclude name="testConfigKeyUsedForUri" />
```
avpinchuk, line 29: I don't know yet why this test fails ;(

avpinchuk:
Comments and suggestions are strongly welcome

appserver/tests/tck/microprofile/rest-client/tck-suite.xml
dmatej: @arjantijms @hs536 @OndroMih Any ideas?

avpinchuk: So, only in this test provider setup done with fqcn_od_client/mp-rest/provider=fqcn_of_provider property in the `microprofile-config.properties` file. Config file resides in `WEB-INF/lib/test_name.jar/META-INF` directory of the test war and contains also another property. This other property reads as expected. It seems to like a provider property reads as a "regular" property, not as provider, in this paticular case, and thus provider not registered.

Nevertheless, standalone sample application with same structure, deployed to GlassFish Server, works as expected.

It seems to like that this problem is only related to the test environment.

If we place the `microprofile-config.properties` file with provider property to the `WEB-INF\classes\META-INF` directory, all works as expected.

avpinchuk: **UPDATE:** Test fails because both `ConfigKeyTest` and `ConfigKeyForMultipleInterfacesTest` deploys war with the same name `ConfigKeyTest.war`

OndroMih: I raised this https://github.com/eclipse/microprofile-rest-client/issues/363. You could raise a PR against the TCK.

As a workaround, maybe we could change the archive name in an Arquillian transformer to always add some random numbers to the deployment file?

avpinchuk: Thank you for reply, @OndroMih.

Latest GlassFish Arquillian managed server plugin contains needed ability.

appserver/tests/tck/microprofile/rest-client/src/main/java/org/glassfish/microprofile/rest/client/tck/client/BeansXmlTransformer.java
```diff
@@ -0,0 +1,114 @@
...
+ * assumption that CDI < 4 is still defaulting to the 'all' type.
+ */
+public class BeansXmlTransformer implements ApplicationArchiveProcessor {
+
+    private static final System.Logger LOG = System.getLogger(BeansXmlTransformer.class.getName());
+
+    private static final String LIB_DIR_PATH = format("WEB-INF%slib", ArchivePath.SEPARATOR);
+
+    private static final String[] BEANS_XML_PATHS = {
+        format("/META-INF%sbeans.xml", ArchivePath.SEPARATOR),
```
pzygielo, line 48: Unclear, so path can start with `/` but then its elements have to be separated with `AP.SEPARATOR`?

avpinchuk: @pzygielo, cite from the `org.jboss.shrinkwrap.api.ArchivePath` JavaDoc:

> Represents a target context within an [Archive](https://javadoc.io/static/org.jboss.shrinkwrap/shrinkwrap-api/2.0.0-beta-1/org/jboss/shrinkwrap/api/Archive.html) under which an [Node](https://javadoc.io/static/org.jboss.shrinkwrap/shrinkwrap-api/2.0.0-beta-1/org/jboss/shrinkwrap/api/Node.html) may be found. All [ArchivePath](https://javadoc.io/static/org.jboss.shrinkwrap/shrinkwrap-api/2.0.0-beta-1/org/jboss/shrinkwrap/api/ArchivePath.html) contexts are absolute (ie. prepended with the '/' character).

Thanks for comment, I fixed this in another place :)

OndroMih: Looking into Shrinkwrap code, Archive.SEPARATOR is always '/': https://github.com/shrinkwrap/shrinkwrap/blob/main/api/src/main/java/org/jboss/shrinkwrap/api/ArchivePath.java#L31.

So it doesn't matter if we use '/' directly, or via the constant. But I wouldn't mix it. I think the Javadoc writes '/' but means ArchivePath.SEPARATOR, as there's no reason to use '/' in the beginning and something else later in the path.

I suggest the following code:

```
BEANS_XML_PATHS = {
        "/META-INF/beans.xml".replace('/', ArchivePath.SEPARATOR),
```

Or just use plain '/' because I don't believe hat the separator will ever change.

dmatej: I vote for just a simple `/` character without replace - it is just a hardcoded constant, so we are replacing `/` by `/` here.

dmatej:
Tests passed locally too ;)
```
[WARNING] Tests run: 225, Failures: 0, Errors: 0, Skipped: 14, Time elapsed: 400.4 s -- in TestSuite
[INFO]
[INFO] Results:
[INFO]
[WARNING] Tests run: 225, Failures: 0, Errors: 0, Skipped: 14
....
[INFO] BUILD SUCCESS
[INFO] Total time:  06:56 min
[INFO] Finished at: 2023-08-13T16:20:38+02:00
```

appserver/tests/tck/microprofile/rest-client/src/main/java/org/glassfish/microprofile/rest/client/tck/client/BeansXmlTransformer.java
avpinchuk: Done.

Many thanks all for advice.

dmatej (approved):

pzygielo (approved):

OndroMih (approved):

PR #24539, opened 2023-08-11
Tests to verify that tasks submitted to executor during deployment are executed

OndroMih:
Adds tests for https://github.com/eclipse-ee4j/glassfish/pull/24544.

This PR was previously raised to fix https://github.com/eclipse-ee4j/glassfish/issues/21087 but then this was fixed by https://github.com/eclipse-ee4j/glassfish/pull/24544. What remains in this PR is tests to verify the expected behavior.

PR #24540, opened 2023-08-11
Tasks submitted to executor during deployment executed after deployment.

OndroMih:
An improvement of https://github.com/eclipse-ee4j/glassfish/pull/24539.

If tasks submitted during deployment, e.g. in @Initialized(ApplicationScoped.class) observer, they are added to a queue in the deployment context and triggered after deployment finished. This draft currently modifies the ManagedExecutorService to execute tasks after deployment complete. It doesn't modify ManagedScheduledExecutorService.

Description of the solution:
1. Application lifecycle listener creates storage for tasks when deployment starts
2. Executor detects that storage exists in deployment context (deployment is not complete) - it won't execute tasks immediately but will add tasks to the storage
3.  When deployment completes, application lifecycle listener executes the tasks from the storage and removes storage.
4. Executor no longer detects storage and executes the tasks immediately

To do:

* Implement also for scheduled executor service
* Add some logging
* An option to trigger tasks immediately during deployment, or trigger tasks immediately for plain executor and only delay with scheduled executor (the logic is that tasks are scheduled after deployment) - there are multiple options, see below
* tests (task submitted during deployment is triggered after startup method completes, task submitted after deployment triggers immediately)

There are multiple options to implement this whole feature:

* execute all tasks after deployment
* execute only tasks submitted by scheduled executor after deployment but tasks submitted by plain executor immediately (the logic is tasks should be scheduled to execute after deployment and thus should be submitted by a scheduled executor)
* support an option on the executor or deployment descriptor to switch the behavior - either execute immediately or execute after deployment completes.

### DIFF
--- a/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/AppLifecycleListener.java
+++ b/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/AppLifecycleListener.java
@@ -1,0 +1,59 @@
+/*
+ * Click nbfs://nbhost/SystemFileSystem/Templates/Licenses/license-default.txt to change this license
+ * Click nbfs://nbhost/SystemFileSystem/Templates/Classes/Class.java to edit this template
+ */
+package org.glassfish.concurrent.runtime;
+
+import java.util.ArrayList;
+import java.util.List;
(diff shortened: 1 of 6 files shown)


OndroMih:
This needs some more work and I don't have time for this now so I'll close this PR. 

If somebody wants to pick this up, feel free to start from my branch and finish the tasks in the to do list in the description. I'll try to get to this later when I have more time.

PR #24544, opened 2023-08-14
Propagate JNDI context if none exists. Fixes that tasks submitted to executor during deployment are executed

### LINKED ISSUE #21087: ManagedExecutorService does not execute tasks submitted during application startup
I'm attempting to schedule a task from inside a singleton bean which is initialised at startup:

```
@Singleton
@Startup
public class StartupSingleton {
      @Resource
      ManagedExecutorService executorService;

      @PostConstruct
      public void init() {
            System.out.println("init");
            Future<?> future = executorService.submit(new Runnable() {
  public void run() {
        System.out.println("run");
  }
            });

            try {
  Thread.sleep(1000);
  if(future.isDone()) {
        future.get();
        System.out.println("ok");
  } else {
        System.out.println("not run");
  }
            } catch (Exception e) {
  e.printStackTrace();
            }
      }
}
```

I then get the following in the logs:

```
2014-06-10T15:45:07.497+0930|Info: init
2014-06-10T15:45:08.499+0930|Severe: java.util.concurrent.ExecutionException: javax.enterprise.concurrent.AbortedException: Module glassfish-executor-test is disabled
      at java.util.concurrent.FutureTask.report(FutureTask.java:122)
      at java.util.concurrent.FutureTask.get(FutureTask.java:188)
      at foo.StartupSingleton.init(StartupSingleton.java:33)
      at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
      at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
      at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
      at java.lang.reflect.Method.invoke(Method.java:606)
      at com.sun.ejb.containers.interceptors.BeanCallbackInterceptor.intercept(InterceptorManager.java:1035)
      at com.sun.ejb.containers.interceptors.CallbackChainImpl.invokeNext(CallbackChainImpl.java:72)
      at com.sun.ejb.containers.interceptors.CallbackInvocationContext.proceed(CallbackInvocationContext.java:205)
      at org.jboss.weld.ejb.AbstractEJBRequestScopeActivationInterceptor.aroundInvoke(AbstractEJBRequestScopeActivationInterceptor.java:55)
      at sun.reflect.GeneratedMethodAccessor129.invoke(Unknown Source)
      at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
      at java.lang.reflect.Method.invoke(Method.java:606)
      at com.sun.ejb.containers.interceptors.CallbackInterceptor.intercept(InterceptorManager.java:986)
      at com.sun.ejb.containers.interceptors.CallbackChainImpl.invokeNext(CallbackChainImpl.java:72)
      at com.sun.ejb.containers.interceptors.CallbackInvocationContext.proceed(CallbackInvocationContext.java:205)
      at com.sun.ejb.containers.interceptors.SystemInterceptorProxy.doCall(SystemInterceptorProxy.java:163)
      at com.sun.ejb.containers.interceptors.SystemInterceptorProxy.init(SystemInterceptorProxy.java:125)
      at sun.reflect.GeneratedMethodAccessor130.invoke(Unknown Source)
      at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
      at java.lang.reflect.Method.invoke(Method.java:606)
      at com.sun.ejb.containers.interceptors.CallbackInterceptor.intercept(InterceptorManager.java:986)
      at com.sun.ejb.containers.interceptors.CallbackChainImpl.invokeNext(CallbackChainImpl.java:72)
      at com.sun.ejb.containers.interceptors.InterceptorManager.intercept(InterceptorManager.java:412)
      at com.sun.ejb.containers.interceptors.InterceptorManager.intercept(InterceptorManager.java:375)
      at com.sun.ejb.containers.BaseContainer.intercept(BaseContainer.java:1949)
      at com.sun.ejb.containers.AbstractSingletonContainer.createSingletonEJB(AbstractSingletonContainer.java:475)
      at com.sun.ejb.containers.AbstractSingletonContainer.access$000(AbstractSingletonContainer.java:81)
      at com.sun.ejb.containers.AbstractSingletonContainer$SingletonContextFactory.create(AbstractSingletonContainer.java:654)
      at com.sun.ejb.containers.AbstractSingletonContainer.instantiateSingletonInstance(AbstractSingletonContainer.java:396)
      at org.glassfish.ejb.startup.SingletonLifeCycleManager.initializeSingleton(SingletonLifeCycleManager.java:219)
      at org.glassfish.ejb.startup.SingletonLifeCycleManager.initializeSingleton(SingletonLifeCycleManager.java:180)
      at org.glassfish.ejb.startup.SingletonLifeCycleManager.doStartup(SingletonLifeCycleManager.java:158)
      at org.glassfish.ejb.startup.EjbApplication.start(EjbApplication.java:166)
      at org.glassfish.internal.data.EngineRef.start(EngineRef.java:122)
      at org.glassfish.internal.data.ModuleInfo.start(ModuleInfo.java:291)
      at org.glassfish.internal.data.ApplicationInfo.start(ApplicationInfo.java:352)
      at com.sun.enterprise.v3.server.ApplicationLifecycle.deploy(ApplicationLifecycle.java:497)
      at com.sun.enterprise.v3.server.ApplicationLifecycle.deploy(ApplicationLifecycle.java:219)
      at org.glassfish.deployment.admin.DeployCommand.execute(DeployCommand.java:491)
      at com.sun.enterprise.v3.admin.CommandRunnerImpl$2$1.run(CommandRunnerImpl.java:527)
      at com.sun.enterprise.v3.admin.CommandRunnerImpl$2$1.run(CommandRunnerImpl.java:523)
      at java.security.AccessController.doPrivileged(Native Method)
      at javax.security.auth.Subject.doAs(Subject.java:356)
      at com.sun.enterprise.v3.admin.CommandRunnerImpl$2.execute(CommandRunnerImpl.java:522)
      at com.sun.enterprise.v3.admin.CommandRunnerImpl.doCommand(CommandRunnerImpl.java:546)
      at com.sun.enterprise.v3.admin.CommandRunnerImpl.doCommand(CommandRunnerImpl.java:1423)
      at com.sun.enterprise.v3.admin.CommandRunnerImpl.access$1500(CommandRunnerImpl.java:108)
      at com.sun.enterprise.v3.admin.CommandRunnerImpl$ExecutionContext.execute(CommandRunnerImpl.java:1762)
      at com.sun.enterprise.v3.admin.CommandRunnerImpl$ExecutionContext.execute(CommandRunnerImpl.java:1674)
      at com.sun.enterprise.v3.admin.AdminAdapter.doCommand(AdminAdapter.java:534)
      at com.sun.enterprise.v3.admin.AdminAdapter.onMissingResource(AdminAdapter.java:224)
      at org.glassfish.grizzly.http.server.StaticHttpHandler.service(StaticHttpHandler.java:297)
      at com.sun.enterprise.v3.services.impl.ContainerMapper.service(ContainerMapper.java:246)
      at org.glassfish.grizzly.http.server.HttpHandler.runService(HttpHandler.java:191)
      at org.glassfish.grizzly.http.server.HttpHandler.doHandle(HttpHandler.java:168)
      at org.glassfish.grizzly.http.server.HttpServerFilter.handleRead(HttpServerFilter.java:189)
      at org.glassfish.grizzly.filterchain.ExecutorResolver$9.execute(ExecutorResolver.java:119)
      at org.glassfish.grizzly.filterchain.DefaultFilterChain.executeFilter(DefaultFilterChain.java:288)
      at org.glassfish.grizzly.filterchain.DefaultFilterChain.executeChainPart(DefaultFilterChain.java:206)
      at org.glassfish.grizzly.filterchain.DefaultFilterChain.execute(DefaultFilterChain.java:136)
      at org.glassfish.grizzly.filterchain.DefaultFilterChain.process(DefaultFilterChain.java:114)
      at org.glassfish.grizzly.ProcessorExecutor.execute(ProcessorExecutor.java:77)
      at org.glassfish.grizzly.nio.transport.TCPNIOTransport.fireIOEvent(TCPNIOTransport.java:838)
      at org.glassfish.grizzly.strategies.AbstractIOStrategy.fireIOEvent(AbstractIOStrategy.java:113)
      at org.glassfish.grizzly.strategies.WorkerThreadIOStrategy.run0(WorkerThreadIOStrategy.java:115)
      at org.glassfish.grizzly.strategies.WorkerThreadIOStrategy.access$100(WorkerThreadIOStrategy.java:55)
      at org.glassfish.grizzly.strategies.WorkerThreadIOStrategy$WorkerThreadRunnable.run(WorkerThreadIOStrategy.java:135)
      at org.glassfish.grizzly.threadpool.AbstractThreadPool$Worker.doWork(AbstractThreadPool.java:564)
      at org.glassfish.grizzly.threadpool.AbstractThreadPool$Worker.run(AbstractThreadPool.java:544)
      at java.lang.Thread.run(Thread.java:744)
Caused by: javax.enterprise.concurrent.AbortedException: Module glassfish-executor-test is disabled
      at org.glassfish.enterprise.concurrent.internal.ManagedFutureTask.run(ManagedFutureTask.java:146)
      at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
      at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
      at java.lang.Thread.run(Thread.java:744)
```

And the submitted task is not run. (glassfish-executor-test is just the name of the standalone test case that I created).

Surely either the executor service should be initialized and available by the time <tt>@PostConstruct</tt> stuff is executed, or it should queue submitted tasks and execute them once the application is started.

This is just using the default managed executor service in a default <tt>domain.xml</tt> file.
#### Environment
Glassfish 4.0, Java 7, currently testing on Windows 7
#### Affected Versions
[4.0]

OndroMih:
This addresses the conversation in https://github.com/eclipse-ee4j/glassfish/pull/24539#discussion_r1292819699 and fixes https://github.com/eclipse-ee4j/glassfish/issues/21087.

Changes the logic in ContextSetupProviderImpl to always store the invocation context for another thread, and then restore it only if needed. The application name is retrieved from the stored context, so it's always available and should never be `null`. 

I also fixed the problem that appName was sometimes null by always copying the appName from the previous context to the new one. 

Passes Concurrency TCK in appserver/tests/tck/concurrency. Passes tests in appserver/tests/admin and application. Although 2 tests failed when I ran locally, I rerun the tests individually and they passed. One in Concurrency TCK failed because GlassFish failed to start, one in Admin failed because it parsed the log file incorrectly (org.glassfish.main.admin.test.progress.DetachAttachITest) and expected a number where it was text.- Probably because of my slower machine there was some difference in the logs and possibly some extra log lines. I saw something similar in `ProgressStatusComplexITest.java`, which I fixed in this PR, but in `DetachAttachITest.java` it wasn't clear how to fix and running the test separately passed the test.

----

I also created a draft of a more complex improvement to execute the tasks after deployment is complete: https://github.com/eclipse-ee4j/glassfish/pull/24540. There are multiple options

* execute all tasks after deployment
* execute only tasks submitted by scheduled executor after deployment but tasks submitted by plain executor immediately (the logic is tasks should be scheduled to execute after deployment and thus should be submitted by a scheduled executor)
* support an option on the executor or deployment descriptor to switch the behavior - either execute immediately or execute after deployment completes.

### DIFF
--- a/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
+++ b/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
@@ -131,9 +131,7 @@ public class ContextSetupProviderImpl implements ContextSetupProvider {
         ComponentInvocation invocation = invocationCtx.getInvocation();
         final String appName = invocation == null ? null : invocation.getAppName();
 
-        if (!isApplicationEnabled(appName)) {
-            throw new IllegalStateException("Module " + appName + " is disabled");
-        }
+        verifyApplicationEnabled(appName);
 
         final ClassLoader resetClassLoader;
         if (invocationCtx.getContextClassLoader() != null) {
@@ -153,6 +151,9 @@ public class ContextSetupProviderImpl implements ContextSetupProvider {
         if (setup.isUnchanged(JNDI) || (!setup.isPropagated(JNDI) && !setup.isClear(JNDI))) {
             ComponentInvocation currentInvocation = invocationManager.getCurrentInvocation();
             if (currentInvocation != null) {
+                if (currentInvocation.getAppName() == null) {
+                    currentInvocation.setAppName(invocation.getAppName());
+                }
                 invocation = currentInvocation;
             }
         }
@@ -242,7 +243,7 @@ public class ContextSetupProviderImpl implements ContextSetupProvider {
             return null;
         }
         if (setup.isClear(JNDI)) {
-            return new ComponentInvocation();
+            return new ComponentInvocation(null, null, null, currentInvocation.getAppName(), null);
         }
         return cloneComponentInvocation(currentInvocation);
     }
@@ -252,16 +253,19 @@ public class ContextSetupProviderImpl implements ContextSetupProvider {
      * Check whether the application component submitting the task is still running.
      * Throw IllegalStateException if not.
      */
-    private boolean isApplicationEnabled(String appId) {
+    private void verifyApplicationEnabled(String appId) {
         if (appId == null) {
-            // we don't know the application name yet, it is starting.
-            return true;
+            IllegalStateException ex = new IllegalStateException("The appId is null");
+            LOG.log(Level.WARNING, "Context handle not in valid state. Do not run the task. " + ex.getMessage(), ex);
+            throw ex;
         }
         Application app = applications.getApplication(appId);
-        if (app != null) {
-            return deployment.isAppEnabled(app);
+        if (app != null && !deployment.isAppEnabled(app)) {
+            IllegalStateException ex = new IllegalStateException("Module " + appId + " is disabled");
+            LOG.log(Level.WARNING, "Context handle not in valid state. Do not run the task. " + ex.getMessage(), ex);
+            throw ex;
         }
-        return false;
+        // if we know the application name but don't have an application yet, it is starting, and thus enabled
     }
 
 
--- a/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
+++ b/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
@@ -149,11 +149,18 @@ public class ContextSetupProviderImpl implements ContextSetupProvider {
             resetSecurityContext = SecurityContext.getCurrent();
             SecurityContext.setCurrent(invocationCtx.getSecurityContext());
         }
-        if (invocation != null) {
-            // Each invocation needs a ResourceTableKey that returns a unique hashCode for TransactionManager
-            invocation.setResourceTableKey(new PairKey(invocation.getInstance(), Thread.currentThread()));
-            invocationManager.preInvoke(invocation);
+
+        if (setup.isUnchanged(JNDI) || (!setup.isPropagated(JNDI) && !setup.isClear(JNDI))) {
+            ComponentInvocation currentInvocation = invocationManager.getCurrentInvocation();
+            if (currentInvocation != null) {
+                invocation = currentInvocation;
+            }
         }
+
+        // Each invocation needs a ResourceTableKey that returns a unique hashCode for TransactionManager
+        invocation.setResourceTableKey(new PairKey(invocation.getInstance(), Thread.currentThread()));
+        invocationManager.preInvoke(invocation);
+
         // Ensure that there is no existing transaction in the current thread
         if (transactionManager != null && setup.isClear(StandardContextType.WorkArea)) {
             transactionManager.clearThreadTx();
@@ -237,10 +244,7 @@ public class ContextSetupProviderImpl implements ContextSetupProvider {
         if (setup.isClear(JNDI)) {
             return new ComponentInvocation();
         }
-        if (setup.isPropagated(JNDI)) {
-            return cloneComponentInvocation(currentInvocation);
-        }
-        return null;
+        return cloneComponentInvocation(currentInvocation);
     }
 
 
--- a/appserver/tests/admin/tests/src/test/java/org/glassfish/main/admin/test/progress/ProgressStatusComplexITest.java
+++ b/appserver/tests/admin/tests/src/test/java/org/glassfish/main/admin/test/progress/ProgressStatusComplexITest.java
@@ -67,9 +67,9 @@ public class ProgressStatusComplexITest {
         assertArrayEquals(new String[] {
             "Starting", "2 seconds supplemental command", "Parsing",
             "Working on main part", "Finished", "3 seconds supplemental" },
-            uniqueMessages(prgs));
+                uniqueMessages(prgs));
 
-        assertThat(prgs, hasSize(21));
+        assertThat(prgs, hasSize(greaterThanOrEqualTo(21)));
         assertThat(prgs, isIncreasing());
         assertFalse(prgs.get(4).isPercentage());
         assertTrue(prgs.get(10).isPercentage());
--- a/appserver/tests/admin/tests/src/test/java/org/glassfish/main/admin/test/progress/ProgressStatusComplexITest.java
+++ b/appserver/tests/admin/tests/src/test/java/org/glassfish/main/admin/test/progress/ProgressStatusComplexITest.java
@@ -67,9 +67,9 @@ public class ProgressStatusComplexITest {
         assertArrayEquals(new String[] {
             "Starting", "2 seconds supplemental command", "Parsing",
             "Working on main part", "Finished", "3 seconds supplemental" },
-                uniqueMessages(prgs));
+            uniqueMessages(prgs));
 
-        assertThat(prgs, hasSize(greaterThanOrEqualTo(21)));
+        assertThat(prgs, hasSize(21));
         assertThat(prgs, isIncreasing());
         assertFalse(prgs.get(4).isPercentage());
         assertTrue(prgs.get(10).isPercentage());
--- a/appserver/tests/application/src/main/java/org/glassfish/main/test/app/concurrency/executor/ManagedExecutorDefinitionServlet.java
+++ b/appserver/tests/application/src/main/java/org/glassfish/main/test/app/concurrency/executor/ManagedExecutorDefinitionServlet.java
@@ -94,7 +94,7 @@ public class ManagedExecutorDefinitionServlet extends HttpServlet {
                     "Completation stage that is supplied to copy must not be modified by the ManagedExecutorService.");
             }
 
-            final String result = stage3.get(1000, TimeUnit.SECONDS);
+            final String result = stage3.get(10, TimeUnit.SECONDS);
             if ("StringContext propagated; IntContext unchanged".equals(result)) {
                 return;
             }
--- a/appserver/tests/application/src/main/java/org/glassfish/main/test/app/concurrency/executor/ManagedExecutorDefinitionServlet.java
+++ b/appserver/tests/application/src/main/java/org/glassfish/main/test/app/concurrency/executor/ManagedExecutorDefinitionServlet.java
@@ -94,7 +94,7 @@ public class ManagedExecutorDefinitionServlet extends HttpServlet {
                     "Completation stage that is supplied to copy must not be modified by the ManagedExecutorService.");
             }
 
-            final String result = stage3.get(10, TimeUnit.SECONDS);
+            final String result = stage3.get(1000, TimeUnit.SECONDS);
             if ("StringContext propagated; IntContext unchanged".equals(result)) {
                 return;
             }
--- a/appserver/tests/tck/concurrency/pom.xml
+++ b/appserver/tests/tck/concurrency/pom.xml
@@ -53,7 +53,7 @@
             <dependency>
                 <groupId>org.jboss.arquillian</groupId>
                 <artifactId>arquillian-bom</artifactId>
-                <version>1.7.0.Alpha13</version>
+                <version>1.7.0.Final</version>
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
--- a/nucleus/common/glassfish-api/src/main/java/org/glassfish/api/invocation/ComponentInvocation.java
+++ b/nucleus/common/glassfish-api/src/main/java/org/glassfish/api/invocation/ComponentInvocation.java
@@ -246,6 +246,10 @@ public class ComponentInvocation implements Cloneable {
         return appName;
     }
 
+    public void setAppName(String appName) {
+        this.appName = appName;
+    }
+
     /**
      * Returns the moduleName for the current invocation, equivalent to the value bound to java:module/ModuleName, without
      * the cost of lookup. For invocations that are not on Jakarta EE components, returns null.


appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
```diff
@@ -149,11 +147,21 @@ public ContextHandle setup(ContextHandle contextHandle) {
...
-            invocation.setResourceTableKey(new PairKey(invocation.getInstance(), Thread.currentThread()));
-            invocationManager.preInvoke(invocation);
+
+        if (setup.isUnchanged(JNDI) || (!setup.isPropagated(JNDI) && !setup.isClear(JNDI))) {
+            ComponentInvocation currentInvocation = invocationManager.getCurrentInvocation();
+            if (currentInvocation != null) {
+                if (currentInvocation.getAppName() == null) {
+                    currentInvocation.setAppName(invocation.getAppName());
+                }
+                invocation = currentInvocation;
```
avpinchuk, line 157: It's safe to replace invocation with current invocation here? Or we should more carefully copy only some state from current invocation?

OndroMih: The idea is that if application context is not cleared or propagated, it shouldn’t be changed if it’s already set. That’s according to the concurrency spec. That’s why I replace the new invocation with an existing one, if it’s present.

OndroMih:
I see that some EJB tests failed. I’ll investigate why.

appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
avpinchuk: Thank you for clarification.

avpinchuk:
@OndroMih, build successful.

dmatej (commented):
It looks like black magic resolving old black magic, but if it works, I can say just WOW! :-)

appserver/tests/application/src/main/java/org/glassfish/main/test/app/concurrency/executor/ManagedExecutorDefinitionServlet.java
```diff
@@ -94,7 +94,7 @@ private void copyCompletableFuture(int iContextValue) {
                     "Completation stage that is supplied to copy must not be modified by the ManagedExecutorService.");
             }
 
-            final String result = stage3.get(10, TimeUnit.SECONDS);
+            final String result = stage3.get(1000, TimeUnit.SECONDS);
```
dmatej, line 97: Why one thousand seconds? That is too much ...

appserver/tests/admin/tests/src/test/java/org/glassfish/main/admin/test/progress/ProgressStatusComplexITest.java
```diff
@@ -67,9 +67,9 @@ public void executeCommandWithSupplements() {
         assertArrayEquals(new String[] {
             "Starting", "2 seconds supplemental command", "Parsing",
             "Working on main part", "Finished", "3 seconds supplemental" },
-            uniqueMessages(prgs));
+                uniqueMessages(prgs));
 
-        assertThat(prgs, hasSize(21));
+        assertThat(prgs, hasSize(greaterThanOrEqualTo(21)));
```
dmatej, line 72: Maybe use the uniqueMessages for the check exact number, not sure, I am just reviewing using github. I am not sure what exactly uniqueMessages does.

appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
```diff
@@ -235,29 +243,29 @@ private ComponentInvocation getSavedInvocation() {
...
+        return cloneComponentInvocation(currentInvocation);
     }
 
 
     /**
      * Check whether the application component submitting the task is still running.
      * Throw IllegalStateException if not.
      */
-    private boolean isApplicationEnabled(String appId) {
+    private void verifyApplicationEnabled(String appId) {
```
dmatej, line 256: I still think that this method brings and always brought some fragility. It is completely heuristic, while we should rather use some lifecycle stages of the application to check where we are and what we can do .... sounds like sci-fi now :-)
I think there are also other places feeling this pain in GF.

avpinchuk: And second question: `appId == null` are illegal always or found some legal cases where it may be `null`?

appserver/tests/application/src/main/java/org/glassfish/main/test/app/concurrency/executor/ManagedExecutorDefinitionServlet.java
OndroMih: I’ll revert it. I needed this for debugging and committed it by accident.

appserver/tests/admin/tests/src/test/java/org/glassfish/main/admin/test/progress/ProgressStatusComplexITest.java
OndroMih: Exact number isn’t very useful. The log contains lines like:

20% done
35% done
80% done
….

When the system is slower, it will just produce more lines, like:

15% done
28% done
40% done
65% done
…

So the behaviour is not deterministic. It would be better to check the result in a different way but I didn’t want to bother.

appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
avpinchuk: @OndroMih, we already it the #24544 :)

appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
```diff
@@ -131,9 +131,7 @@ public ContextHandle setup(ContextHandle contextHandle) {
         ComponentInvocation invocation = invocationCtx.getInvocation();
         final String appName = invocation == null ? null : invocation.getAppName();
 
-        if (!isApplicationEnabled(appName)) {
-            throw new IllegalStateException("Module " + appName + " is disabled");
-        }
+        verifyApplicationEnabled(appName);
```
avpinchuk, line 134: Hmm, you check for null here, on original invocation, before setting up current invocation. Or it setting up somewhere else?

appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
OndroMih: Sorry, I’m on phone, I meant this PR: https://github.com/eclipse-ee4j/glassfish/pull/24540

appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
OndroMih: Yes. My assumption is that the original invocation is always set and contains the appName. The current invocation never contains the app name (current invocation is actually null all the time, unless the app modifies it with some GlassFish API if that’s possible). This method sets the appName of the current invocation if it’s null.

appserver/tests/admin/tests/src/test/java/org/glassfish/main/admin/test/progress/ProgressStatusComplexITest.java
dmatej: I'm not sure how it is computed, I have to take a look. I would expect it is deterministic, but it can skip some lines. But then it is suspicious that it changes behavior with this PR and it didn't fail before.

OndroMih: It’s not suspicious. It just doesn’t pass on my computer, when I run it together with other tests. It passes on my computer when I run the test alone. I can revert it and see if it still passes on Jenkins.

But the assertion is clearly non-deterministic.

dmatej: Integration tests should not run in parallel with other tests. The problem with these tests is that they strictly follow some scenario, once you allow them to be more tolerant, they tolerate real bugs. I hope you didn't find some :-)
Btw these "asadmin commands with progress and extensions" are unused feature, but maybe they could be interesting in the future, ie. if GF would want to support something like "ear/war installers" in cloud ... but I am not sure if it would be even useful today ... just intuition.

OndroMih: I didn't run anything in parallel. But it seems that GlassFish or my computer slows down when I execute all the test sequentially. When I executed only the failed test, it passed.

Anyway, I reverted my changes in ProgressStatusComplexITest now. Myabe it was failing only on my computer and it will pass on Jenkins. In the future, we should make the assertions more robust, I'm pretty sure that the test isn't deterministic and might fail on slower systems.

dmatej (changes requested):
Yet those 1000 seconds.

appserver/tests/admin/tests/src/test/java/org/glassfish/main/admin/test/progress/ProgressStatusComplexITest.java
dmatej: Hmm, there might be some issue, it should not slow down or cause these issues. I will take a look, we have yet one suspicious test, where I still did not find the cause, but it probably depends on order of tests.

We should make GlassFish robust, not make tests tolerant to issues, we improved it a lot, but we did no finish yet. Those percents are not random numbers. However it is a test with threads, so it is not simple to debug it.

appserver/tests/application/src/main/java/org/glassfish/main/test/app/concurrency/executor/ManagedExecutorDefinitionServlet.java
OndroMih: Reverted with the last commit.

OndroMih:
> Yet those 1000 seconds.

Reverted now.

dmatej (approved):

avpinchuk (approved):

arjantijms (approved):

PR #24549, opened 2023-08-19
Build GlassFish Embedded artifacts as Multi-Release JARs

### LINKED ISSUE #24541: Embedded glassfish cannot do context lookup() because of java.lang.NoSuchMethodError
If I deploy a very simple web app to an embedded glassfish based on glassfish-embedded-all-7.0.6.jar, and that app does a context lookup() to lookup, say, a remote EJB, the lookup will fail with a java.lang.NoSuchMethodError referencing method java.lang.Class sun.misc.Unsafe.defineClass(java.lang.String, byte[], int, int, java.lang.ClassLoader, java.security.ProtectionDomain).

## Environment Details
glassfish 7.0.6
java version "17.0.7" 2023-04-18 LTS
Java(TM) SE Runtime Environment (build 17.0.7+8-LTS-224)
Java HotSpot(TM) 64-Bit Server VM (build 17.0.7+8-LTS-224, mixed mode, sharing)
Mac OS 12.4 (Monterey)
Database not applicable

----------

## Problem Description

If I deploy a very simple web app to an embedded glassfish based on glassfish-embedded-all-7.0.6.jar, and that app does a context lookup() to lookup, say, a remote EJB, the lookup will fail with a java.lang.NoSuchMethodError referencing method java.lang.Class sun.misc.Unsafe.defineClass(java.lang.String, byte[], int, int, java.lang.ClassLoader, java.security.ProtectionDomain).

The lookup() call should return a proxy for the remote EJB.

The problem is reliably reproducible.

The problem is caused by org.glassfish.pfl.basic in glassfish-embedded-all-7.0.6.jar being out of date. Unsafe.defineClass was removed at JDK 9 so the org.glassfish.pfl.basic in the embedded jar should no longer be referring to it. This is proven by downloading pfl-basic-4.1.2.jar and adding it ahead of glassfish-embedded-all-7.0.6.jar in the reproducer class path. (The lookup still fails (and that is the subject of a separate issue) but the lookup() attempt gets a whole lot further.)

## Steps to reproduce

I have reproduced this problem by creating a web app as described and deploying it's war programmatically to a newly started embedded glassfish instance in the reproducer. I will if necessary provide all the artefacts which include:

- a modified domain.xml file (modified to prevent port conflicts with standard glassfish, an instance of which has to be running to populate java:global)
- the source of the web app, which does a context lookup()
- the reproducer itself which starts a remote glassfish and deploys the web app to it
- a class library that provides a remote EJB to be added to the glassfish server (to be looked up) and to the reproducer (to lookup)
- etc

but its a lot of work for both of us given that it should be clear, now pointed out, that org.glassfish.pfl.basic is out of date in the glassfish-embedded-all-7.0.6 JAR I imagine a quick check of what version of pfl-basic was used in the construction of the embedded jar will confirm that it is out of date.

However, previous experience has suggested a short cut. A context lookup() done from a POJO Java app with the embedded glassfish JAR in its class path (instead of gf-client.jar) does a lookup based on the classes in the embedded JAR and reproduces the problem. This is a whole lot easier to use as a reproducer so I have included that.

So, to reproduce (the first three steps are optional - the problem reproduces without them):
1. Start a convenient glassfish instance that contains an EJB
2. Get the portable JNDI name for an EJB from the server log
3. Modify the reproducer source to make this the value of the lookup string.
4. Compile and run the reproducer with embedded-all-7.0.6 in the class path, and, optionally, a class library that contains the class of the EJB to be looked up (it will never be used, hence the optionality); you will need the --add JVM options specified in the source of the reproducer for it to run
5. The problem reproduces
6. Download pfl-basic-4.1.2.jar 
7. Recompile and rerun the reproducer with the new library ahead of embedded-all-7.0.6 in the class path, and, optionally, the above class library (it will never be used, hence the optionality)
8. The problem transforms into a name space not found error; (this shouldn't be and is the subject of a separate issue)

## Impact of Issue
This error doesn't affect me now because I have worked around it by including pfl-basic-4.1.2.jar in my class path. The error that then results is a different matter but that is dealt with elsewhere.

[StackTrace.txt](https://github.com/eclipse-ee4j/glassfish/files/12328897/StackTrace.txt)

[reproducer.txt](https://github.com/eclipse-ee4j/glassfish/files/12328925/reproducer.txt)

avpinchuk:
This closes #24541.

### DIFF
--- a/appserver/extras/embedded/all/pom.xml
+++ b/appserver/extras/embedded/all/pom.xml
@@ -1775,6 +1775,7 @@
                             <outputDirectory>${project.build.directory}</outputDirectory>
                             <archive>
                                 <manifestEntries>
+                                    <Multi-Release>true</Multi-Release>
                                     <Bundle-SymbolicName>org.glassfish.main.embedded.all</Bundle-SymbolicName>
                                     <Main-Class>com.sun.enterprise.glassfish.bootstrap.UberMain</Main-Class>
                                 </manifestEntries>
--- a/appserver/extras/embedded/nucleus/pom.xml
+++ b/appserver/extras/embedded/nucleus/pom.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
+    Copyright (c) 2021, 2023 Contributors to Eclipse Foundation.
     Copyright (c) 1997, 2021 Oracle and/or its affiliates. All rights reserved.
 
     This program and the accompanying materials are made available under the
@@ -30,6 +31,7 @@
     <artifactId>glassfish-embedded-nucleus</artifactId>
 
     <name>Embedded GlassFish nucleus</name>
+    <packaging>jar</packaging>
 
     <dependencies>
         <!-- glassfish-hk2 -->
@@ -365,6 +367,7 @@
                             <outputDirectory>${project.build.directory}</outputDirectory>
                             <archive>
                                 <manifestEntries>
+                                    <Multi-Release>true</Multi-Release>
                                     <Bundle-SymbolicName>org.glassfish.main.embedded.nucleus</Bundle-SymbolicName>
                                     <Main-Class>com.sun.enterprise.glassfish.bootstrap.UberMain</Main-Class>
                                 </manifestEntries>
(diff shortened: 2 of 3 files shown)


dmatej (approved):
Hmmm, I really hate multirelease jars, but ... aaah, I understand now why we have to do this! We are repacking dependencies -> then we have to respect their structure. You are right, thank you for the fix!

PR #24551, opened 2023-08-20
Archive `glassfish-embedded-all.jar` along bundles

pzygielo:


pzygielo:
- https://github.com/eclipse-ee4j/glassfish/issues/24541#issuecomment-1685181588

pzygielo:
- IMO #24552 is better

PR #24554, opened 2023-08-21
Fix processing of the Multi-Release libraries in EAR application

avpinchuk:
Fixes processing of the MR-JARs places at the `lib` directory of the EAR application.


avpinchuk:
Should we also supports Multi-Release EJB-JARs modules of EAR application? Or EJB-JARs cannot be Multi-Release?

arjantijms:
> Should we also supports Multi-Release EJB-JARs modules of EAR application? Or EJB-JARs cannot be Multi-Release?

I think the spec is silent about that.

arjantijms (approved):

PR #24561, opened 2023-08-24
Don't log a stacktrace if MANIFEST file not found in deployed JAR

OndroMih:
Avoids polluting log by an unnecessary stacktrace when GlassFish Embedded  used in an Arquillian connector.

Only with Java 11 or Java less than 17.

To reproduce the log message, run a test using Arquillian GlassFish Embedded connector. This can be reproduced with this PR: https://github.com/eclipse-ee4j/cargotracker/pull/273 - with Java 11, run `mvn clean package -Pglassfish`

### DIFF
--- a/appserver/web/weld-integration/src/main/java/org/glassfish/weld/FilteringClassLoader.java
+++ b/appserver/web/weld-integration/src/main/java/org/glassfish/weld/FilteringClassLoader.java
@@ -35,6 +35,8 @@ import java.util.stream.Collectors;
 import static java.util.Collections.enumeration;
 import static java.util.stream.StreamSupport.stream;
 
+import java.io.FileNotFoundException;
+
 /**
  * This classloader filters out extensions which are incompatible with current environment.
  *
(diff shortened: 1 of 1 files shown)


appserver/web/weld-integration/src/main/java/org/glassfish/weld/FilteringClassLoader.java
```diff
@@ -124,7 +124,11 @@ private Manifest loadManifest(final URL manifestURL) {
         try (InputStream stream = manifestURL.openStream()) {
             return new Manifest(stream);
         } catch (final IOException e) {
-            LOG.log(Level.WARNING, "Could not read manifest at " + manifestURL, e);
+            if (LOG.isLoggable(Level.FINE)) {
+                LOG.log(Level.WARNING, "Could not read manifest at " + manifestURL, e);
+            } else {
+                LOG.log(Level.WARNING, "Could not read manifest. " + e.getMessage());
+            }
```
avpinchuk, line 131: Maybe would be better, e.g. something like

```java
LOG.log(Level.WARNING, "Could not read manifest at " + manifestURL);
if (LOG.isLoggable(Level.FINE) {
    LOG.log(Level.WARNING, "An error occurred while reading manifest. ", e);
}
```

avpinchuk: Or even

```java
LOG.log(Level.WARNING, "Could not read manifest at " + manifestURL);
LOG.log(Level.FINE, () -> "An error occurred while reading manifest.", e);

dmatej: 1. It should log the warning OR better error if it cannot load the manifest.
2. If we want to accept the situation that the manifestURL doesn't link to any existing file, it has to be pulled out from the try statement and probably extracted to another method returning null if not found. Then try-with would log an error if the URL existed, but the Manifest was invalid (both can be source of IOE).
3. Those if branches are a logging antipattern. Don't make log messages dependent on the level. It is hard to find something - you just increase verbosity and the original message is gone.

dmatej (changes requested):

appserver/web/weld-integration/src/main/java/org/glassfish/weld/FilteringClassLoader.java
OndroMih: OK, fair points.

I updated the code in this way:
* moved exception handling one level higher, to the `isCompatible()` method. With this, we don't have to return and check a `null` value, we can return the `boolean`. We can also log a more detailed error message.
* if the MANIFEST file doesn't exist, a FINE message is logged. This is because we can tolerate a missing MANIFEST file and automatically assume the extension is compatible
* If reading the MANIFEST file fails for another IO reason, the same message is logged as before, but as ERROR instead of WARNING. This can be a serious issue with the JAR file, so ERROR is logged, but we still continue and assume the JAR file is compatible.

dmatej (approved):

PR #24566, opened 2023-08-28
Fixes generated EJB remote interfaces loading in Embedded GlassFish

### LINKED ISSUE #24521: Can't deploy app with remotely accessible bean to embedded glassfish - IllegalArgumentException xxx_Remote referenced from a method is not visible from class loader
Can't deploy a very simple web application with a remotely-accessible bean to a running embedded glassfish instance; attempt gives an IllegalArgumentException with message '<interface name>_Remote referenced from a method is not visible from class loader'.

## Environment Details
Glassfish 7.0.6 embedded

java version "17.0.7" 2023-04-18 LTS
Java(TM) SE Runtime Environment (build 17.0.7+8-LTS-224)
Java HotSpot(TM) 64-Bit Server VM (build 17.0.7+8-LTS-224, mixed mode, sharing)

MacOS 12.4 (Monterey)

Database n/a

----------

Created a very simple web app with a single remotely accessible bean with one method. Deployed the app to a running non-embedded glassfish 7.0.6 successfully. Ran an SE client that looks the bean up (rather than an enterprise client that injects the bean). Successfully got a reference to the bean and accessed the bean's method.

Created another SE client that successfully starts glassfish 7.0.6 embedded and attempts to deploy the same app war prior to looking up and accessing the bean as above. Deploy fails as described. Deploy should succeed allowing the bin to be looked up and accessed. 

Problem is reliably reproducible with this and two other apps.

Stack trace and program output attached.

## Steps to reproduce
1. Compile attached RemoteCalculatorInterface.java as a class library; no library JARs required
2. Compile attached RemoteCalculatorBean as a glassfish 7 web app; requires library JARs as per in-code documentation
3. Deploy web app to a running glassfish 7 server; deploy works. 
4. Compile and run the attached GF7EmbeddedDeployReproducerRemoteClient as a Java application; requirements are documented in the code; works fine showing that the deploy worked and that bean lookup and the bean work
5. Shut down the running glassfish 7 server to prevent conflicts with the embedded server to be run in the next step
6. Compile and run the attached GF7EmbeddedDeployReproducerEmbeddedClient as a Java application; requirements are documented in the code; deploy fails as described

## Impact of Issue
16,000 man-hour project being upgraded from Java 8 / glassfish 6 / Netbeans 8 to Java 17 / glassfish 7 / Netbeans 18. This is almost a development stopper.

[Stack trace.txt](https://github.com/eclipse-ee4j/glassfish/files/12189850/Stack.trace.txt)
[GF7EmbeddedDeployReproducerEmbeddedClient.txt](https://github.com/eclipse-ee4j/glassfish/files/12190052/GF7EmbeddedDeployReproducerEmbeddedClient.txt)
[GF7EmbeddedDeployReproducerRemoteClient.txt](https://github.com/eclipse-ee4j/glassfish/files/12190053/GF7EmbeddedDeployReproducerRemoteClient.txt)
[RemoteCalculatorBean.txt](https://github.com/eclipse-ee4j/glassfish/files/12190054/RemoteCalculatorBean.txt)
[RemoteCalculatorInterface.txt](https://github.com/eclipse-ee4j/glassfish/files/12190055/RemoteCalculatorInterface.txt)

avpinchuk:
This is a bug fix.

Fixes #24521.

Changes generated class loading criteria:

* If run-time `Package` of a generated class are defined by provided loader, then load class with  the `MethodHandles.Lookup.defineClass()` method,
* Load by one of a `ClassLoader.defineClass()` protected methods otherwise.

### DIFF
--- a/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/codegen/Generator.java
+++ b/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/codegen/Generator.java
@@ -45,6 +45,7 @@ import static org.glassfish.pfl.dynamic.codegen.spi.Wrapper._package;
  * The base class for all code generators.
  */
 public abstract class Generator {
+
     private static final Logger LOG = Logger.getLogger(Generator.class.getName());
 
     private static final String DEFAULT_PACKAGE_NAME = "";
@@ -99,7 +100,7 @@ public abstract class Generator {
      * which change between JDK versions.
      *
      * @return {@link Class}
-     * @throws IllegalAccessException
+     * @throws IllegalAccessException if a reflective access error occurred
      */
     public Class<?> generate() throws IllegalAccessException {
         final String packageName = getPackageName();
@@ -205,7 +206,7 @@ public abstract class Generator {
                 nodups.add(method);
             }
         }
-        return nodups.toArray(new Method[nodups.size()]);
+        return nodups.toArray(new Method[0]);
     }
 
 
--- a/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/codegen/Generator.java
+++ b/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/codegen/Generator.java
@@ -206,7 +206,7 @@ public abstract class Generator {
                 nodups.add(method);
             }
         }
-        return nodups.toArray(new Method[0]);
+        return nodups.toArray(Method[]::new);
     }
 
 
--- a/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/codegen/Generator.java
+++ b/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/codegen/Generator.java
@@ -1,6 +1,6 @@
 /*
+ * Copyright (c) 2021, 2023 Contributors to the Eclipse Foundation.
  * Copyright (c) 1997, 2018 Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 2021-2022 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -47,6 +47,8 @@ import static org.glassfish.pfl.dynamic.codegen.spi.Wrapper._package;
 public abstract class Generator {
     private static final Logger LOG = Logger.getLogger(Generator.class.getName());
 
+    private static final String DEFAULT_PACKAGE_NAME = "";
+
     private final ClassLoader loader;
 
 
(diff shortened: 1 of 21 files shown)


dmatej:
Hmmm, so lets try to verify it with the [TCK](https://ci.eclipse.org/jakartaee-tck/job/jakartaee-tck/job/10.0.x/73/) ...

dmatej:
Could we switch to the failsafe plugin? It uses jar file instead of classes and it is probably better, I am not sure, but I even think you don't need to have the change defaults which are different than for surefire. Just rename all tests to *ITest or *IT.

Also anything used just by tests should be under src/test/java, then the ejb dependency can have a test scope and for ShrinkWrap you need just a standalone package for war/ejb/ear etc.

dmatej (approved):

avpinchuk:
I'll try with tests tomorrow (or today, but later, I have 1 AM) ;)

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/codegen/Generator.java
```diff
@@ -189,7 +206,7 @@ protected Method[] removeRedundantMethods(final Method[] methods) {
                 nodups.add(method);
             }
         }
-        return nodups.toArray(new Method[nodups.size()]);
```
dmatej, line 192: Let's do that the best now:
`nodups.toArray(Method[]::new)`

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/codegen/ServiceInterfaceGenerator.java
```diff
@@ -159,6 +159,6 @@ private static Method[] calculateMethods(Class sib, Method[] initialList) {
                 }
             }
         }
-        return list.toArray(new Method[list.size()]);
```
dmatej, line 162: `list.toArray(Method[]::new)`

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/codegen/Generator.java
avpinchuk: You're right! Thank you for advice. Fixed.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/codegen/ServiceInterfaceGenerator.java
avpinchuk: Fixed too.

dmatej (approved):

PR #24571, opened 2023-09-03
Configure the Concurrency TCK tests to produce log files in the "target" directory

OndroMih:


avpinchuk (approved):

dmatej:
It would be better if they would be ignored under the target directory instead. Can you take a look if there is some option we should set for that?

OndroMih:
> It would be better if they would be ignored under the target directory instead. Can you take a look if there is some option we should set for that?

I agree. I found a solution with `workingDirectory` for the `surefire` and `failsafe` plugins that I already used here: https://github.com/eclipse-ee4j/glassfish/pull/24576.

OndroMih:
I found even a simpler solution, @dmatej I think this is what you meant initially :)

OndroMih:
Please squash the commits when merging

arjantijms (approved):

dmatej:
> I found even a simpler solution, @dmatej I think this is what you meant initially :)

Yeah, that was exactly that! I deleted those logs so many times but I was always lazy/distracted to resolve this issue, thanks! :-)

PR #24582, opened 2023-09-08
Update to derby 10.16.1.1

ctabin:
Update to the [latest version](https://db.apache.org/derby/releases/release-10_16_1_1.cgi) of Derby.

### DIFF
--- a/appserver/pom.xml
+++ b/appserver/pom.xml
@@ -174,7 +174,7 @@
         <!-- Other -->
         <dbschema.version>6.7</dbschema.version>
         <schema2beans.version>6.7</schema2beans.version>
-        <derby.version>10.15.2.0</derby.version>
+        <derby.version>10.16.1.1</derby.version>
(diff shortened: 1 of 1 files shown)


ctabin:
Regarding the [issues](https://db.apache.org/derby/releases/release-10_16_1_1.cgi#Issues), I'm not sure it is actually a good idea since it won't run on JVM less that JDK17... Maybe that would be better suited for the next major update GlassFish 8 ? :thinking:

arjantijms:
Yes, we've been holding off on doing this upgrade until we change over to JDK 17 at least. This should indeed be targeted to GlassFish 8.

arjantijms:
There are later versions of Derby available now, and a PR for that should target 8.0. I'll close this one, but update for Derby very welcome in 8.0.

Do note that just updating Derby doesn't seem to work and a little bit extra seems to be necessary.

PR #24605, opened 2023-09-25
Using glassfishbuild-maven-plugin 4.0.0 + fixes

dmatej:
- Reviewed all assemblies and custom lifecycles
- Fixed file permissions: ug+rw(x)
- Permissions are really controlled just for distributed artifacts.
- Depends on https://github.com/eclipse-ee4j/glassfish-build-maven-plugin/pull/145

### DIFF
--- a/appserver/distributions/pom.xml
+++ b/appserver/distributions/pom.xml
@@ -76,6 +76,7 @@
         <template.jar>${stage.dir}/${install.dir.name}/glassfish/common/templates/gf/appserver-domain.jar</template.jar>
         <create-domain.args>--user admin create-domain --template=${template.jar} --nopassword --savelogin=true --checkports=false --adminport 4848 --instanceport 8080 --keytooloptions CN=localhost domain1</create-domain.args>
         <glassfishbuild-maven-plugin.assemblyInputDirectory>${stage.dir}</glassfishbuild-maven-plugin.assemblyInputDirectory>
+        <asadmin.executable>${stage.dir}/${install.dir.name}/bin/asadmin${file.executable.suffix}</asadmin.executable>
     </properties>
 
     <build>
@@ -142,7 +143,7 @@
                                 <goal>exec</goal>
                             </goals>
                             <configuration>
-                                <executable>${stage.dir}/${install.dir.name}/bin/asadmin</executable>
+                                <executable>${asadmin.executable}</executable>
                                 <commandlineArgs>${create-domain.args}</commandlineArgs>
                             </configuration>
                         </execution>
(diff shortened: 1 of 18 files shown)


avpinchuk:
We need also raise Maven version to 3.9.0+ in [prerequisites](https://github.com/eclipse-ee4j/glassfish#prerequisites).

appserver/distributions/pom.xml
```diff
@@ -104,88 +108,66 @@
...
-                                    </fileset>
-                                    <fileset>
-                                        <fileMode>640</fileMode>
-                                        <dir>${stage.dir}</dir>
-                                        <excludes>**/bin/*</excludes>
-                                        <excludes>**/bin/*.bat</excludes>
-                                        <excludes>**/lib/nadmin*</excludes>
-                                    </fileset>
-                                </filesets>
+                                <executable>${stage.dir}/${install.dir.name}/bin/asadmin</executable>
```
avpinchuk, line 145: Failed on Windows

dmatej:
> We need also raise Maven version to 3.9.0+ in [prerequisites](https://github.com/eclipse-ee4j/glassfish#prerequisites).

Yes, that is why I still  keep it as a Draft PR. I want to check also if it compiles on Windows too.

nucleus/distributions/pom.xml
```diff
@@ -92,69 +99,66 @@
...
-                                        <includes>**/lib/nadmin*</includes>
-                                    </fileset>
-                                    <fileset>
-                                        <dir>${stage.dir}</dir>
-                                        <excludes>**/bin/*</excludes>
-                                        <excludes>**/bin/*.bat</excludes>
-                                        <excludes>**/lib/nadmin*</excludes>
-                                    </fileset>
-                                </filesets>
+                                <executable>${stage.dir}/${nucleus.install.dir.name}/bin/nadmin</executable>
```
avpinchuk, line 136: Failed on Windows

appserver/distributions/pom.xml
dmatej: I hate Windows
- git clone failed, too long file paths, resolved by git config
- gjule tests failed, `\r`. For some reason `-Dfile.separator='\n'` doesn't have any effect. Test skipped.
- exousia 2.1.1 is missing in maven central ... reported to @arjantijms 

I will get there, can you give me some hint how to avoid the `'\r'` issue?

avpinchuk: @dmatej, instead of `split("\n")` you sould use either `System.lineSeparator()` or `\\r?\\n` pattern in place of `\n`. In compiled patterns use same pattern `\\r?\\n` instead of `\\n`. E.g. `\\n\\n` would be translated to `(\\r?\\n){2}` and so on

Hope help :)

nucleus/distributions/pom.xml
avpinchuk: Here we should use script without extension on UNIX, and `.bat` on Windows. 

Hint: I make a quick build with adding extension via profile - all builds. May be found a better way ;)

dmatej: It should be fixed now

appserver/distributions/pom.xml
dmatej: Hmm, I hoped I can somehow just force windows to switch to unix lines, but it seems I really have to do this.

nucleus/distributions/pom.xml
avpinchuk: Yes, now it works.

pzygielo (approved):

avpinchuk (approved):

PR #24611, opened 2023-09-28
Improve GlassFish process runtime type detection

avpinchuk:
Detects EMBEDDED instance at runtime.

 This fixes all resources cleanup at shutdown.

### DIFF
--- a/nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/StaticGlassFishRuntime.java
+++ b/nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/StaticGlassFishRuntime.java
@@ -160,6 +160,7 @@ public class StaticGlassFishRuntime extends GlassFishRuntime {
         String installRootValue = System.getProperty("org.glassfish.embeddable.installRoot");
         if (installRootValue == null) {
             installRootValue = instanceRoot.getAbsolutePath();
+            gfProps.setProperty("-type", "EMBEDDED");
             JarUtil.extractRars(installRootValue);
         }
         JarUtil.setEnv(installRootValue);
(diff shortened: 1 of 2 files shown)


dmatej:
Hmm, I have a perhaps one year old branch with fixes of resource leaks, time to push that. Not sure if this was in between, probably not, but I know I have also new bugs in that, that's why it is still waiting ... however I am not sure how many conflicts I have to resolve, I didn't rebase it few versions ...

dmatej (approved):

nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/StaticGlassFishRuntime.java
```diff
@@ -171,84 +166,78 @@ private void setEnv(GlassFishProperties gfProps) throws Exception {
...
+            // If the user has specified a custom domain.xml then copy it.
             String configFileURI = gfProps.getConfigFileURI();
             if(configFileURI != null) {
-                copy(URI.create(configFileURI).toURL(),
-                        new File(instanceConfigDir, "domain.xml"), true);
+                copy(URI.create(configFileURI).toURL(), new File(instanceConfigDir, "domain.xml"), true);
             }
         } catch (Exception ex) {
-            ex.printStackTrace();
+            LOG.log(SEVERE, "Failed to create instanceRoot", ex);
```
dmatej, line 220: Can you do an experiment here for sure? If it is not possible to create an instance root -> it is not possible to load also logging.properties (maybe) -> no logging config available -> the error might end in the log buffer forever. Just try to throw an exception from this block and check if it is logged or not.
If not, revert this change and add comment why do we want to use STDERR instead of the JVM logging system.
I never know which part is executed  from which process. :-)

avpinchuk: @dmatej, it works ;)

image

dmatej (approved):

PR #24616, opened 2023-10-01
Fixes for resource leaks

dmatej:
I have this branch here since November 2022, rebasing it from time to time. Perhaps it would be good to merge it.

Summary of Changes:
- In general - replacing GlassFish custom implementations where are modern implementations in the JDK already available.
- Using `try-with`
- Safer `ZipEntry` management, new closeable `WritableArchiveEntry`
- Revisited `FileUtils`
  - deleted unused methods
  - added `USER_HOME` file
  - new `ensureWritableDir` method; if the directory cannot be used or created, throws ISE
- `ProprietaryReader.readFrom` closes the input OR sets that to a returned object
  - The behavior is not much consistent and I don't like it, but now it is documented and better controlled.
  - Implementations are final to limit risks.
- Fix: `MIMEMessage` is closeable!
- Fix: `JsonParser` doesn't close the input if it is a stream!
- `AsadminSecurityUtil.GF_CLIENT_DIR` constant
  - by default subdir in user's home as it was before this commit
  - can be reset by `GF_CLIENT_DIR` environment property.
  - MUST be writeable, client should crash otherwise.
- `ArchivistUtils` deleted - duplicit to `FileUtils`
- Fixed JAR/ZIP file management, entries  
    - Safer impl without resource leaks (I hope, there's yet one class with really wide scope and also some finalizers, should be revisited later again).
    - More logs - it is hard to find what is wrong when it just does something in the background.
    - I wanted to write some tests around `DeploymentContext`, but it has too many dependencies, so it is covered just by integration tests (persistence_all ie)

The build locally passed:

```
mvn clean install
...
[INFO] BUILD SUCCESS
[INFO] Total time:  21:19 min
[INFO] Finished at: 2024-02-20T19:05:05+01:00
```

I have read my own code now, it took me around 90 minutes. This is not a final state, there is several FIXMEs documenting discovered potential leaks which were too complicated for my brain and also they would need yet more radical changes. However this PR should improve the reliability+speed of GF a bit.


dmatej:
My past me sent a message :D
```
     * FIXME: Document and write test with large zip file to see if it wouldn't be better to use
     * BufferedInputStream and BufferedOutputStream.
     * FIXME: Usually used with JarFile and it's entry. Add API.

```

nucleus/admin/util/src/main/java/com/sun/enterprise/admin/remote/reader/MultipartProprietaryReader.java
```diff
@@ -69,50 +72,54 @@ public ParamsWithPayload readFrom(final InputStream is, final String contentType
...
+                        payload.add(name, stream2String(mimePart.readOnce()), mimePart.getAllHeaders());
+                    } else {
+                        payload.add(name, mimePart.read(), ct, mimePart.getAllHeaders());
+                    }
                 }
             }
+            // Result
+            return new ParamsWithPayload(payload, parameters, actionReport);
+        } finally {
+            is.close();
```
hs536: Is it safe to close the stream passed in as a method argument inside the method?

> public ParamsWithPayload readFrom(final InputStream is, final String contentType) throws IOException {

nucleus/admin/util/src/main/java/com/sun/enterprise/admin/remote/reader/ProgressStatusDTOJsonProprietaryReader.java
```diff
@@ -55,6 +58,7 @@ public ProgressStatusDTO readFrom(final InputStream is, final String contentType
             return readProgressStatus(jp);
         } finally {
             jp.close();
+            is.close();
```
hs536, line 61: wondering about the same point:
https://github.com/eclipse-ee4j/glassfish/pull/24616#discussion_r1496778273

nucleus/cluster/cli/src/main/java/com/sun/enterprise/admin/cli/cluster/SetupSshKey.java
```diff
@@ -53,65 +57,55 @@ public final class SetupSshKey extends NativeRemoteCommandsBase {
...
                 }
             }
-        }
-        else {
-            promptPass = SSHUtil.validateKeyFile(sshkeyfile);
-            if (SSHUtil.isEncryptedKey(sshkeyfile)) {
+        } else {
+            File keyFile = new File(sshkeyfile);
+            promptPass = SSHUtil.validateKeyFile(keyFile);
+            if (SSHUtil.isEncryptedKey(new File(sshkeyfile))) {
```
hs536, line 86: Isn't it unnecessary to re-new File(sshkeyfile) here since the `keyFile` is created just above?

nucleus/admin/util/src/main/java/com/sun/enterprise/admin/remote/reader/MultipartProprietaryReader.java
dmatej: The short answer is no and this  was reverted in the next commit. But it is even worse, this has to be rewritten completely, that is why I added this comment:
```
        // FIXME: Big refactoring required because of possible resource leaks:
        // MIMEMessage produces MIMEPart objects based on the is parameter.
        // All these objects are Closeable.
        // The final result of this method is the payload, based on mimePart object.
        // That is used later, and at least in one branch none of these objecs can be closed,
        // because the payload's input stream would be closed too.
```
The whole implementation is not "symmetric" for try-with, so close methods were originally "here and there" in the code. And still are and ie RemoteRestAdminCommand still can leak. I will revisit that in another iteration.

nucleus/admin/util/src/main/java/com/sun/enterprise/admin/remote/reader/ProgressStatusDTOJsonProprietaryReader.java
dmatej: Here it is because I tried to push the close method of the input stream closer to where it should be closed. Usually these input streams are byte array input streams or input streams from HttpUrlConnection, which should be closed too - in try-with, but that is impossible in current impl.

I wrote the javadoc for the readFrom method: 
`Reads and closes the stream, or can transfer both respnsibilities to a returned object.`

Originally some branches did not close the stream at all. Some branches even don't close the connection, seems it is left for garbage collector, but I am not sure with that. The old code is quite overcomplicated, hard to find all ends. So I am sure this is not the final version, rather a "milestone" of the refactoring. I would like to continue in smaller steps.

hs536 (approved):

avpinchuk (approved):

arjantijms (approved):
LGTM. TBH, didn't read each and every line.

PR #24626, opened 2023-10-08
Refactored building tools for Ant-based tests

dmatej:
* Fixed closing resources
* reportbuilder and reporter moved to the standard Maven lifecycle
   * Before their usage by Ant scripts execute a special local maven project which collects dependencies for tests - it is more visible than using Ant's copy task and also it is less work to add/remove something.* Removed binaries from the lib directory of Ant tests
* Some Ant tasks were reimplemented - created custom HTTP clients able to send authentication requests (original used Tomcat's GTask from 2001, which doesn't have successor).
* Some tests were moved to application-tests, executed by maven. 
   * I will probably continue in another PR with the rest of security_all tests soon


pzygielo:
Is that necessary to have `appserver/tests/appserv-tests/lib/reportbuilder.jar` checked-in?

dmatej:
> Is that necessary to have `appserver/tests/appserv-tests/lib/reportbuilder.jar` checked-in?

I had the same question, but I have to admit that for now I resigned. It is small and the module stands outside the standard reactor for some reason. It seems it is directly used by Ant based tests, so I would have to change it also in Jenkinsfile, etc. 

But perhaps I should try to change it in this PR, it should be doable ...

appserver/security/core-ee/src/main/java/com/sun/enterprise/security/ee/jmac/MessageSecurityConfigEventListenerImpl.java
```diff
@@ -86,7 +86,10 @@ public <T extends ConfigBeanProxy> NotProcessed changed(TYPE type, Class<T> chan
 
     private <T extends ConfigBeanProxy> NotProcessed handle(T instance) {
         if (instance instanceof MessageSecurityConfig) {
-            AuthConfigFactory.getFactory().refresh();
```
dmatej, line 89: Fixed NPE

appserver/security/webservices.security/src/main/java/com/sun/enterprise/security/webservices/server/ServerSecurityPipe.java
```diff
@@ -157,6 +157,7 @@ private Packet processRequest(Packet request) throws Exception {
             authenticationService.auditInvocation(validatedRequest, status);
         }
 
+        _logger.log(FINE, "Validation status: {0}", status);
```
dmatej, line 160: I moved the log and fixed the log message - the original message key did not exist.

appserver/tests/application/pom.xml
```diff
@@ -122,6 +149,29 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
```
dmatej, line 152: The original Ant based impl executed wsgen, while in maven it would lead to a chicken - egg problem. So I created WSDL files, used the plugin to generate required classes which were then available for the test.

appserver/tests/application/src/test/java/org/glassfish/main/test/app/security/jmac/https/JmacHttpsTest.java
```diff
@@ -0,0 +1,203 @@
...
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.glassfish.main.itest.tools.GlassFishTestEnvironment.getDomain1Directory;
+import static org.glassfish.main.itest.tools.GlassFishTestEnvironment.getDomain1TrustStore;
+import static org.glassfish.main.itest.tools.GlassFishTestEnvironment.openConnection;
+import static org.glassfish.main.itest.tools.asadmin.AsadminResultMatcher.asadminOK;
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.stringContainsInOrder;
+
+@Disabled("Fails with GF 7.0.11, the attribute jakarta.servlet.request.X509Certificate is not set for some reason.")
```
dmatej, line 63: This test was failing in Ant too, I have found that the attribute can be set by Catalina or by Grizzly, but it is not set. I have no idea why, it seems as a bug. I transformed the test to Maven, so now it should be easier to reproduce and fix it.
I remember we extended this feature in Payara, so it was possible to use even parts of the DN, maybe we were fixing yet something before, but I don't remember what.

avpinchuk (approved):

PR #24630, opened 2023-10-10
Refactor Realm and related code

arjantijms:
* Split out the over complicated Realm class is a few pieces that are easier to understand
* Specifically, create an interface for getting user data (used by the authentication code) and an interface for managing user data (used by admin console)
* Split between the all static methods and the stateful instance methods
* Remove classes not longer in use and/or deprecated
* Move all realm related exceptions to their own package

Note:

There is more to be done still, especially the individual realm implementations can be refactored wrt enhanced for loops, pairing with their LoginModule peer, etc.


hs536 (approved):

nucleus/security/core/src/main/java/com/sun/enterprise/security/auth/realm/RealmsManagerHolder.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+package com.sun.enterprise.security.auth.realm;
+
+import java.lang.ref.WeakReference;
+
+import org.glassfish.internal.api.Globals;
+
+public class RealmsManagerHolder {
+
+    private static WeakReference<RealmsManager> realmsManagerReference = new WeakReference<RealmsManager>(null);
```
avpinchuk, line 25: @arjantijms, `realmsManagerReferense` must be volatile to ensure visibility by other threads. Or method `getRealmsMnager()` must be synchronized instead of `_getRealmsManager()`.

avpinchuk (changes requested):

appserver/security/core-ee/src/main/java/com/sun/enterprise/security/auth/realm/jdbc/JDBCRealm.java
```diff
@@ -139,37 +140,31 @@ public synchronized void init(Properties props) throws BadRealmException, NoSuch
...
         String groupNameColumn = props.getProperty(PARAM_GROUP_NAME_COLUMN);
         String groupTableUserNameColumn = props.getProperty(PARAM_GROUP_TABLE_USER_NAME_COLUMN, userNameColumn);
-        cr = (ActiveDescriptor<ConnectorRuntime>) Util.getDefaultHabitat()
-                .getBestDescriptor(BuilderHelper.createContractFilter(ConnectorRuntime.class.getName()));
+
+        connectorRuntimeDescriptor = (ActiveDescriptor<ConnectorRuntime>)
+            getDefaultHabitat().getBestDescriptor(BuilderHelper.createContractFilter(ConnectorRuntime.class.getName()));
 
         if (jaasCtx == null) {
-            String msg = sm.getString("realm.missingprop", BaseRealm.JAAS_CONTEXT_PARAM, "JDBCRealm");
```
dmatej, line 146: String.format uses $s, while MessageFormat is faster and uses {0} .

appserver/security/core-ee/src/main/java/com/sun/enterprise/security/auth/realm/jdbc/JDBCRealm.java
```diff
@@ -509,24 +508,24 @@ private void close(Connection conn, PreparedStatement stmt, ResultSet rs) {
...
                 connection = dataSource.getConnection(dbUser, dbPassword);
             } else {
                 connection = dataSource.getConnection();
             }
+
             return connection;
         } catch (Exception ex) {
-            String msg = sm.getString("jdbcrealm.cantconnect", dsJndi, dbUser);
-            LoginException loginEx = new LoginException(msg);
+            LoginException loginEx = new LoginException(String.format("Unable to connect to datasource {0} for database user {1}.", dsJndi, dbUser));
```
dmatej, line 528: Same here

nucleus/security/core/src/main/java/com/sun/enterprise/security/auth/realm/file/FileRealm.java
```diff
@@ -180,23 +177,23 @@ protected void init(Properties props) throws BadRealmException, NoSuchRealmExcep
...
+        _logger.log(FINE, "FileRealm : " + JAAS_CONTEXT_PARAM + "={0}", jaasCtx);
 
         try {
             if (isEmbeddedServer()) {
                 file = writeConfigFileToTempDir(file).getAbsolutePath();
             }
             fileRealmHelper = new FileRealmHelper(file);
         } catch (IOException ioe) {
-            throw new BadRealmException(sm.getString("filerealm.noaccess", ioe.toString()));
+            throw new BadRealmException(String.format("Unable to create keyfile: {0}", ioe.toString()));
```
dmatej, line 196: here

nucleus/security/core/src/main/java/com/sun/enterprise/security/auth/realm/file/FileRealm.java
```diff
@@ -222,7 +219,7 @@ public Enumeration<String> getUserNames() throws BadRealmException {
     public User getUser(String name) throws NoSuchUserException {
         FileRealmHelper.User user = fileRealmHelper.getUser(name);
         if (user == null) {
-            throw new NoSuchUserException(sm.getString("filerealm.nouser", name));
+            throw new NoSuchUserException(String.format("No such user [{0}]", name));
```
dmatej, line 222: here

nucleus/security/core/src/main/java/com/sun/enterprise/security/auth/realm/ldap/LDAPRealm.java
```diff
@@ -189,52 +186,53 @@ public synchronized void init(Properties props) throws BadRealmException, NoSuch
...
         ldapBindProps.setProperty(Context.PROVIDER_URL, url);
-        this.setProperty(PARAM_USERDN, dn);
-        this.setProperty(BaseRealm.JAAS_CONTEXT_PARAM, jaasCtx);
+        setProperty(PARAM_USERDN, dn);
+        setProperty(JAAS_CONTEXT_PARAM, jaasCtx);
 
         String mode = props.getProperty(PARAM_MODE, MODE_DEFAULT);
         if (!MODE_DEFAULT.equals(mode)) {
-            throw new BadRealmException(sm.getString("ldaprealm.badmode", mode));
+            throw new BadRealmException(String.format("Unsupported mode {0}.", mode));
```
dmatej, line 203: here ... perhaps search for {0} in this diff.

dmatej (changes requested):

nucleus/security/core/src/main/java/com/sun/enterprise/security/auth/realm/RealmsManagerHolder.java
arjantijms: Good point, although this is something that was already in the existing code. I just moved it to another class (which is why I also kept the original Oracle copyright)

nucleus/security/core/src/main/java/com/sun/enterprise/security/auth/realm/ldap/LDAPRealm.java
arjantijms: I made the same mistake everywhere 🫣

nucleus/security/core/src/main/java/com/sun/enterprise/security/auth/realm/ldap/LDAPRealm.java
```diff
@@ -189,52 +187,53 @@ public synchronized void init(Properties props) throws BadRealmException, NoSuch
 
         String url = props.getProperty(PARAM_DIRURL);
         String dn = props.getProperty(PARAM_USERDN);
-        String jaasCtx = props.getProperty(BaseRealm.JAAS_CONTEXT_PARAM);
+        String jaasCtx = props.getProperty(JAAS_CONTEXT_PARAM);
 
         if (isAnyNull(url, dn, jaasCtx)) {
-            throw new BadRealmException(sm.getString("ldaprealm.badconfig", url, dn, jaasCtx));
+            throw new BadRealmException(String.format(
```
avpinchuk, line 193: And here

appserver/security/core-ee/src/main/java/com/sun/enterprise/security/auth/realm/jdbc/JDBCRealm.java
```diff
@@ -139,37 +141,31 @@ public synchronized void init(Properties props) throws BadRealmException, NoSuch
...
         if (jaasCtx == null) {
-            String msg = sm.getString("realm.missingprop", BaseRealm.JAAS_CONTEXT_PARAM, "JDBCRealm");
-            throw new BadRealmException(msg);
+            throw new BadRealmException(String.format(MISSING_PROPERTY, JAAS_CONTEXT_PARAM, "JDBCRealm"));
         }
 
         if (dsJndi == null) {
-            String msg = sm.getString("realm.missingprop", PARAM_DATASOURCE_JNDI, "JDBCRealm");
-            throw new BadRealmException(msg);
+            throw new BadRealmException(String.format(MISSING_PROPERTY, PARAM_DATASOURCE_JNDI, "JDBCRealm"));
```
avpinchuk, line 153: Here an below in six places `MISSING_PROPERTY` constant uses `{...}` placeholders

arjantijms: Thanks again, I think I now finally got them all!

avpinchuk (approved):

dmatej (approved):

dmatej:
Just one note - shouldn't we release 7.1.0 at the end of the month? This seems like a breaking change, I am not sure, I did not write a realm and login module for a while, but some classes changed packages (exceptions, ie).

arjantijms:
It does depend on what people would use from GlassFish internals. I personally think that if internal code changes (with the exception of the classes listed in the xml files) it would not constitute a breaking change in the traditional sense.

Of course the problem is that GlassFish for building LoginModules/Realms doesn't distinguish between external and internal code. Types that people can use for custom LoginModules/Realms should have gone into one of the many GlassFish API modules, but obviously this has never been done.

dmatej:
This is rather an API of GlassFish ... years ago I added the security-ee in my dependencies, with `provided` scope, implemented realm and login module, and then I had a script to install it into the GlassFish (copy the jar file, update `server.policy` and `login.conf`). 
When you upgrade from 7.0.9 to 7.0.10 (7.1.0), you have to rebuild your library. Sad we don't have any tests for this, perhaps we should write them. Until that this is rather my hypothesis, but I would expect library compiled against 7.0.9 would not work with current 7.0.10-SNAPSHOT.

PR #24639, opened 2023-10-18
Fix the startserv script for Mac OS

OndroMih:
Works even with zsh.

I tested only in Linux, so needs to be tested on Mac OS by somebody.

To test, it should be enough to execute the 2 scenarios:

* `bin/startserv` - should run successfully and start the `domain1` domain
* `bin/startserv non-existent-domain` - should end in an error and print the error from the GF launcher (cannot start domain `non-existent-domain`)


OndroMih:
@arjantijms , please, test on Mac.

arjantijms:
Works perfectly, thanks!

arjantijms (approved):

PR #24644, opened 2023-10-21
Draft: Adds lookup in system properties for WAR contextRoot in EAR

ctabin:
Hi GlassFish team !

I'm looking up to implement a simple system property lookup in GlassFish so we can externalize `contextRoot` and `urlPattern` (and more, such as JNDI names, EJB persistance unit names, ...).

This is mainly a prototype to see how this can be implemented since I don't know how it would be the most correct place to put this translation, neither if this is the correct way to do it (for instance, in Payara it only looks for the properties declared in the `domain.xml`). The goal of this PR is to get some input to see in which way this can go.

Based on my [example](https://github.com/ctabin/gf-test-lucene), the goal is to have a definition like below, so we can build an EAR where the URLs can be configured outside the application itself (so we could eventually have the EAR/WAR/JAR file signed):
```xml
<modules>
  <webModule>
    <groupId>ch.astorm.gf-embedded</groupId>
    <artifactId>war</artifactId>
    <contextRoot>/${ch.astorm.root}</contextRoot>
    <bundleFileName>sample-war.war</bundleFileName>
  </webModule>
  <ejbModule>
    <groupId>ch.astorm.gf-embedded</groupId>
    <artifactId>ejb</artifactId>
    <bundleFileName>sample-ejb.jar</bundleFileName>
  </ejbModule>
</modules>
```

Here are some places that I'd like to see this implemented:
- context root WAR in the EAR
- URL patterns in `@WebServlet`
- `jta-data-source` in `persistence.xml`

Any feedback is welcome :-)

### DIFF
--- a/nucleus/deployment/common/src/main/java/org/glassfish/deployment/common/ModuleDescriptor.java
+++ b/nucleus/deployment/common/src/main/java/org/glassfish/deployment/common/ModuleDescriptor.java
@@ -24,6 +24,8 @@ import java.util.jar.Manifest;
 
 import org.glassfish.api.deployment.archive.ArchiveType;
 import org.glassfish.deployment.versioning.VersioningUtils;
+import org.jvnet.hk2.config.TranslationException;
+import org.jvnet.hk2.config.VariableResolver;
 
 /**
  * This class describes a module information for an applicaiton module
@@ -145,7 +147,13 @@ public class ModuleDescriptor<T extends RootDeploymentDescriptor> extends Descri
      * @param contextRoot the contextRoot
      */
     public void setContextRoot(String contextRoot) {
-        this.contextRoot = contextRoot;
+        VariableResolver resolver = new VariableResolver() {
+            @Override
+            protected String getVariableValue(String varName) throws TranslationException {
+                return System.getProperty(varName, "");
+            }
+        };
+        this.contextRoot = resolver.translate(contextRoot);
     }
 
     /**


nucleus/deployment/common/src/main/java/org/glassfish/deployment/common/ModuleDescriptor.java
```diff
@@ -145,7 +147,13 @@ public T getDescriptor() {
      * @param contextRoot the contextRoot
      */
     public void setContextRoot(String contextRoot) {
-        this.contextRoot = contextRoot;
+        VariableResolver resolver = new VariableResolver() {
```
OndroMih, line 150: It's better to use the existing functionality in [TranslatedConfigView.expandValue()](https://github.com/eclipse-ee4j/glassfish/blob/master/nucleus/admin/config-api/src/main/java/org/glassfish/config/support/TranslatedConfigView.java#L52). This method is used to expand values in domain.xml and supports references to system properties and password aliases. In the future, we can also add support to MicroProfile Config values (e.g. also env variables).

I suggest creating a new static method in `TranslatedConfigView`, e.g. `expandValueForDescriptor` and delegate to the `expandValue()` method. This is just in case we need to add additional behavior when dealing with deployment descriptors. E.g. disable it with a config property like Payara did in this PR because it broke some TCK tests: https://github.com/payara/Payara/pull/4132.

OndroMih:
You can see most, if not all places where Payara supports expansion of deployment descriptor values or annotation values in this PR: https://github.com/payara/Payara/pull/4132/files.

URL patterns are set in [WebComponentDescriptorImpl.addUrlPattern()](https://github.com/eclipse-ee4j/glassfish/blob/master/appserver/web/web-glue/src/main/java/org/glassfish/web/deployment/descriptor/WebComponentDescriptorImpl.java#L203C17-L203C30) method. Other @WebServlet properties can be expanded there too. For servletName, `WebServletHandler` uses it to find an existing web servlet definition if it exists. The `servletName` has to be translated also in this search. However, I would do it inside the [WebBundleDescriptor.getWebComponentByCanonicalName](https://github.com/eclipse-ee4j/glassfish/blob/0244305efd0436b767e7a8c2dde0d40916a8fb10/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/WebBundleDescriptor.java#L1221) 

`persistence.xml` values are I think set in setters in [PersistenceUnitDescriptor](https://github.com/eclipse-ee4j/glassfish/blob/master/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/PersistenceUnitDescriptor.java).

nucleus/deployment/common/src/main/java/org/glassfish/deployment/common/ModuleDescriptor.java
ctabin: Hi @OndroMih,
Thanks for the hints ! I'll take a look and come back with some similar implementation.

OndroMih:
We at OmniFish are working on this in another branch so that such lookup works in all deployment descriptors and annotations, not only for the contextRoot property. I'll close this PR now. We'll create a PR with the more complete solution soon.

ctabin:
@OndroMih Amazing :thumbsup: Looking forward to see this implemented in GlassFish and sorry I didn't have the time to work on this.

PR #24647, opened 2023-10-22
Update README.md

arjantijms:


dmatej (approved):
Did we really pass the TCK with 21? Nope (not everything runs on it, but we also didn't pass the TCK with JDK 18). The text there is just a general statement about Java versions.

PR #24655, opened 2023-10-25
Avoid protocol relative redirects (Fixes CVE-2023-41080)

arjantijms:
See: https://github.com/apache/tomcat/commit/e3703c9abb8fe0d5602f6ba8a8f11d4b6940815a

Fixes [CVE-2023-41080](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-41080)

### DIFF
--- a/appserver/web/web-core/src/main/java/org/apache/catalina/authenticator/FormAuthenticator.java
+++ b/appserver/web/web-core/src/main/java/org/apache/catalina/authenticator/FormAuthenticator.java
@@ -532,6 +532,11 @@ public class FormAuthenticator extends AuthenticatorBase {
             sb.append(saved.getQueryString());
         }
 
+        // Avoid protocol relative redirects
+        while (sb.length() > 1 && sb.charAt(1) == '/') {
+            sb.deleteCharAt(0);
+        }
+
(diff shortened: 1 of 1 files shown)


avpinchuk:
Looks good. But will give us absolute-path referense same redirect URI as protocol relative uri?

dmatej (approved):

dmatej:
> Looks good. But will give us absolute-path reference same redirect URI as protocol relative uri?

Ah, sorry I didn't notice that you added the question and merged the PR. @arjantijms , please, answer...

OndroMih:
> Looks good. But will give us absolute-path reference same redirect URI as protocol relative uri?

The point here is to not allow an attacker to post data which contain a redirect URL with protocol relative URL, which opens a vulnerability [CVE-2023-41080](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-41080), which is fixed in the recent Tomcat in the same way:  https://tomcat.apache.org/security-11.html#Fixed_in_Apache_Tomcat_11.0.0-M11

arjantijms:
Note to self; we do need to check this at several other places too.

PR #24665, opened 2023-10-31
Clean unhandled messages in admingui

kaido207:
I checked where `System.out.println` is used in the admingui project.

### DIFF
--- a/appserver/admingui/jms-plugin/src/main/java/org/glassfish/admingui/plugin/jms/JmsHandlers.java
+++ b/appserver/admingui/jms-plugin/src/main/java/org/glassfish/admingui/plugin/jms/JmsHandlers.java
@@ -131,10 +131,12 @@ public class JmsHandlers {
     protected static final String PROP_NAME = "name";
     protected static final String PROP_DEST_TYPE = "desttype";
 
-    @Handler(id = "getPhysicalDestination", input = {
-            @HandlerInput(name = "name", type = String.class, required = true),
-            @HandlerInput(name = "type", type = String.class, required = true) }, output = {
-                    @HandlerOutput(name = "destData", type = java.util.Map.class) })
+    @Handler(id = "getPhysicalDestination",
+    input = {
+        @HandlerInput(name = "name", type = String.class, required = true),
+        @HandlerInput(name = "type", type = String.class, required = true) },
+    output = {
+        @HandlerOutput(name = "destData", type = java.util.Map.class) })
     public static void getPhysicalDestination(HandlerContext handlerCtx) {
         String name = (String) handlerCtx.getInputValue("name");
         String type = (String) handlerCtx.getInputValue("type");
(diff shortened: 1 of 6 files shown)


appserver/admingui/plugin-service/src/main/java/org/glassfish/admingui/plugin/ConsolePluginService.java
```diff
@@ -48,35 +48,39 @@
...
+ * @author Ken Paulsen (ken.paulsen@sun.com)
  */
 @Service
 public class ConsolePluginService {
-    @Inject Logger logger;
-    @Inject ServiceLocator habitat;
-    @Inject IterableProvider<ConsoleProvider> providers;
-
-/*
-    @Inject ModulesRegistry modulesRegistry;
```
dmatej, line 64: I don't think the automatic formatting of all these classes helped. Maybe this commented out code could be even deleted.

appserver/admingui/plugin-service/src/main/java/org/glassfish/admingui/plugin/ConsolePluginService.java
```diff
@@ -48,35 +48,39 @@
...
+    @Inject
+    IterableProvider<ConsoleProvider> providers;
+
+    /*
+     * @Inject ModulesRegistry modulesRegistry; for(HK2Module m : modulesRegistry.getModules()) { url =
+     * m.getClassLoader().getResource(ConsoleProvider.DEFAULT_CONFIG_FILENAME); if(url!=null) ; // TODO: parse url }
+     */
 
     /**
-     * <p> Default constructor.</p>
```
dmatej, line 73: Those` <p> `elements are worthless for a single line.

appserver/admingui/plugin-service/src/main/java/org/glassfish/admingui/plugin/ConsolePluginService.java
```diff
@@ -128,17 +130,14 @@ protected synchronized void init() {
 //System.out.println("IP Map: " + pointsByType.toString());
 
         // Log some trace messages
-        if (logger.isLoggable(Level.FINE)) {
-            logger.fine("Console Plugin Service has been Initialized!");
-            if (logger.isLoggable(Level.FINEST)) {
```
dmatej, line 133: toString is here resolved before the log method is invoked. It would be better to join both lines into 
```
logger.log(Level.CONFIG, "Console Plugin Service has been initialized. Integration points by type: \n{0}", pointsByType);
```

appserver/admingui/plugin-service/src/main/java/org/glassfish/admingui/plugin/ConsolePluginService.java
```diff
@@ -323,8 +322,7 @@ public Map<String, List<URL>> getResources(String name) {
                     } catch (Exception ex) {
                         // Ignore b/c this should not ever happen, we're not
                         // changing the URL
-                        System.out.println(
```
dmatej, line 326: I believe this is an ERROR message, not INFO. And it would be good to log the stacktrace, especially if comment says that it should not happen :-)

```
logger.log(Level.SEVERE, "ConsolePluginService: URL Copy Failed!, ex");
```

appserver/admingui/jms-plugin/src/main/java/org/glassfish/admingui/plugin/jms/JmsHandlers.java
```diff
@@ -153,17 +151,14 @@ public static void getPhysicalDestination(HandlerContext handlerCtx) {
...
 
-
         handlerCtx.setOutputValue("destData", valueMap);
     }
 
-    @Handler(id = "getPhysicalDestinationStats",
-    input = {
-        @HandlerInput(name = "name", type = String.class, required = true),
-        @HandlerInput(name = "type", type = String.class, required = true),
-        @HandlerInput(name = "target", type = String.class, required = true)},
```
dmatej, line 164: I think that it is more readable to have the output attribute on a  new line, but I can survive even this.

kaido207:
@dmatej Thank you for your review, done.

dmatej (approved):
There are still some `<p>` and logs which could be optimized, but lets move on :-)

avpinchuk (approved):

arjantijms (approved):

PR #24691, opened 2023-11-29
Fix WebSocket not working for default web module applications.

arjantijms:
This corrects the URI to include the context path without "maskDefaultContextMapping" set.
 This is the URI on which Tyrus expects to find a WebSocket. Without setting this, the
 WebSocket will not be found when accessing the application via the default web module
 URL (the context root).

 E.g. without this correction, for application "foo", "http://localhost:8080/foo" websockets would work, but
 "http://localhost:8080" would not when set to a default web module via a command like:
` asadmin set server-config.http-service.virtual-server.server.default-web-module=foo`

### DIFF
--- a/appserver/web/web-core/src/main/java/org/apache/catalina/core/WebSocketFilterWrapper.java
+++ b/appserver/web/web-core/src/main/java/org/apache/catalina/core/WebSocketFilterWrapper.java
@@ -69,7 +69,7 @@ public class WebSocketFilterWrapper implements Filter  {
 
                 @Override
                 public String getRequestURI() {
-                    RequestFacade wrappedRequest = (RequestFacade) super.getRequest();;
+                    RequestFacade wrappedRequest = (RequestFacade) super.getRequest();
                     String requestURI = wrappedRequest.getRequestURI();
 
                     // Get the contextPath without masking the default context mapping.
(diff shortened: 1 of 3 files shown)


OndroMih (approved):
Brilliant! Just added a suggestion to clean up the code.

appserver/web/web-core/src/main/java/org/apache/catalina/core/WebSocketFilterWrapper.java
```diff
@@ -0,0 +1,95 @@
...
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException, ServletException {
+        HttpServletRequest httpServletRequest = (HttpServletRequest) request;
+
+        if (httpServletRequest.getHeader(SEC_WEBSOCKET_KEY) != null) {
+            httpServletRequest = new HttpServletRequestWrapper(httpServletRequest) {
+
+                @Override
+                public String getRequestURI() {
+                    RequestFacade wrappedRequest = (RequestFacade) super.getRequest();;
```
avpinchuk, line 72: Redundant semicolon here )

arjantijms: Oh yeah, I saw that before, then somehow didn't remove it :X

avpinchuk (approved):

hs536 (approved):

PR #24702, opened 2023-12-07
fix cmtTimeoutInSeconds config, it can't take effect when set value 0 in admin UI.

zistrong:
in adminUI server-config--->Transaction Service page, if I set Transaction Timeout to value 0, accoring to the description, we should disable timeout setting. but now, it is set to 120s in hard code. I fix the bug. 
after this refactor, the default values is 120s, if we set 0, timeout setting will be disabled.

image


avpinchuk:
@zistrong, please accept ECA

avpinchuk (approved):

dmatej (approved):

hs536:
This change not only changes the default behavior of GlassFish, but also causes the following issue to recur when the Transaction Timeout is set to 0. 
https://github.com/eclipse-ee4j/glassfish/issues/21175

I'll try to come up with another fix this week.

PR #24709, opened 2023-12-16
Fix: The startserv script didn't report errors correctly

OndroMih:
The `bin/startserv` script failed to report error correctly, gave a bash syntax error instead. 

This PR also simplifies the script code that removes unwanted lines in dry run output. 
It also fixes the `glassfish/bin/startserv` script which was broken because of the wrong path. 
It also switches to use local `modules/admin-cli.jar` because it's enough and the `lib/client/appserver-cli.jar` already adds it to the classpath plus adds some more things.

### DIFF
--- a/appserver/distributions/glassfish-common/src/main/resources/bin/startserv
+++ b/appserver/distributions/glassfish-common/src/main/resources/bin/startserv
@@ -17,10 +17,10 @@
 #
 
 AS_INSTALL=`dirname "$0"`/../glassfish
-AS_INSTALL_LIB="$AS_INSTALL/modules"
 case "`uname`" in
   CYGWIN*) AS_INSTALL=`cygpath --windows $AS_INSTALL`
 esac
+AS_INSTALL_LIB="$AS_INSTALL/modules"
 . "${AS_INSTALL}/config/asenv.conf"
 JAVA=java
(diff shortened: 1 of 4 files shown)


appserver/distributions/glassfish-common/src/main/resources/bin/startserv
```diff
@@ -16,10 +17,10 @@
 #
 
 AS_INSTALL=`dirname "$0"`/../glassfish
+AS_INSTALL_LIB="$AS_INSTALL/modules"
 case "`uname`" in
   CYGWIN*) AS_INSTALL=`cygpath --windows $AS_INSTALL`
```
hs536, line 22: Hm... I don't see the need for this block, though, if this line is really needed, it may also be necessary to redefine `AS_INSTALL_LIB` in this case block using `AS_INSTALL` converted to windows path format.

OndroMih: Thanks for noticing. I moved `AS_INSTALL_LIB` definitiona under the Cygwin line.

I think not many people use Cygwin to run GlassFish, but it already was in the `lgassfish/bin/asadmin` script (although not in `bin/asadmin`), so I decided to keep support for Cygwin.

hs536 (approved):

PR #24733, opened 2024-01-09
Reduce usage of Java SE Security Manager

arjantijms:


appserver/appclient/client/acc/src/main/java/org/glassfish/appclient/client/acc/ACCClassLoader.java
```diff
@@ -18,31 +18,26 @@
 
 package org.glassfish.appclient.client.acc;
 
+import static java.security.AccessController.doPrivileged;
```
avpinchuk, line 21: `doPriveleged()` at lines 69 and 134 are OK?

avpinchuk (approved):

appserver/appclient/client/acc/src/main/java/org/glassfish/appclient/client/acc/ACCClassLoader.java
arjantijms: I can still remove them. But this PR just takes a first step at removing them. Many, many more need to be removed. My plan is to do that in followup PRs.

arjantijms:
There are some tests failing now, but these are at a glance testing "permission.xml", which has been removed. So those tests need to go too.

dmatej (approved):

PR #24740, opened 2024-01-12
Give authentication modules masked request

### LINKED ISSUE #24739: Authentication mechanisms for the default context see wrong context root
When an application is deployed on a virtual server and set to a default context root via e.g.

```
./asadmin set standaloneInst1-config.http-service.virtual-server.server.default-web-module=SomeName
```

Then authentication modules, both Jakarta Authentication based ones (and by extension of those, Jakarta Security ones), and native ones (Tomcat based) get to see the "SomeName" context root when accessed via the default root (""). When such authentication mechanism issues for instance a redirect it may use "request.getRequestURI", which then includes "SomeName".

As a result, the URL changes for the user. The app continues to work, as in GlassFish with a default web module both "/" and "/"SomeName" are supported at the same time, but it is confusing and doesn't look good.

arjantijms:
Fixes #24739 

Authentication modules (Jakarta Authentication and native) did not get the masked request, but the raw request. Masking is needed for applications that are on the default context root ("/") to they get to see "" as a context root instead of "/someappname".


avpinchuk:
Checkstyle violations

avpinchuk (approved):

Issue #24065, opened 2022-08-03
Launcher should not touch server.log


dmatej:
The first log record in server.log is not printed by the server, but by it's launcher. That means ...

* the launcher prints the log even if the server startup failed. The cause is not printed anywhere.
  - it would be better if it would print this log record and the cause to the stderr instead.
  - the log leads to a confusion aka _"server logged something into the server.log, so i started ... and then crashed"_ while it did not start at all.

* the launcher uses it's own configuration and ignores server's logging.properties
  - it should have own configuration and own file OR stderr+stdout
  - first log record can have different log format than what is configured for the server.log

* refactoring required, there is quite a lot of copy&paste, dead code, etc.

PR #24743, opened 2024-01-14
Fixes around logging

### LINKED ISSUE #24737: GlassFishlogHandler can drop several LogRecord instances when it rotates the server.log file
## Environment Details
* GlassFish Version (and build number): 7.0.x - 7.0.11
* JDK version: 17
* OS: Windows, Linux

----------

## Problem Description
StreamHandler.isLoggable in JUL returns false when writer is null. The writer is null when the stream is to be closed; GJULE closes the stream when it rolls the file, but still resolves loggability of messages before putting them into the input buffer.

It relates to the #24736 where domain1 GlassFish Service cannot roll server.log files, but I missed several added logs.

dmatej:
- The stop-domain did not print that the domain is not running, but just some method name instead.
- When rolling the server.log file, several messages was lost. Closes #24737 
- Reduced using method handles as parameters, better separation of features.
- Some editors (Vim) delete before save, sometimes GF noticed the logging.properties vanished and then GF stopped its monitoring and reloading. As the file is critical, on this event now GF creates a special thread which resets everything to correct state after the file reappears.
- Rolling the server.log file did not work on Windows when used with --verbose or --watch (also used by Windows Services), because Launcher had it's own handle to the file and Windows blocked attempts to move the file. Now Launcher removed its handler right before it starts the server which configures its own logging and continues writing to the server.log. This relates to #24065 .

### DIFF
--- a/nucleus/admin/launcher/src/main/java/com/sun/enterprise/admin/launcher/GFLauncher.java
+++ b/nucleus/admin/launcher/src/main/java/com/sun/enterprise/admin/launcher/GFLauncher.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
  * Copyright (c) 2008, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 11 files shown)


avpinchuk:
`rotate-log` command works on Windows 10 for GlassFish started with `--verbose` and `--watchdog` options

avpinchuk (approved):

arjantijms (approved):

mirozitnansky:
It seems to be working in test environment, but we still didn't deploy it to the production.

PR #24744, opened 2024-01-15
Add debug option to start-cluster, to start all instances in debug mode

arjantijms:
This adds a "--debug" option to the "start-cluster" command, analogous to the "start-domain" command, and "start-instance" command. It will pass the "--debug" option to each individual "start-instance" command that will be called from  "start-domain".

e.g.

```
./asadmin start-cluster --debug clusterA
```

### DIFF
--- a/nucleus/cluster/admin/src/main/manpages/com/sun/enterprise/v3/admin/cluster/start-cluster.1
+++ b/nucleus/cluster/admin/src/main/manpages/com/sun/enterprise/v3/admin/cluster/start-cluster.1
@@ -4,7 +4,8 @@ NAME
        start-cluster - starts a cluster
 
 SYNOPSIS
-           start-cluster [--help] [--autohadboverride={true|false}]
+           start-cluster [--help] 
+           [--debug={false|true}] [--autohadboverride={true|false}]
            [--verbose={false|true}] cluster-name
 
 DESCRIPTION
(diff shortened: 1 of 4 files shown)


dmatej (commented):
Does it work well when you have instances of the same cluster on a single machine? (usually when you do some local testing).

nucleus/cluster/admin/src/main/manpages/com/sun/enterprise/v3/admin/cluster/start-cluster.1
```diff
@@ -38,6 +39,23 @@ DESCRIPTION
 OPTIONS
        --help, -?
            Displays the help text for the subcommand.
+       
+       --debug
+           Specifies whether each instance in the cluster is started with Java Platform
+           Debugger Architecture (JPDA)
+
+           (http://www.oracle.com/technetwork/java/javase/tech/jpda-141715.html)
```
avpinchuk, line 47: Link are alive?

avpinchuk (approved):

dmatej (approved):

nucleus/cluster/admin/src/main/manpages/com/sun/enterprise/v3/admin/cluster/start-cluster.1
arjantijms: Probably some updating needed, but across the board.

This one seems to be the correct one: https://docs.oracle.com/en/java/javase/17/docs/specs/jpda/jpda.html

PR #24748, opened 2024-01-16
Optimize servlet caching filter

avpinchuk:
Eliminates possible unnecessary multiple loads of a same resource under high request rate


dmatej (approved):

hs536 (approved):
LGTM. The nesting will become deeper, but it seems that with this addition, we can avoid the problem without locking the cache.

PR #24750, opened 2024-01-17
Raise bytecode level

avpinchuk:
Set up ASM-generated bytecode level to `V17`


dmatej (approved):

arjantijms:
Note that we have to revert the JDK 21 changes, and go back to JDK 17. See https://github.com/jakartaee/platform/issues/820

However, for GlassFish 8.1 which we can release very quickly after Jakarta EE 11 goes final we can go back to 21.

avpinchuk:
@arjantijms, thank you very much for the link!

PR #24753, opened 2024-01-17
Try the application classloader to restore the Principal for HA.

### LINKED ISSUE #24752: HA single sign on can't restore custom Principal from application
If an authentication mechanism sets a custom principal where the class resides within the application archive, then this principal fails to deserialise an another node when clustering and HASingleSingOn is used.

The problem seems to be that the HASingleSingOn valve (which restores the principal) runs before the valve that sets the context class loader to the one that knows about the application provided classes. Hence, the class definition can't be found.

arjantijms:
Fixes #24752 

This allows a custom Principal to be used that's defined by the application (the .class file resides in a .war).

The application classloader is tried last so that the existing behaviour is preserved as much as possible.

### DIFF
--- a/appserver/web/web-ha/src/main/java/org/glassfish/web/ha/authenticator/HASingleSignOnEntry.java
+++ b/appserver/web/web-ha/src/main/java/org/glassfish/web/ha/authenticator/HASingleSignOnEntry.java
@@ -190,8 +190,7 @@ public class HASingleSignOnEntry extends SingleSignOnEntry {
 
             try {
                 return (Principal) ois.readObject();
-            }
-            catch (ClassNotFoundException e) {
+            } catch (ClassNotFoundException e) {
                 closeSafely(bais);
                 closeSafely(bis);
                 closeSafely(ois);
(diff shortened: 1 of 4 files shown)


appserver/web/web-ha/src/main/java/org/glassfish/web/ha/authenticator/HASingleSignOnEntry.java
```diff
@@ -180,7 +183,25 @@ private Principal parse(byte[] pbytes) {
...
+                return (Principal) ois.readObject();
+            }
+
+            try {
+                return (Principal) ois.readObject();
+            }
+            catch (ClassNotFoundException e) {
+                closeSafely(bis);
+                closeSafely(ois);
+                closeSafely(ois);
```
hs536, line 197: `ois` is double-closed here, shouldn't it be `bais`?

arjantijms: That was a typo indeed. I wonder if those inner streams really need to be closed explicitly. Specially closing the byte array stream doesn't do anything really, and the outer one should propagate to the inner ones anyway.

But I wanted, for now, to keep the code as close as possible to the original.

avpinchuk (approved):

PR #24754, opened 2024-01-18
Use standard proxy default method invocation handler

avpinchuk:
Replace custom proxy default method invocation handler with standard (since Java 16) one.


dmatej (approved):

hs536 (approved):
It's a curious API that didn't exist as a standard until JDK 16.

dmatej:
Hmmm, I tried to introduce JPMS to GF7, but this is a blocker. In GF8 is already used the InvocationHandler, so I will retry there ...
https://bugs.openjdk.org/browse/JDK-8209078
```
java.lang.IllegalAccessException: symbolic reference class is not accessible: interface org.jvnet.hk2.config.test.example.GenericConfig, from class com.sun.proxy.jdk.proxy2.$Proxy38 (module jdk.proxy2), previous lookup org.jvnet.hk2.config.ProxyHelper (module org.glassfish.main.hk2.config.generator)
at java.base/java.lang.invoke.MemberName.makeAccessException(MemberName.java:894)
at java.base/java.lang.invoke.MethodHandles$Lookup.checkSymbolicClass(MethodHandles.java:3787)
at java.base/java.lang.invoke.MethodHandles$Lookup.resolveOrFail(MethodHandles.java:3747)
at java.base/java.lang.invoke.MethodHandles$Lookup.findSpecial(MethodHandles.java:3097)
at org.glassfish.main.hk2.config.generator@7.0.18-SNAPSHOT/org.jvnet.hk2.config.ProxyHelper.defaultMethodHandle(ProxyHelper.java:96)
... 12 more
```
I am putting it here just for the record if I would forget and tried that again :-/
I tried to find workaround, but the jdk.proxy2 module is really invisible outside the JDK.

Issue #24784, opened 2024-02-04
Documentation modules: 'administration-guide' and 'error-references-module' fail to build on Windows


escay:
## Environment Details
* GlassFish Version (and build number): glassfish master (7.0.13-SNAPSHOT 418e00767ab1b937e4d32bcc4a8205467ce75c21)
* JDK version: openjdk version "17.0.10" 2024-01-16
* OS: Windows 11

----------

## Problem Description
`mvn install` fails on Windows for two modules:

```
/glassfish/docs/administration-guide
/glassfish/docs/error-messages-reference
```

Both fail the same:
```
djesc@Z13 MINGW64 /c/github/glassfish/docs/administration-guide (master)
$ mvn install
[INFO] Scanning for projects...
[INFO]
[INFO] ------------< org.glassfish.main.docs:administration-guide >------------
[INFO] Building Eclipse GlassFish Administration Guide, Release 7 7.0.13-SNAPSHOT
[INFO]   from pom.xml
[INFO] --------------------------------[ jar ]---------------------------------
[INFO]
[INFO] --- enforcer:3.1.0:enforce (enforce-maven) @ administration-guide ---
[INFO]
[INFO] --- osgiversion:3.0.6:compute-osgi-version (generate-osgi-version-property) @ administration-guide ---
[INFO]
[INFO] --- compiler:3.11.0:compile (generate-hk2-config) @ administration-guide ---
[INFO] No sources to compile
[INFO]
[INFO] --- glassfish-doc:1.3:book (generate-book) @ administration-guide ---
[INFO]
[INFO] --- build-helper:3.4.0:add-resource (add-resource) @ administration-guide ---
[INFO]
[INFO] --- build-helper:3.4.0:add-resource (add-legal-resource) @ administration-guide ---
[INFO]
[INFO] --- dependency:3.6.1:unpack (unpack-resource) @ administration-guide ---
[INFO] Configured Artifact: org.glassfish.main:nucleus-parent:resources:7.0.13-SNAPSHOT:zip
[INFO]
[INFO] --- resources:3.3.1:resources (default-resources) @ administration-guide ---
[INFO] skip non existing resourceDirectory C:\github\glassfish\docs\administration-guide\src\main\resources
[INFO] skip non existing resourceDirectory C:\github\glassfish\docs\administration-guide\src\main\manpages
[INFO] Copying 2 resources from target\common-resources\legal to target\classes\META-INF
[INFO]
[...]

PR #24785, opened 2024-02-04
Fixes issue #24784, cannot build doc modules on Windows

escay:
Fixes issue #24784, cannot build doc modules on Windows
ECA username 'rmisingnamesda'

### DIFF
--- a/docs/administration-guide/src/main/asciidoc/logging.adoc
+++ b/docs/administration-guide/src/main/asciidoc/logging.adoc
@@ -381,10 +381,10 @@ If there already is a file with the same name, the implementation tries to add a
 with a counter until it finds a name which doesn't exist yet.
 
 ----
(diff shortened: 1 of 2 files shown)


arjantijms (approved):

hs536 (approved):
LGTM

avpinchuk:
I think this is a particular issue of the MinGW and WLS. C with caron is a valid UTF character.

Documentation builds without any errors on native Windows, Linux and MacOS.

dmatej (changes requested):
čec is czech abbreviation for červenec (= July), so as a compromise you can change it to jul, but not to cec, because cec would be nonsense.

dmatej (approved):

avpinchuk (approved):

PR #24793, opened 2024-02-06
Fixes data race during Admin Console initialization

avpinchuk:
Synchronize access to fields accessed from many threads.


hs536:
What kind of problems can the current code cause?
( Is there a possibility of conflict in the loading process of the RestModule...?

dmatej (approved):
I believe I can see it. The init method can be called by different thread than service method, but they use the same attribute, which can sometimes lead to a situation when the service method would use an old value.

avpinchuk:
> The init method can be called by different thread than service method, but they use the same attribute, which can sometimes lead to a situation when the service method would use an old value.

Yes, this is also one of the possible problems

hs536 (approved):

avpinchuk:
I will make several builds of the master branch at different time. I hope this fix help stabilize `application-tests`.

PR #24802, opened 2024-02-10
Fix the link to the example hello.war

### LINKED ISSUE #24801: doc: broken link to hello.war
## Environment Details
Glassfish master branch as of 2024-02-10

## Problem Description
In the documentation:
https://glassfish.org/docs/latest/quick-start-guide.html#to-obtain-the-sample-application
the link to hello.war results in a file not found error.

## Steps to reproduce
$ curl -vw\\n https://github.com/eclipse-ee4j/glassfishdownloads/quickstart/hello.war
[...]
* Connected to github.com (140.82.121.3) port 443
[...]
> GET /eclipse-ee4j/glassfishdownloads/quickstart/hello.war HTTP/2
[...]
< HTTP/2 404 
[...]
Not Found

## Impact of Issue
The tutorial cannot be performed.

libnoon:
The old link is defunct.  Updated the link.
Fixed #24801.

### DIFF
--- a/docs/quick-start-guide/src/main/asciidoc/basic-features.adoc
+++ b/docs/quick-start-guide/src/main/asciidoc/basic-features.adoc
@@ -322,7 +322,7 @@ The following topics are addressed here:
 ==== To Obtain the Sample Application
 
 1. Download a copy of the `hello.war` sample application from
-`https://github.com/eclipse-ee4j/glassfishdownloads/quickstart/hello.war`.
+`https://raw.githubusercontent.com/eclipse-ee4j/glassfish/master/appserver/tests/appserv-tests/devtests/cluster/apps/helloworld.war`.
 
 2. Save the `hello.war` file in the directory of your choice.
 +


libnoon:
Hmm it looks like the link I suggest as replacement does not contain the same war as the intended one, because the tutorial https://glassfish.org/docs/latest/quick-start-guide.html#deploying-and-undeploying-the-sample-application-from-the-command-line seems to rely on an app which displays a prompt page and responds to it, while the war I'm linking only prints a hello world without interaction.

But then, where is the correct war?

dmatej:
I would say it should be here, but it really doesn't exist. It probably was on original Oracle website somewhere, then we updated links using regexps, but we still did not validate all targets.

But honestly I don't see any advantage of using a binary file without sources as an example. Perhaps we should rewrite it somehow better.

dmatej:
I would say all binaries should be here (and in Maven Central too): https://download.eclipse.org/ee4j/glassfish/

OndroMih:
There's a working link on the old GlassFish web site. But I suggest to fix it properly, as I described here: https://github.com/eclipse-ee4j/glassfish/issues/24801#issuecomment-2031707120

libnoon:
It looks like a better fix for this problem has been merged.  Closing this PR.  Thanks!

PR #24806, opened 2024-02-12
GlassFishORBHelper changed to singleton, lifecycle

dmatej:
- cherrypicked from #24616 + added two commits with code cleanup
- Original commit was created more than year ago, I think it was related to some race condition I noticed in logs, but now I am not sure. However I believe it is still bit better than older version. The CORBA/ORB initialization is a horrible thing - maybe it would be easier to rewrite it instead of refactoring. But that would need to know every piece of it first ...

### DIFF
--- a/appserver/orb/orb-connector/src/main/java/org/glassfish/enterprise/iiop/api/GlassFishORBHelper.java
+++ b/appserver/orb/orb-connector/src/main/java/org/glassfish/enterprise/iiop/api/GlassFishORBHelper.java
@@ -129,11 +129,9 @@ public class GlassFishORBHelper implements ORBLocator {
                 // Any callbacks that result from the protocol manager initialization itself cannot
                 // depend on having access to the protocol manager.
                 orb = orbFactory.createORB(props);
-
-                if (!isServer || protocolManager != null) {
-                    return orb;
+                if (isServer && protocolManager == null) {
+                    protocolManager = initProtocolManager(orb);
                 }
-                protocolManager = initProtocolManager(orb);
                 return orb;
             } catch (Exception e) {
                 orb = null;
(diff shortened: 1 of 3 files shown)


appserver/orb/orb-connector/src/main/java/org/glassfish/enterprise/iiop/api/GlassFishORBHelper.java
```diff
@@ -112,55 +103,54 @@ public ORB getORB() {
...
-        if (orb == null) {
-
-            synchronized (this) {
-                if (orb == null) {
-                    try {
-                        boolean isServer = processEnv.getProcessType().isServer();
-
-                        Properties props = new Properties();
-                        props.setProperty(GlassFishORBFactory.ENV_IS_SERVER_PROPERTY, Boolean.valueOf(isServer).toString());
+        if (orb != null || destroyed) {
```
avpinchuk, line 106: A small note on optimization ))

We can read volatile `orb` field into a local variable and work with it. This will reduce number of the volatile reads.

avpinchuk (approved):

arjantijms (approved):

appserver/orb/orb-connector/src/main/java/org/glassfish/enterprise/iiop/api/GlassFishORBHelper.java
dmatej: As it is outside the sync block it could even change between usages. It is not much probable here, but it was still a mistake. I have extracted the initProtocolManager method with the ORB parameter too.

appserver/orb/orb-connector/src/main/java/org/glassfish/enterprise/iiop/api/GlassFishORBHelper.java
```diff
@@ -108,55 +103,58 @@ public ORB getORB() {
...
+                // won't be available.
+                // Any callbacks that result from the protocol manager initialization itself cannot
+                // depend on having access to the protocol manager.
+                orb = orbFactory.createORB(props);
+
+                if (!isServer || protocolManager != null) {
+                    return orb;
                 }
+                protocolManager = initProtocolManager(orb);
+                return orb;
```
avpinchuk, line 137: Maybe simplify this?

```java
if (isServer && protocolManager == null) {
    protocolManager = initProtocolManager(orb);
}
return orb;
```

dmatej: Right!

PR #24817, opened 2024-02-20
Fixes #24816 Non-ASCII characters are not saved correctly in JVM options added using the JVM Options page

### LINKED ISSUE #24816: Non-ASCII characters are not saved correctly in JVM options added using the JVM Options page
## Environment Details
* GlassFish Version (and build number): GlassFish 7.0.8
* OS: CentOS Linux release 7
* Database: none

----------

## Problem Description
Non-ASCII characters are not saved correctly in JVM options added using the JVM Options page.  

jvm_unicode1
jvm_unicode2

"-Dtest=あああ" is converted to "-Dtest=u3042u3042u3042", this is not expected behavior.  

You can also confirm that it is not the expected JVM option value by using the asadmin command:
```
# /glassfish7/glassfish/bin/asadmin list-jvm-options
--add-opens=java.base/java.io=ALL-UNNAMED
:
-Xmx512m
-Dtest=u3042u3042u3042
-Dorg.glassfish.gmbal.no.multipleUpperBoundsException=true
:
-XX:NewRatio=2
Command list-jvm-options executed successfully.
```
## Steps to reproduce
1. access Admin Console.
2. open `Configulation` > `server-config` > `JVM Settings`.
3. open `JVM Options`. 
4. add the JVM option that contain Non-ASCII characters.
5. `save` clicked.

## Impact of Issue
Users cannot create JVM options that contain Non-ASCII characters in Admin GUI, such as directory paths.

kaido207:
* Fixes #24816

Escaping non-ASCII strings is unnecessary in HTTP and was causing the issue.  

also I found a bug around the call to `UtilHandlers.escapePropertyValue()` and fixed it.  
`replaceAll("\\\\", "\\\\\\\\")` on line 280 in `SecurityHandler.java` is unnecessary.  
https://github.com/eclipse-ee4j/glassfish/blob/d322a5705f498a16210ceee9e6b7762a9cfa739c/appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/SecurityHandler.java#L280  
Because the char `\` is escaped in `escapePropertyValue` after that.  
https://github.com/eclipse-ee4j/glassfish/blob/d322a5705f498a16210ceee9e6b7762a9cfa739c/appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/UtilHandlers.java#L1089  
This line was causing the following redundant escapes when the realm was created with the property such as windows paths:
realm_backslash
realm_backslash2

### DIFF
--- a/appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/SecurityHandler.java
+++ b/appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/SecurityHandler.java
@@ -278,7 +278,7 @@ public class SecurityHandler {
                 continue;
             }
             sb.append(oneProp.get("name")).append("=");
-            String value = ((String) oneProp.get("value"));
+            String value = (String) oneProp.get("value");
             value = UtilHandlers.escapePropertyValue(value);
             sb.append(value).append(":");
         }
--- a/appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/SecurityHandler.java
+++ b/appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/SecurityHandler.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation.
  * Copyright (c) 2009, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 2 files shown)


dmatej (approved):
I think years ago I noticed some related inconsistencies too, good to see this fixed!

appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/SecurityHandler.java
```diff
@@ -277,7 +278,7 @@ public static void saveRealm(HandlerContext handlerCtx) {
                 continue;
             }
             sb.append(oneProp.get("name")).append("=");
-            String value = ((String) oneProp.get("value")).replaceAll("\\\\", "\\\\\\\\");
+            String value = ((String) oneProp.get("value"));
```
hs536, line 281: the outer brackets of the value are no longer needed :)

kaido207: Thank you for noticing that, I fixed.

dmatej (approved):

Issue #23483, opened 2021-06-03
It didn't reach the limit of max-pool-size, but there was an maxPoolSize exhausted exception


micRo-wlY:
Hi!

We hava an application that uses jdbc connection pooling.And uses the following settings:
max-pool-size:200
pool-resize-quantity:2
max-wait-time-in-millis:5000

It has been working fine for a long time, but recently we get the following exception while trying to concurrently access.

[java.sql.SQLException:Error in allocating a connection. Cause: In-use connections equal max-pool-size and expired max-wait-time. Cannot allocate more connections.]

The number of concurrent requests is 100.
Actually,It didn't exceed the limit of max-pool-size 200.

By analyzing the ConnectionPool code,we find a problem in ConnectionPool#resizePoolAndGetNewResource.

In order to understand the problem simply,suppose uses the following settings:
max-pool-size:10
pool-resize-quantity:1
max-wait-time-in-millis:5000

and,the problem flow is as follows:
1. now there are five connections in the jdbc connection pool and all these connections are in-used

2. one request comes to get connection(by Thread A)

3. because there's still space to create a connection(current connections < max-pool-size),
   Thread A starts to execute the following resizePoolAndGetNewResource. 
   In **1 method create 1 new connection, and add connection to connection pool.  
   
4. thread A  before executing **2 method (take the connection off from connection pool)
   thread B just comes to get connection,and just takes the connection off from connection pool.
   
5. when thread A executes **2 method,the connection was tooken by thread B, returns null.
   and max-wait-time-in-millis is timeout,in **3 method,the error will happen.

[java.sql.SQLException:Error in allocating a connection. Cause: In-use connections equal max-pool-size and expired max-wait-time. Cannot allocate more connections.]

## Environment Details
* GlassFish Version:5.1.0
* JDK version:1.8.0_201
* OS:ALL
* Database:ALL

## Questions about this case
Now,we hava one question:

When a thread execute the resizePoolAndGetNewResource, should this thread get the connection first?  
If YES,the code of resizePoolAndGetNewResource has a bug.
If NO,the error message [Cause: In-use connections equal max-pool-size and expired max-wait-time.] is not right.

## Related source
com.sun.enterprise.resource.pool.ConnectionPool
```
	private ResourceHandle resizePoolAndGetNewResource(ResourceAllocator alloc) throws PoolingException {
        //Must be called from the thread holding the lock to this pool.
        ResourceHandle result = null;
[...]

PR #24851, opened 2024-03-11
Fixes issue #24849 make relevant methods synchronized in LocalTxConnectionEventListener 

### LINKED ISSUE #24849: Improve Thread safeness of LocalTxConnectionEventListener
Improve Thread safeness of LocalTxConnectionEventListener to prevent errors like:
```
Caused by: java.lang.NullPointerException: Cannot invoke "com.sun.enterprise.resource.ResourceHandle.getResourceSpec()" because "resource" is null
        at com.sun.enterprise.resource.pool.PoolManagerImpl.resourceClosed(PoolManagerImpl.java:367)
        at com.sun.enterprise.resource.listener.LocalTxConnectionEventListener.connectionClosed(LocalTxConnectionEventListener.java:58)
        at com.sun.gjc.spi.ManagedConnectionImpl.connectionClosed(ManagedConnectionImpl.java:735)
        at com.sun.gjc.spi.base.ConnectionHolder.close(ConnectionHolder.java:193)
        at com.sun.gjc.spi.jdbc40.ConnectionHolder40.close(ConnectionHolder40.java:590)
```
to improve the contents of the associatedHandles Map to be more threadsafe.

As part of the investigation of #24805 I ran into similar Payara issue: https://github.com/payara/Payara/issues/3029 which I logged in the past. This was fixed in https://github.com/payara/Payara/commit/c09e8aafd53221a918230db5d2c9c1098c5f2a76 

Fix idea:
- improve calls to "private IdentityHashMap associatedHandles;" to be more thread safe, preventing poolManager.resourceClosed(handle); to be called if another thread already updated associatedHandles.
- refactor com.sun.enterprise.resource.ConnectorXAResource.resetAssociation() logic -> so no external classes are updating associatedHandles anymore, making updating the associatedHandles the responsibility of class LocalTxConnectionEventListener

escay:
Make relevant methods synchronized in LocalTxConnectionEventListener and protect associatedHandles from external clear calls.

I first tried to move most of the ConnectorXAResource.resetAssociation method inside a synchronous method in the LocalTxConnectionEventListener, but that also meant two other methods from ConnectorXAResource needed to be refactored and moved into LocalTxConnectionEventListener. So I undid that and chose for cloning the map and returning a clone to make the method getAssociatedHandles thread safe. This is a different solution as used in the Payara fix I mention in the issue. But I think it is better to leave the association and disassociating of ManagedConnection logic inside the ConnectorXAResource code.

I did not add multi threading unit tests, since only the basic synchronized(this) approach is used to protect the associatedHandles map. No need to test java. I did add some basic unit test logic for the add and remove logic used by ConnectorXAResource.

### DIFF
--- a/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/ConnectorXAResource.java
+++ b/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/ConnectorXAResource.java
@@ -124,9 +124,7 @@ public class ConnectorXAResource implements XAResource {
                 if (handle != null) { // not needed, just to be sure.
                     ManagedConnection associatedConnection = (ManagedConnection) handle.getResource();
                     associatedConnection.associateConnection(userHandle);
-                    if (_logger.isLoggable(FINE)) {
-                        _logger.log(FINE, "connection_sharing_reset_association", userHandle);
-                    }
+                    _logger.log(FINE, "connection_sharing_reset_association", userHandle);
                 }
             }
         } catch (Exception e) {
@@ -241,9 +239,7 @@ public class ConnectorXAResource implements XAResource {
                 ResourceHandle associatedHandle = (ResourceHandle) userHandleEntry.getValue();
                 ManagedConnection associatedConnection = (ManagedConnection) associatedHandle.getResource();
                 associatedConnection.associateConnection(userHandleEntry.getKey());
-                if (_logger.isLoggable(FINE)) {
-                    _logger.log(FINE, "connection_sharing_reset_association", userHandleEntry.getKey());
-                }
+                _logger.log(FINE, "connection_sharing_reset_association", userHandleEntry.getKey());
             }
         } catch (Exception ex) {
             handleResourceException(ex);
--- a/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/ConnectorXAResource.java
+++ b/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/ConnectorXAResource.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, 2022 Contributors to the Eclipse Foundation
+ * Copyright (c) 2021, 2024 Contributors to the Eclipse Foundation
  * Copyright (c) 1997, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -21,7 +21,7 @@ import static java.util.logging.Level.FINE;
 import static java.util.logging.Level.SEVERE;
 
 import java.util.Map;
-import java.util.Set;
+import java.util.Map.Entry;
 import java.util.logging.Logger;
 
 import javax.transaction.xa.XAException;
@@ -46,8 +46,12 @@ import jakarta.transaction.SystemException;
  */
 public class ConnectorXAResource implements XAResource {
 
-    static Logger _logger = LogDomains.getLogger(ConnectorXAResource.class, LogDomains.RSR_LOGGER);
+    private static Logger _logger = LogDomains.getLogger(ConnectorXAResource.class, LogDomains.RSR_LOGGER);
 
+    /**
+     * userHandle meaning: an object representing the "connection handle for the underlying physical connection". In some
+     * code also named connectionHandle.
+     */
     private Object userHandle;
     private ResourceSpec spec;
     private ClientSecurityInfo info;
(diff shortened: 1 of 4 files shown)


appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/ConnectorXAResource.java
```diff
@@ -120,7 +124,9 @@ public void end(Xid xid, int flags) throws XAException {
                 if (handle != null) { // not needed, just to be sure.
                     ManagedConnection associatedConnection = (ManagedConnection) handle.getResource();
                     associatedConnection.associateConnection(userHandle);
-                    _logger.log(FINE, "connection_sharing_reset_association", userHandle);
```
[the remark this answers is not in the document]
escay: You mean I can remove the
if (_logger.isLoggable(FINE)) {

in both cases. I added it here, to be in sync / reflect the resetAssociation() original code which did use the if.
I will remove both ifs.

dmatej (approved):

appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/listener/LocalTxConnectionEventListener.java
```diff
@@ -32,37 +32,51 @@
...
-        this.associatedHandles = new IdentityHashMap(10);
-        this.poolManager = ConnectorRuntime.getRuntime().getPoolManager();
     }
 
     @Override
-    public void connectionClosed(ConnectionEvent evt) {
+    public synchronized void connectionClosed(ConnectionEvent evt) {
         Object connectionHandle = evt.getConnectionHandle();
         ResourceHandle handle = resource;
         if (associatedHandles.containsKey(connectionHandle)) {
```
avpinchuk, line 65: Here we do two map lookups. Maybe it's better to use

```java
ResourceHande handle = associatedHandes.getOrDefault(connectionHandle, resource);
```

appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/listener/LocalTxConnectionEventListener.java
```diff
@@ -72,12 +86,15 @@ public void connectionErrorOccurred(ConnectionEvent evt) {
      * @param evt ConnectionEvent
      */
     @Override
-    public void badConnectionClosed(ConnectionEvent evt) {
+    public synchronized void badConnectionClosed(ConnectionEvent evt) {
         Object connectionHandle = evt.getConnectionHandle();
         ResourceHandle handle = resource;
         if (associatedHandles.containsKey(connectionHandle)) {
```
avpinchuk, line 92: Here too

avpinchuk: Also, because `synchronized` is only used to guard access to the associated handles map, maybe it's better to use `ReentrantReadWriteLock` instead?

avpinchuk:
Also, because `synchronized` is only used to guard access to the associated handles map, maybe it's better to use `ReentrantReadWriteLock` instead?

appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/listener/LocalTxConnectionEventListener.java
escay: Thank you, I was not aware of this getOrDefault option on Maps.
Documentation of getOrDefault defines: "The default implementation makes no guarantees about synchronization or atomicity properties of this method." so I think the synchronization on the method is still needed.
I can make an update later today.

avpinchuk: You're right, synchronization is needed. We simply eliminate unnecessary map access.

escay:
> Also, because `synchronized` is only used to guard access to the associated handles map, maybe it's better to use `ReentrantReadWriteLock` instead?

I cannot oversee the impact of using ReentrantReadWriteLock on only the associatedHandles. Some code like the
 `poolManager.resourceClosed(handle);` calls could also be affected by the synchronized approach. I have no easy way to test this, I know the synchronized version works for me for the last years in Payara: https://github.com/payara/Payara/commit/c09e8aafd53221a918230db5d2c9c1098c5f2a76 to avoid the exception, so I tried to stay near that solution.
 
To know for sure I think I need to have found the real cause of issue #24805 and probably also of issue  #23483 which I think could be related and run my test set many times. 
I am also looking at: 

https://github.com/eclipse-ee4j/glassfish/blob/3e5203c56e63abf68cece6ddc76cf9ffd91c3d69/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/pool/ConnectionPool.java#L679C1-L682C66

where I suspect resourceHandles can get mixed up, and could perhaps cause this issue as side effect.

avpinchuk (approved):

PR #24852, opened 2024-03-11
Fixed report.dir in the Authentication TCK execution

dmatej:
Now it generates the "ts harness" HTML report. It seems the junit xml report is not supported by this TCK, am I right?


arjantijms:
Ohw, this version of the TCK is quite old. Should have been replaced a while ago, but wasn't.

dmatej:
> Ohw, this version of the TCK is quite old. Should have been replaced a while ago, but wasn't.

I work with what I have here for now :-)

arjantijms (approved):

PR #24855, opened 2024-03-12
The secureResponse can return just SEND_* states (based on javadoc)

dmatej:


avpinchuk (approved):

avpinchuk:
The secification also says [this](https://jakarta.ee/specifications/authentication/3.0/jakarta-authentication-spec-3.0#a317) (at point (3) in the message processing model).

appserver/tests/application/src/main/java/org/glassfish/main/test/app/security/jmac/https/HttpsTestAuthModule.java
```diff
@@ -99,7 +99,7 @@ public AuthStatus validateRequest(final MessageInfo messageInfo, final Subject c
 
     @Override
     public AuthStatus secureResponse(final MessageInfo messageInfo, final Subject serviceSubject) throws AuthException {
-        return AuthStatus.SUCCESS;
+        return AuthStatus.SEND_SUCCESS;
```
arjantijms, line 102: That one doesn't need to be overriden, it's already the default:

default AuthStatus secureResponse(MessageInfo messageInfo, Subject serviceSubject) throws AuthException {
        return AuthStatus.SEND_SUCCESS;
    }

PR #24878, opened 2024-03-26
Merge latest changes from master into 8.0

arjantijms:


dmatej:
@arjantijms is the failure of embedded tests OK for now or is it something new?
```
Caused by: java.lang.ClassNotFoundException: io.helidon.common.config.ConfigException
```

dmatej (approved):

dmatej:
I have found the cause and probably how to fix it, so I will push it into standalone PR.

PR #24879, opened 2024-03-27
Fixes #23483 Add lock, add notifyWaitingThreads call for better max pool size logic.

### LINKED ISSUE #23483: It didn't reach the limit of max-pool-size, but there was an maxPoolSize exhausted exception

escay:
Fixes #23483 Add lock logic between 2 methods, add notifyWaitingThreads call for better max pool size logic.

Add lock, add notifyWaitingThreads call for better max pool size logic. No longer getResources while the pool is being resized. And inform waiting threads when resources are marked as failed. 

See my analysis in #23483 for more information. Short version: getResource takes a connection from the pool, and resizes the pool when needed, while in the meantime another thread returns a resource to the pool, causing the resizePoolAndGetNewResource method to make the pool size too large. And in case of failing connections are being returned, the waiting threads are not informed, while they could ask for a fresh new connection from the pool.

Without the changes in ConnectionPool you should be able to reproduce the issues with the unit test. It might be needed to run the test several times, or increase the number of tasks in the test itself.

### DIFF
--- a/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/pool/ConnectionPool.java
+++ b/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/pool/ConnectionPool.java
@@ -44,6 +44,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 import java.util.Timer;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Logger;
 
 import javax.naming.NamingException;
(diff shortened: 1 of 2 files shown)


appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/pool/ConnectionPool.java
```diff
@@ -67,38 +68,136 @@ public class ConnectionPool implements ResourcePool, ConnectionLeakListener, Res
     private static final Logger LOG = LogDomains.getLogger(ConnectionPool.class, LogDomains.RSR_LOGGER);
 
     // pool life-cycle config properties
+    /**
+     * Represents the "max-wait-time-in-millis" configuration value.<br>
```
pzygielo, line 72: I guess it's about different config val.

escay: Fixed in commit [9208483](https://github.com/eclipse-ee4j/glassfish/pull/24879/commits/92084835a8cd6365a37910d96ee4d435e473e3ab)

dmatej (approved):

pzygielo (approved):

PR #24892, opened 2024-04-04
Update various versions in pom.xml to their latest in 8.0

arjantijms:
Includes the Jakarta EE 11 M2 wave


appserver/pom.xml
```diff
@@ -758,7 +758,14 @@
                 <plugin>
                     <groupId>org.glassfish.build</groupId>
                     <artifactId>spec-version-maven-plugin</artifactId>
-                    <version>2.1</version>
+                    <version>2.2</version>
```
dmatej, line 761: Just a question - which Maven version do you use?

arjantijms: 3.9.6

dmatej: Hmmm, so spec version plugin should change something. I met with an issue in glassfihsbuild maven plugin, where one dependency brought 4.0.0, then it did not work at all, because another required older version. Seems here comes another fragile season of changes as some plugins are preparing for Maven4 and break compatibility with 3.9.6.... and some just work even with newer dependencies.

PR #24916, opened 2024-04-16
Align JUnit5 version in REST CDI TCK with the nucleus parent

avpinchuk:
Synchronize JUnit version in REST CDI TCK with the nucleus parent.

### DIFF
--- a/appserver/tests/tck/rest_cdi/pom.xml
+++ b/appserver/tests/tck/rest_cdi/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 2023 Contributors to the Eclipse Foundation. All rights reserved.
+    Copyright (c) 2023, 2024 Contributors to the Eclipse Foundation.
 
     This program and the accompanying materials are made available under the
     terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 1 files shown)


appserver/itest-tools/pom.xml
```diff
@@ -65,6 +83,7 @@
         <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-engine</artifactId>
+            <scope>test</scope>
```
dmatej, line 86: Here it must be a compile scope, not a test scope. I wonder why it was not specified here.

nucleus/parent/pom.xml
```diff
@@ -678,16 +678,11 @@
             </dependency>
 
             <dependency>
-                <groupId>org.junit.jupiter</groupId>
-                <artifactId>junit-jupiter-engine</artifactId>
+                <groupId>org.junit</groupId>
```
dmatej, line 681: I don't like overusing BOMs, because they can cause serious dependency conflicts with other BOMs. Also now you had to declare scope everywhere. Does it have some real advantage? I don't see it.

avpinchuk: Thank you for review. Fixed. I simply align JUnit in TCK with nucleus parent.

dmatej (approved):

Issue #24887, opened 2024-04-02
Release and integrate updated dependencies


dmatej:
- Exousia 2.1.2 for GF7
- Exousia 3.0.0-M3 for GF8
- Epicyro 3.1.0 for GF8 - DONE
- glassfish-build-maven-plugin 4.0.2 for GF7+

escay:
Could glassfish-build-maven-plugin perhaps also be updated to 4.0.1 and integrated?
For small Windows build issue: https://github.com/eclipse-ee4j/glassfish-build-maven-plugin/issues/176

arjantijms:
Build plugin 4.0.1 staged: https://jakarta.oss.sonatype.org/content/repositories/staging/org/glassfish/build/glassfishbuild-maven-plugin/4.0.1/

arjantijms:
Unfortunately can't integrate it in GF (8) at the moment. The build fails when updating the spec plugin (to 2.2) or the build plugin (to the just staged 4.0.1)

```
[ERROR] Failed to execute goal org.glassfish.build:glassfishbuild-maven-plugin:4.0.1:featuresets-dependencies (default-featuresets-dependencies) on project glassfish-common: Execution default-featuresets-dependencies of goal org.glassfish.build:glassfishbuild-maven-plugin:4.0.1:featuresets-dependencies failed: A required class was missing while executing org.glassfish.build:glassfishbuild-maven-plugin:4.0.1:featuresets-dependencies: org/apache/commons/io/build/AbstractStreamBuilder
[ERROR] -----------------------------------------------------
[ERROR] realm =    extension>org.glassfish.build:glassfishbuild-maven-plugin:4.0.1
[ERROR] strategy = org.codehaus.plexus.classworlds.strategy.SelfFirstStrategy
[ERROR] urls[0] = file:/.m2/repository/org/glassfish/build/glassfishbuild-maven-plugin/4.0.1/glassfishbuild-maven-plugin-4.0.1.jar
[ERROR] urls[1] = file:/.m2/repository/org/apache/maven/maven-archiver/3.6.2/maven-archiver-3.6.2.jar
[ERROR] urls[2] = file:/.m2/repository/org/codehaus/plexus/plexus-archiver/4.9.2/plexus-archiver-4.9.2.jar
[ERROR] urls[3] = file:/.m2/repository/org/codehaus/plexus/plexus-utils/4.0.0/plexus-utils-4.0.0.jar
[...]

PR #24919, opened 2024-04-17
Code cleanups made when I was working on Exousia 2.1.2-SNAPSHOT

dmatej:
Should be compatible with 2.1.1 too, but if we would release Exousia 2.1.2 before this would be merged, I can update it yet bit more.

Relates to #24887


arjantijms:
See https://github.com/eclipse-ee4j/glassfish/pull/24923

avpinchuk (approved):

PR #24922, opened 2024-04-17
Fix embedded web test

dmatej:
* Quickly refactored the Server class (Eclipse IDE) - imports, generics
* EmbeddedServerUtils don't print redundant stack traces, don't depend on junit4
* Reimplemented embedded/web/servlet test
  * Original code ignored errors
  * Original code failed with NPE
  * Original code used internal API
  * Original code used java command from the OS, not from Maven


appserver/tests/embedded/web/servlet/src/test/java/org/glassfish/tests/embedded/web/servlet/Application.java
```diff
@@ -0,0 +1,86 @@
...
+            glassfish.dispose();
+        } catch (final GlassFishException e) {
+            throw new IllegalStateException(e);
+        }
+    }
+
+
+    public static Application start() throws IOException, GlassFishException {
+        GlassFishProperties props = new GlassFishProperties();
+        props.setPort("http-listener", 8882);
```
pzygielo, line 74: Could the port be passed to `start`? The parameter could be reused in l. 84 if `8882` there represents the same.

dmatej: I was so happy I fixed that so I forgot, thanks! We can even use any random free port, I pushed that right now.

avpinchuk (approved):

PR #24925, opened 2024-04-18
Revised Ant exec usages in TCK + TCK 10.0.4

dmatej:
- added AS_JAVA to force all asadmins to use the same JDK as Maven
- added JAVA_HOME to force all ant executions to use the same JDK as Maven
- added full path to java command (based on java.home property)
- fixed result processing in several cases
- added failonerror (false by default); not added just to executions where we don't  care about the result.
- passed all TCK tests

### DIFF
--- a/appserver/tests/tck/activation/pom.xml
+++ b/appserver/tests/tck/activation/pom.xml
@@ -148,11 +148,11 @@
 
 
                                 <!-- Change configuration -->
-                                
+
                                 <tck-setting key="TS_HOME" value="${tck.home}"/>
                                 <tck-setting key="JAVA_HOME" value="${java.home}"/>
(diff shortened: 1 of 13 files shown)


dmatej:
How to run the TCKs: https://github.com/eclipse-ee4j/glassfish/wiki/Jakarta-EE-10-TCK-Tests
If you want to run individual tests, you can do for example this:

```
mvn clean install -Ptck -pl :jakarta-pages-tags-tck
mvn clean install -Ptck -pl :glassfish-external-tck-pages-tags -Drun.test=com/sun/ts/tests/signaturetest -Dglassfish.version=7.0.14
```

pzygielo (approved):

avpinchuk (approved):

PR #24933, opened 2024-04-22
Classloader matching for BeanDeploymentArchives

lisa-lthorrold:
The same beanClass may be within several BeanDeploymentArchives. In it's current form, it returns the first matching BDA by checking if the beanClass is can be found within that BDA. If the incorrect BDA is returned, for example in:

https://github.com/eclipse-ee4j/glassfish/blob/5d59bb6a60301341125438bebe36aac35c0337ae/appserver/web/weld-integration/src/main/java/org/glassfish/weld/services/CDIServiceImpl.java#L418

then the classloader of the instantiated object in `createManagedObject` will belong (based on https://github.com/weld/core/blob/aab0ff03a84ea5718a515e9193a64b190e8f8479/impl/src/main/java/org/jboss/weld/manager/BeanManagerImpl.java#L1176) to the initiating class loader which first created the managed bean. This results in a `classcastexception` if subsequent requests for the same beanclass is coming from a different classloader (i.e. the classloader of the `managedClass` is disregarded).

The change here tightens the condition for returning the correct BDA, by matching it's module class loader to the incoming beanClass to ensure no `classcastexception` is thrown when the generated objected is cast back to beanClass.

### DIFF
--- a/appserver/web/weld-integration/src/main/java/org/glassfish/weld/DeploymentImpl.java
+++ b/appserver/web/weld-integration/src/main/java/org/glassfish/weld/DeploymentImpl.java
@@ -321,8 +321,9 @@ public class DeploymentImpl implements CDI11Deployment {
      * @param beanClass The beanClass to get the BeanDeploymentArchive for.
      *
      * @return If the beanClass is in the archive represented by the BeanDeploymentArchive then return that BeanDeploymentArchive.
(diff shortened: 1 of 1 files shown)


dmatej (approved):
Wow, I think you are right! Thank you!

avpinchuk (approved):

arjantijms:
I was a bit slow to see this PR, but really great work! Thanks from me too!

PR #24941, opened 2024-05-01
Fixes resources leak in `WebappClassLoader`

avpinchuk:
When we undeploy a web application that uses deploy time libraries, these libraries remain open until the domain stops/restart.

Now we disable deploy time libraries caching and close their associated class loaders during undeploy.

### DIFF
--- a/appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
+++ b/appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
@@ -24,6 +24,7 @@ import com.sun.enterprise.security.integration.PermsHolder;
 import com.sun.enterprise.util.io.FileUtils;
 
 import java.io.ByteArrayInputStream;
+import java.io.Closeable;
 import java.io.File;
 import java.io.FilePermission;
 import java.io.IOException;
@@ -1188,6 +1189,17 @@ public final class WebappClassLoader extends GlassfishUrlClassLoader
 
     public void preDestroy() {
         LOG.log(TRACE, "preDestroy()");
+
+        // Close deployment time libraries.
+        ClassLoader parent = getParent();
+        if (parent instanceof Closeable) {
+            try {
+                ((Closeable) parent).close();
+            } catch (IOException e) {
+                LOG.log(ERROR, "There were issues with closing deploy time libraries", e);
+            }
+        }
+
         try {
             close();
         } catch (Exception e) {
--- a/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/AppLibClassLoaderServiceImpl.java
+++ b/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/AppLibClassLoaderServiceImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Contributors to the Eclipse Foundation
+ * Copyright (c) 2023, 2024 Contributors to the Eclipse Foundation.
  * Copyright (c) 2007, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -64,8 +64,6 @@ public class AppLibClassLoaderServiceImpl {
     @Inject
     CommonClassLoaderServiceImpl commonCLS;
 
-    private final Map<URI, DelegatingClassLoader.ClassFinder> classFinderRegistry = new HashMap<>();
-
     /**
      * @see org.glassfish.internal.api.ClassLoaderHierarchy#getAppLibClassLoader(String, List<URI>)
      */
@@ -102,14 +100,8 @@ public class AppLibClassLoaderServiceImpl {
 
         ClassLoader commonCL = commonCLS.getCommonClassLoader();
         for (URI libURI : libURIs) {
-            synchronized (this) {
-                DelegatingClassLoader.ClassFinder libCF = classFinderRegistry.get(libURI);
-                if (libCF == null) {
-                    libCF = new URLClassFinder(new URL[]{libURI.toURL()}, commonCL);
-                    classFinderRegistry.put(libURI, libCF);
-                }
-                holder.addDelegate(libCF);
-            }
+            DelegatingClassLoader.ClassFinder libCF = new URLClassFinder(new URL[]{libURI.toURL()}, commonCL);
+            holder.addDelegate(libCF);
         }
     }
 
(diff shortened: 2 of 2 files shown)


appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
```diff
@@ -1188,6 +1189,17 @@ public String toString() {
 
     public void preDestroy() {
         LOG.log(TRACE, "preDestroy()");
+
+        // Close deployment time libraries.
+        ClassLoader parent = getParent();
```
pzygielo, line 1194: Could we extract new method, name it like `close deployment time libraries`, call it here, and avoid comment?

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/AppLibClassLoaderServiceImpl.java
```diff
@@ -129,6 +120,22 @@ public DelegatingClassLoader run() {
...
+
+        public AppLibClassLoader(ClassLoader parent) {
+            super(parent);
+        }
+
+        @Override
+        public void close() throws IOException {
+            for (ClassFinder classFinder : getDelegates()) {
+                if (classFinder instanceof URLClassFinder) {
+                    ((GlassfishUrlClassLoader) classFinder).close();
```
pzygielo, line 133: Could the types `instanceof`ed and cast be synchronized?

appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
dmatej: Yeah, I support this idea. Are we sure that those parent classloaders are not shared?

I always avoided using external dependencies from the file system. I have doubts it always caused issues. It can be in the domain, but then it is manager by the domain cl, or in ear/war, then it is bound to the app lifecycle. Otherwise... issues.

avpinchuk: > Could we extract new method, name it like `close deployment time libraries`, call it here, and avoid comment?

Certainly. Thank you for advice.

avpinchuk: > I always avoided using external dependencies from the file system. I have doubts it always caused issues. It can be in the domain, but then it is manager by the domain cl, or in ear/war, then it is bound to the app lifecycle. Otherwise... issues.

This is an attempt to solve test issue on Windows (again ;( ). This line failed:

https://github.com/eclipse-ee4j/glassfish/blob/6a77e1cf5f69666837d487232b1c132c4f175d53/appserver/tests/application/src/test/java/org/glassfish/main/test/app/security/jmac/http/servlet/basic/HttpServletBasicAuthTest.java#L111

when trying to delete auth module file.

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/AppLibClassLoaderServiceImpl.java
avpinchuk: I don't think it's necessary. Instead, delegates list should be properly synchronized.

pzygielo: by _synchronized_ I meant: to use one type, like `GlassfishUrlClassLoader` (not like MT-`synchronized`)

avpinchuk: I apologize for the misunderstanding. No, we cannot. Delegates list may contain 2 types of `ClassFinder's` and both of them are `Closeable` (one extends `GlassFishClassLoader` directry, second - indrectly (extends `ASURLClassLoader`)). The only thing we can do - cast to `URLClassLoader` or even to `Closeable` instead of `GlassFishUrlClassLoader`.

pzygielo: Using `T := URLClassFinder`:
```java
                 if (classFinder instanceof URLClassFinder) {
                    ((URLClassFinder) classFinder).close();
                 }
```
I can't see what's wrong with that.

avpinchuk: Hmm... Yo're right)) Thanks.

dmatej:
But then what is the purpose of those libraries? If they could be shared between web applications, now the class loader will be created again for every war, right? Then they can be included directly to the war file right?

Then I understand why the test failed on windows - we don't have any way to unload these libraries. Perhaps it could be done automatically when we don't have any application classloader referring their classloader. Do I understand it well now? The last time I was reviewing on the cell phone, sorry ;-)

avpinchuk:
> But then what is the purpose of those libraries?

E.g. in failed tests ServerAuthModule installed with `--libraries` option.



> If they could be shared between web applications, now the class loader will be created again for every war, right? Then they can be included directly to the war file right?

Before their `ClassFinder`s shared, but we should deploy every app with the `--libraries` option. Those libraries not visible by default to others app. Now every app has its own `ClassFinder` 

> Then I understand why the test failed on windows - we don't have any way to unload these libraries. Perhaps it could be done automatically when we don't have any application classloader referring their classloader. Do I understand it well now?

You understand correctly ) But they are never released automatically, even if we undeploy an app. 

This service has singleton scope:

https://github.com/eclipse-ee4j/glassfish/blob/f256659a33a8444aefd6e8ffb7d9dec1b369d6f8/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/AppLibClassLoaderServiceImpl.java#L53

This map

https://github.com/eclipse-ee4j/glassfish/blob/f256659a33a8444aefd6e8ffb7d9dec1b369d6f8/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/AppLibClassLoaderServiceImpl.java#L67

never cleaned and contains `URLClassLoader`s and we never invoke their `close` method. Thus, app lib classloaders stay open.

avpinchuk:
I'm working on another option now. Stateless class loader which doesn't store open JAR file but open and close it on demand when we need lo load a class. This approach fixes errors in all places, where app libs used, not only in web app.

avpinchuk:
Resolved in #24946.

PR #24944, opened 2024-05-05
Move some info logged by the WebApp CL as INFO to DEBUG/FINE level to reduce noise

OndroMih:
- NOTE: This PR includes #24943, otherwise Admin Console failed to start.

With this, the following is logged in INFO during app deployment:

```
[#|2024-05-05T05:23:24.750875+02:00|INFO|GF 7.0.15-SNAPSHOT|org.glassfish.web.loader.WebappClassLoader|_ThreadID=300;_ThreadName=RunLevelControllerThread-1714879404711;_LevelValue=800;|
  close(), WebappClassLoader classloader for application: myapp|#]
```

The following is not logged as INFO anymore, rather as DEBUG/FINE. As INFO is the default log level, it's a noise in logs which is mostly for troubleshooting and it's better to have it in DEBUG level, only if requested:

```
[2024-05-05T05:10:08.836483+02:00] [GF 7.0.15-SNAPSHOT] [FINE] [] [org.glassfish.web.loader.WebappClassLoader] [tid: _ThreadID=64 _ThreadName=admin-listener(4)] [levelValue: 500] [[
  close(), this:
org.glassfish.web.loader.WebappClassLoader@44694e1c[name=unknown], urls=[
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/applications/myapp/WEB-INF/classes/
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/applications/myapp/WEB-INF/lib/byte-buddy-1.12.18.jar
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/applications/myapp/WEB-INF/lib/jakarta.activation-api-2.1.0.jar
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/applications/myapp/WEB-INF/lib/lombok-1.18.28.jar
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/applications/myapp/WEB-INF/lib/antlr4-runtime-4.10.1.jar
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/applications/myapp/WEB-INF/lib/jboss-logging-3.5.0.Final.jar
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/applications/myapp/WEB-INF/lib/jaxb-runtime-4.0.2.jar
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/applications/myapp/WEB-INF/lib/jakarta.transaction-api-2.0.1.jar
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/applications/myapp/WEB-INF/lib/commons-logging-1.2.jar
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/applications/myapp/WEB-INF/lib/txw2-4.0.2.jar
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/applications/myapp/WEB-INF/lib/classmate-1.5.1.jar
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/applications/myapp/WEB-INF/lib/hibernate-core-6.2.6.Final.jar
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/applications/myapp/WEB-INF/lib/jaxb-core-4.0.2.jar
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/applications/myapp/WEB-INF/lib/commons-dbcp2-2.9.0.jar
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/applications/myapp/WEB-INF/lib/h2-2.1.214.jar
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/applications/myapp/WEB-INF/lib/jakarta.inject-api-2.0.1.jar
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/applications/myapp/WEB-INF/lib/jakarta.xml.bind-api-4.0.0.jar
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/applications/myapp/WEB-INF/lib/angus-activation-2.0.0.jar
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/applications/myapp/WEB-INF/lib/hibernate-commons-annotations-6.0.6.Final.jar
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/applications/myapp/WEB-INF/lib/jakarta.persistence-api-3.1.0.jar
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/applications/myapp/WEB-INF/lib/istack-commons-runtime-4.1.1.jar
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/applications/myapp/WEB-INF/lib/commons-pool2-2.10.0.jar
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/applications/myapp/WEB-INF/lib/jandex-3.0.5.jar
file:/glassfish/appserver/distributions/glassfish/target/stage/glassfish7/glassfish/domains/domain1/generated/ejb/myapp/
][webAppName=myapp, delegate=true, context=unknown, status=RUNNING, antiJARLocking=false, securityManager=false, packageDefinitionSecurityEnabled=false, repositories=RepositoryManager[WEB-INF/classes/], notFound.size=2, pathTimestamps.size=24, resourceEntryCache.size=5]]]
```

### DIFF
--- a/appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
+++ b/appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
@@ -166,6 +166,8 @@ public final class WebappClassLoader extends GlassfishUrlClassLoader
     /** The list of not found resources to avoid slow repeated searches. */
     private final Set<String> notFoundResources = ConcurrentHashMap.newKeySet();
 
+    /* Name of the application this class loader is for */
+    private String webappName;
 
     /** Associated directory context giving access to the resources in this webapp. */
     private DirContext jndiResources;
@@ -267,7 +269,12 @@ public final class WebappClassLoader extends GlassfishUrlClassLoader
      * but no defined repositories.
      */
     public WebappClassLoader(ClassLoader parent) {
+        this(parent, "unknown");
+    }
+
+    public WebappClassLoader(ClassLoader parent, String webappName) {
         super(new URL[0], parent);
+        this.webappName = webappName;
         this.cleaner = new ReferenceCleaner(this);
         this.system = WebappClassLoader.class.getClassLoader();
         if (SECURITY_MANAGER != null) {
@@ -1131,7 +1138,9 @@ public final class WebappClassLoader extends GlassfishUrlClassLoader
         if (status == LifeCycleStatus.CLOSED) {
             return;
         }
-        LOG.log(INFO, "close(), this:\n{0}", this);
+        LOG.log(INFO, () -> "close(), " + this.getClass().getSimpleName()
+                + " classloader for application: " + webappName);
+        LOG.log(DEBUG, () -> "close(), this:\n" + this);
 
         cleaner.clearReferences(clearReferencesStatic ? resourceEntryCache.values() : null);
         status = LifeCycleStatus.CLOSED;
@@ -1171,7 +1180,8 @@ public final class WebappClassLoader extends GlassfishUrlClassLoader
     public String toString() {
         StringBuilder sb = new StringBuilder(4096);
         sb.append(super.toString());
-        sb.append("[delegate=").append(delegate);
+        sb.append("[webAppName=").append(webappName);
+        sb.append(", delegate=").append(delegate);
         sb.append(", context=").append(contextName);
         sb.append(", status=").append(status);
         sb.append(", antiJARLocking=").append(antiJARLocking);
--- a/appserver/web/war-util/src/main/java/com/sun/enterprise/glassfish/web/WarHandler.java
+++ b/appserver/web/war-util/src/main/java/com/sun/enterprise/glassfish/web/WarHandler.java
@@ -67,6 +67,8 @@ import static javax.xml.stream.XMLStreamConstants.END_DOCUMENT;
 import static javax.xml.stream.XMLStreamConstants.END_ELEMENT;
 import static javax.xml.stream.XMLStreamConstants.START_ELEMENT;
 
+import org.glassfish.api.deployment.OpsParams;
+
 /**
  * Implementation of the ArchiveHandler for war files.
  *
@@ -125,7 +127,8 @@ public class WarHandler extends AbstractArchiveHandler {
 
     @Override
     public ClassLoader getClassLoader(final ClassLoader parent, DeploymentContext context) {
-        PrivilegedAction<WebappClassLoader> action = () -> new WebappClassLoader(parent);
+        String appName = context.getCommandParameters(OpsParams.class).name();
+        PrivilegedAction<WebappClassLoader> action = () -> new WebappClassLoader(parent, appName);
         WebappClassLoader cloader = AccessController.doPrivileged(action);
         try {
             WebDirContext webDirContext = new WebDirContext();
--- a/nucleus/parent/pom.xml
+++ b/nucleus/parent/pom.xml
@@ -144,7 +144,7 @@
         <antlr.version>2.7.8</antlr.version>
         <ant.version>1.10.12</ant.version>
         <ant.external.version>1.10.2</ant.external.version>
-        <jackson.version>2.16.1</jackson.version>
+        <jackson.version>2.17.1</jackson.version>
         <fasterxml.classmate.version>1.7.0</fasterxml.classmate.version>
         <stax-api.version>1.0-2</stax-api.version>
         <jettison.version>1.5.4</jettison.version>


appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
```diff
@@ -267,7 +269,12 @@ public final class WebappClassLoader extends GlassfishUrlClassLoader
      * but no defined repositories.
      */
     public WebappClassLoader(ClassLoader parent) {
+        this(parent, "unknown");
```
dmatej, line 272: I am against this  change, it is confusing and doesn't bring a value.

appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
```diff
@@ -1131,7 +1138,9 @@ public void close() throws IOException {
         if (status == LifeCycleStatus.CLOSED) {
             return;
         }
-        LOG.log(INFO, "close(), this:\n{0}", this);
+        LOG.log(INFO, () -> "close(), " + this.getClass().getSimpleName()
```
dmatej, line 1141: The original state was very useful for quite regular issues of users who used incompatible dependencies packed directly to war. And it doesn't log too often, just on CL close.

OndroMih: It logs one every app deployment. To me, the big INFO message with urls is just noise. INFO messages to me are short statements about what's going on, with a few simple values. Not something that takes the whole screen to show the whole message. That's either CONFIG (which isn't supported by System.Logger), or DEBUG, because it's only relevant in case of issues. In the end, the urls list only shows what's inside `WEB-INF/lib` in the WAR, people can unpack the WAR and see.

appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
OndroMih: You mean it doesn't bring any value that the log message says which application is related to the web app classloader? I thought it's very important, otherwise the new INFO message just says that some web app classloader was closed.

OndroMih: Alternatively, I can just change the log level of the original message rom INFO to DEBUG. I don't think it's important to log an INFO message when classloader is closed, what do you think?

dmatej: I mean that it logs the value just in one case of five possible creations of the classloader. If it should know the name of the application, it should know it always, without any "unknown" value.

appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
dmatej: Knowing simple name of the closed classloader class is worthless. Currently it reports some statistics and also the list of managed libraries, which is very useful in many use cases (missing/duplicit/harmful dependencies, etc). I use it nearly everyday, with some issues detected which were not resolved yet.
We can discuss that with others - if I am the only one who uses it or not.

Btw it logs also the context which is also used for the name of the classloader - and has the same issue with "unknown" value - I would like rather to clear it and make it final rather than increasing number of ways to make it "weird".

OndroMih: I admit I was just too lazy to look at other places that use the constructor. It's better it all places pass the app name so that it can be in the log.

But I would still turn this log into the DEBUG level. If you think it's essential for development, then we can make the default log configuration set the log level of the `org.glassfish.web.loader.WebappClassLoader` logger to FINE. But for production, it makes more sense to disable this log with the NORMAL level. Would you agree with that?

arjantijms: > just on CL close.

Maybe we need to look a bit at the wording of the log message too.

I mean, it now sounds like a command.

"Close this: [list]"

Which may make the user wonder if they need to close something explicitly, but they have no idea how to do it.

dmatej: Basically the important information should be always logged:

* Name and all endpoints of the application
* Classpaths 
* When components changed their state, are usable or when they were disabled and why

This reports the classpath of the application. Maybe it would be better to report it on start rather than on close, but because the start is bit more complicated, this is now better.

OndroMih:
I'll close this PR. I see there's no consensus and this change is not important enough to seek for an agreement. I'm OK with logging the full classpath, although I still think it's more a debugging info and it's a noise in normal usage.

PR #24946, opened 2024-05-06
Improve application-specific libraries loading

avpinchuk:
When Jakarta EE applications and modules use shared framework classes (such as utility classes and libraries), the classes can be put in the path for an application-specific class loader rather than in an application or module.

Application libraries are included in the *AppLib* class loader.

If multiple applications or modules refer to the same libraries, classes in those libraries are automatically shared.

Since these libraries are external to applications and the GlassFish server itself, they can be deleted or replaced with new versions using operating system tools. The result is system dependent.

It is necessary to make the behavior of application libraries the same on all supported platforms.

The proposed solution attempts to achive this by creating temporary snapshots of the application libraries during application startup.

An application library is identified by its `URI` and some set of basic file attributes (`size`, `lastModifiedTime` and `fileKey`). If a library with the same URI and attributes is already loaded, use it. If not, load it and put in the registry of loaded application libraries for reuse by other applications.

First we try to create a temporary snapshot of the application library:

1. Open the `FileInputStream`, get the native open file descriptor from it and read its basic file attributes.
2. Reading data from the open input stream into temporary file.
3. Closing the input stream.

Performing these steps ensures that the file attributes match exactly the library content. This gives us some degree of atomicity.

A discrepancy is possible, for example, in the following scenario:

1. Read basic file attributes.
2. Replacing a library with an externa process.
3. Create a temporary snapshot of a library

In this case basic fie attributes will not match the library.

If we are unable to create a temporary snapshot of the library using native descriptor, we fall back to using standard NIO.2 functions.

If we cannot create a snapshot of the library for some reason, we fall back to using the original library.

Last but not least, we close application libraries class loaders and delete snapshots when the GlassFish server is stopped.

### DIFF
--- a/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/AppLibClassLoaderServiceImpl.java
+++ b/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/AppLibClassLoaderServiceImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Contributors to the Eclipse Foundation
+ * Copyright (c) 2023, 2024 Contributors to the Eclipse Foundation.
  * Copyright (c) 2007, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -17,10 +17,25 @@
 
 package com.sun.enterprise.v3.server;
 
+import com.sun.enterprise.util.OS;
+import com.sun.enterprise.util.io.FileUtils;
+
+import jakarta.inject.Inject;
+import jakarta.inject.Singleton;
+
+import java.io.File;
+import java.io.FileDescriptor;
+import java.io.FileInputStream;
 import java.io.IOException;
+import java.io.InputStream;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.attribute.BasicFileAttributes;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.Collection;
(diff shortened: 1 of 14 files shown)


avpinchuk:
@dmatej, what do you think about the concept itself?

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/AppLibClassLoaderServiceImpl.java
```diff
@@ -129,6 +140,55 @@ public DelegatingClassLoader run() {
...
+            super(parent);
+            this.libUri = libUri;
+        }
+
+        @Override
+        public Class<?> findClass(String name) throws ClassNotFoundException {
+            try (JarFile jarFile = new JarFile(new File(libUri), false, ZipFile.OPEN_READ, Runtime.version())) {
+                String path = name.replace('.', '/').concat(".class");
+                JarEntry jarEntry = jarFile.getJarEntry(path);
+                if (jarEntry != null) {
```
dmatej, line 157: I would check it for null (inverted if) and throw CNFE, then you have one inner block level less.

dmatej: Another issue might be that going to the file system for every class, opening the file and unpacking the entry we want might not be effective ... but ... the original code does that too, am I right? And it probably has to ... because in theory there might be overlapping packages, resources, etc ... 

Btw do we really always want to return null for resources?
Also I am now not sure how it is with Java Modules, I am always forgetting these things ... we should start writing tests using these features.

Maybe add some logs so we could watch the sequence.

avpinchuk: > Another issue might be that going to the file system for every class, opening the file and unpacking the entry we want might not be effective

There will be overhead when opening and closing a file. But this will get rid of the problem of blocking the file at the filesystem level on the OSes with mandatory file locking.

> because in theory there might be overlapping packages, resources, etc ...



> Btw do we really always want to return null for resources?

This is a skeleton now. All methods will be implemented, also many other checks will be added. )

> Also I am now not sure how it is with Java Modules

ClassFinder`s doesnt't deal with Java Module System at this time. This ability should be implemented separately.

avpinchuk:
Main part is done. Only minor changes need to be made. What you think, @dmatej?

dmatej:
> Main part is done. Only minor changes need to be made. What you think, @dmatej?

I will try before the end of the week, now I am struggling with some network issues and I am several days behind my email notifications ...

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/AppLibClassLoaderServiceImpl.java
```diff
@@ -17,165 +17,547 @@
...
+        /**
+         * Represents a field to access to the native descriptor for an open file.
+         */
+        private static final Field nativeDescriptorField;
+
+        static {
+            Method method;
+            Field field;
+            try {
+                if (OS.isWindows()) {
```
dmatej, line 341: I am not sure now, but isn't there already a way how to get these with NIO? Maybe just with newer JDK versions, I am not sure now. I have to check it later ...

avpinchuk: > Isn't there already a way how to get these with NIO? Maybe just with newer JDK versions, I am not sure now. I have to check it later ...

I just need to add logging and do some code cleanup... After that, I will add a detailed description of why the decision was made to use certain platform-dependent means.

avpinchuk:
Tested on Ubuntu 22.04.4 LTS, Windows 10 22H2 and macOS 14.5.

hs536 (commented):
LGTM. That snapshots is needed to troubleshoot app problems, but I'm not sure if the system's temp directory is the best place for it.

hs536:
> Any suggestions?

Humm... there are several candidates...
* system tmp dir (now)
* domain/node root dir: `/glassfish/domains/domain1(/snapshots?)`
* domain/node lib dir: `/glassfish/domains/domain1/lib(/applibs? /snapshots?)`
* applications dir: `ls -ltr /glassfish/domains/domain1/applications(/snapshots? /__internal/snapshots?)`

The `/glassfish/domains/domain1/lib/applibs` directory may not be appropriate since it is where users install external jars.
How about creating a new directory for the snapshots in the same hierarchy as applibs? 
-  `/glassfish/domains/domain1/lib/snapshots`

arjantijms:
> > Any suggestions?
> 
> Humm... there are several candidates...
> 
> * system tmp dir (now)
> * domain/node root dir: `/glassfish/domains/domain1(/snapshots?)`
> * domain/node lib dir: `/glassfish/domains/domain1/lib(/applibs? /snapshots?)`
> * applications dir: `ls -ltr /glassfish/domains/domain1/applications(/snapshots? /__internal/snapshots?)`
> 
> The `/glassfish/domains/domain1/lib/applibs` directory may not be appropriate since it is where users install external jars. How about creating a new directory for the snapshots in the same hierarchy as applibs?
> 
> * `/glassfish/domains/domain1/lib/snapshots`

There's cons and pros to system tmp and domain/[somewhere]. 

System tmp gets cleaned automatically, at some point. The con is that the connection between the usage of the tmp folder and the using application is less clear. Lots of other tmp stuff is put inside the GF folders now, like e.g. all the generated java code for Pages (JSP).

arjantijms (approved):

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/AppLibClassLoaderServiceImpl.java
dmatej: The problem is that it may change or even be abandoned in the future. How critical is the difference from this? https://www.baeldung.com/java-nio2-file-attribute

Is this the whole explanation?

> If we are unable to create a temporary snapshot of the library using native descriptor, we fall back to using standard NIO.2 functions.

dmatej:
To the temporary dir - on linux user can use soft link and link it to the operating system. However I don't remember how it would affect replications in cluster - and also if it should affect them.

avpinchuk:
> To the temporary dir - on linux user can use soft link and link it to the operating system. However I don't remember how it would affect replications in cluster - and also if it should affect them.

Snapshots are excluded from replication. Replication of an original libraries left unchanged.

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/AppLibClassLoaderServiceImpl.java
avpinchuk: > The problem is that it may change or even be abandoned in the future. How critical is the difference from this? https://www.baeldung.com/java-nio2-file-attribute

This prevents inconsistency between file attributes and library content in case where library has been replaced with new one after the attributes were read, but before the snapshot was created.

First, I open `FileInputStream`. This block the file from being replaced or deleted on Windows and saves the contents of the file (inode?) on UNIX until the input stream closes. Then I get the open file descriptor (handle on Windows) and read the attributes from it. Then I copy the open input stream to a temporary file and close input stream. This ensures that the attributes match the file exactly.

If for some reason it is impossible to read attributes using os-dependent functions, we return to using standard NIO.2 functions/

avpinchuk: > it may change or even be abandoned in the future

In this case snandard NIO.2 will be used. Maybe until another solution is found.

avpinchuk: I did test on Temurin 11-22 and some latest Semeru.

I also tried using simple tests how to get the attributes of open files on Windows, Linux, macOS and Solaris(OpenIndiana). For OpenJDK from 11 to 22 all works now.

arjantijms:
LGTM now. If anything needs to be enhanced still, let's do it in a follow-up PR.

dmatej:
> LGTM now. If anything needs to be enhanced still, let's do it in a follow-up PR.

You should wait a bit until we finish the discussion. Now I cannot add my approval and two years later I might argue I did not approve it :D

Btw all tests including TCK I have on my Jenkins passed two hours ago, so I would approve it now. So I do that at least as a comment.

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/AppLibClassLoaderServiceImpl.java
avpinchuk: > It still smells to me as some bad practice

Unfortunately, Java doesn't have methods `FileInputStream.getAttibutes()`, `FileChannel.getAttributes()`, `Files.readAttributes(FileDescriptor fd, ...)` or similar. Therefore we are forced to use the "reflection magic".


> lets try that, if somebody finds it problematic in practice, I hope he will report it

This should never fail. If "reflection magic" does not work, the standard path will:

```java
            // Fallback to the standard NIO.2 method
            if (attributes == null) {
                attributes = readAttributes(libURI);
            }
```

In `readAttributes() method`

```java
            return Files.readAttributes(Path.of(libURI), BasicFileAttributes.class);
```

PR #24948, opened 2024-05-07
Update various versions

arjantijms:


dmatej:
It would be better to do most of these upgrades on master, because when something moved again, it would create a conflict. Easy to resolve, but still conflict with takes a minute of someone's life ;-)

arjantijms:
Yeah, some of them could be in master, but these files have many changes so wonder how mergeable they still are anyway.

PR #24949, opened 2024-05-10
Performance test and fix for logger part of the logging system

dmatej:
Years ago I made some comparisons for JUL logging, but I never pushed those tests.
Yesterday I had some doubts about the best way to format log messages, so I created a test. Meanwhile I noticed that one method in logger was not overridden and when I overrode it, it started failing the test. Long story short, GlassFish usually uses multiple handlers in parallel, so GJULE pushes common parts to caller threads to avoid duplicating and even blocking on handlers and formatters. But that means that the early resolution should not do work which is not required later - especially it should not analyze stacktraces when the detected class and method is not used in any output.

Oh, the original idea - yeah, MessageFormat is really slow compared to "+", StringBuilder, even behind suppliers. String.format is quiet slow too.

### DIFF
--- a/nucleus/glassfish-jul-extension/src/main/java/org/glassfish/main/jul/cfg/GlassFishLoggingConstants.java
+++ b/nucleus/glassfish-jul-extension/src/main/java/org/glassfish/main/jul/cfg/GlassFishLoggingConstants.java
@@ -72,11 +72,12 @@ public class GlassFishLoggingConstants {
      * If this key is set to true, GJULE will detect the caller class and method from stacktrace,
      * which is quite expensive operation affecting logging throughput.
      * <p>
-     * If it is set to null, GJULE will not perform such detection.
+     * If it is set to false, GJULE will not perform such detection.
      * <p>
-     * If the property is not set, GJULE makes the decision based on known handlers
-     * (<code>*.printSource</code> value) - if any handler requires this feature, it is enabled.
-     * It is disabled otherwise or you can forcibly enable it by setting the value to true or false.
+     * If the property is not set, GJULE makes the decision based on the (<code>*.printSource</code>
+     * property) - if any formatter requires this feature, the feature is enabled.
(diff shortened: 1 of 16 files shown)


avpinchuk (approved):

nucleus/glassfish-jul-extension/src/main/java/org/glassfish/main/jul/cfg/GlassFishLoggingConstants.java
```diff
@@ -67,6 +68,23 @@ public class GlassFishLoggingConstants {
     /** If this key is set to true, GJULE will print really detailed tracing info to the standard output */
     public static final String KEY_TRACING_ENABLED = "org.glassfish.main.jul.tracing.enabled";
 
+    /**
+     * If this key is set to true, GJULE will detect the caller class and method from stacktrace,
+     * which is quite expensive operation affecting logging throughput.
+     * <p>
+     * If it is set to null, GJULE will not perform such detection.
```
pzygielo, line 75: I don't understand that. It says: _if set to **null**_ but the code seems to check for `false`.

I'm not very familiar with properties handling here, but _set to null_ is different case than _not set_, isn't it?

Just would like to know what are my options here :grin: and where the `false` from 
```java
        if (Boolean.FALSE.toString().equalsIgnoreCase(value)) {
            return false;
        }
```
fits in.

dmatej: You are right, I will take a sleep and I will read and fix it after myself :-)
Thanks for noticing that!
And I should add that to the documentation too.

dmatej:
Ok, yet once more to the Draft - missing documentation, and OneLineFormatter has printSource as true by default.

pzygielo (approved):

PR #24958, opened 2024-05-20
Updated TCK to Servlet 6.1

arjantijms:
(still fails several tests which we have to address)

### DIFF
--- a/appserver/tests/tck/servlet/pom.xml
+++ b/appserver/tests/tck/servlet/pom.xml
@@ -1,23 +1,26 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-
-    Copyright (c) 2021, 2022 Contributors to the Eclipse Foundation. All rights reserved.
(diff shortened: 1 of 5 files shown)


appserver/tests/tck/servlet/pom.xml
```diff
@@ -27,309 +30,171 @@
...
 
             <plugin>
                 <artifactId>maven-dependency-plugin</artifactId>
-                <configuration>
-                    <skip>${skipITs}</skip>
-                </configuration>
                 <executions>
                     <execution>
                         <id>unpack-glassfish</id>
-                        <phase>pre-integration-test</phase>
```
dmatej, line 105: I prefer using failsafe plugin (instead of surefire) and integration test lifecycle. It is more consistent for so huge things.

dmatej (approved):

avpinchuk (approved):

PR #24959, opened 2024-05-20
Fixed IllegalArgumentException + refactoring

dmatej:
* The fix is just this: https://github.com/eclipse-ee4j/glassfish/commit/3b4e132a81fd8b3a3de44b92f44146c6e38decd2
* When I tried to find the cause (log message with unresolved parameters) I touched some other things
  * renamed habitat in the naming manager
  * fixed few warnings in DOL
  * LogDomains printed warning about missing (and unused) resource bundle for the SecurityContainer
  * Originally I suspected JarArchive to make some silly exception when trying to resolve the file name. I was wrong, but I added two unit tests and did some trivial cleanup.
* These two commits were originally created in different branch, but I think they should go to master first.

### DIFF
--- a/appserver/common/glassfish-naming/src/main/java/com/sun/enterprise/naming/impl/GlassfishNamingManagerImpl.java
+++ b/appserver/common/glassfish-naming/src/main/java/com/sun/enterprise/naming/impl/GlassfishNamingManagerImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022, 2024 Contributors to the Eclipse Foundation
  * Copyright (c) 1997, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -83,7 +83,7 @@ public final class GlassfishNamingManagerImpl implements GlassfishNamingManager 
     private static final Logger LOG = System.getLogger(GlassfishNamingManagerImpl.class.getName());
 
     @Inject
-    private ServiceLocator habitat;
+    private ServiceLocator serviceLocator;
 
     private final InitialContext initialContext;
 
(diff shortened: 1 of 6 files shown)


appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/Application.java
```diff
@@ -1249,7 +1250,7 @@ public boolean isUniqueIdSet() {
      * @param id unique id for this application
      */
     public void setUniqueId(long id) {
-        LOG.log(Level.FINE, "[Application] " + getName() + " , uid: " + id);
+        LOG.log(Level.FINE, () -> "setUniquiId: [Application] " + getName() + " , uid: " + id);
```
pzygielo, line 1253: `setUniquiId` -> `setUniqueId`

pzygielo (approved):

avpinchuk (approved):

appserver/common/glassfish-naming/src/main/java/com/sun/enterprise/naming/impl/GlassfishNamingManagerImpl.java
```diff
@@ -684,7 +684,7 @@ private ArrayList<SimpleJndiName> listNames(SimpleJndiName name) throws NamingEx
     private String getComponentId() throws NamingException {
         final ComponentInvocation ci;
```
arjantijms, line 685: At some point (maybe not now) make these things "componentInvocation" instead of "ci", etc,

arjantijms (approved):

appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/Application.java
dmatej: Oh no, I did it again!

appserver/common/glassfish-naming/src/main/java/com/sun/enterprise/naming/impl/GlassfishNamingManagerImpl.java
dmatej: I agree, but then I would have to go through the whole module as this is "a convention" here if I remember right.
But I will take a look once again in the morning, I already made a typo.

PR #24954, opened 2024-05-15
Concurrency for GlassFish 8

dmatej:
Draft PR depending on work of @aubi , see https://github.com/eclipse-ee4j/glassfish-concurro/pull/85

Requires also new Concurrency TCK 3.1.1:
https://github.com/jakartaee/concurrency/issues/511

PR #24962, opened 2024-05-24
Backport of improvements in deployment from GF8 concurrency branch

dmatej:
Backported some cleanups from #24954 + much more rather small cleanups because originally it did not pass QL tests and I spent whole week investigating why it always breaks with every small change until I reached this safepoint.

* Added preInvoke before executing CDI extensions in WeldDeployer - some extensions need it.
* Added logging of ignored empty elements in SaxParserHandler
* Fixed few typos
* DeploymentException is now a standard exception - serialization of stacktraces was revisited in Java 9, so I expect that this is ok.
* RestExtension class was deleted, it is not required any more.
* ACLSingletonProvider - cleanup, revided comments, using ConcurrentHashMap instead of Hashtable
* Fixed NPE in EventListener (noticed once in logs)
* ComponentInvocation now has toString
* Added logs to WeldDeployer, - especially events are interesting
* CommonClassLoaderServiceImpl - removed attempts to load Derby from JDK (was possible in JDK6 which we don't support any more for a while)
* Replaced many `@Context` annotations with `@Inject`

### Note

* Disable whitespaces when reviewing, it will be more readable


appserver/web/weld-integration/src/main/java/org/glassfish/weld/ACLSingletonProvider.java
```diff
@@ -108,10 +103,27 @@ public T get(String id) {
...
+            LOG.log(DEBUG, () -> "set(id=" + id + ", object=" + object + ")");
+            store.put(getClassLoader(), object);
+        }
+
+        @Override
+        public void clear(String id) {
+            LOG.log(DEBUG, () -> "clear(id=" + id + ")");
+            store.remove(getClassLoader());
+        }
+
```
pzygielo, line 126: A bit strange that `id`s are completely ignored in those methods. Could the parameters be renamed to express this intention?

dmatej: Nope, the Singleton class is Weld's interface. The funny thing is that the id is somehow useful, however I doubt it is guaranteed that it is unique. Probably that is the reason why somebody uses the classloader:
`set(id=DeploymentDescriptorTests_web, object=org.jboss.weld.Container@3bf27b0)]]`

dmatej:
Don't merge it yet, I will yet need to extend the preInvoke-postInvoke range - I rebased the #24954  to this locally and it failed few tests with qualifiers which was possible to fix just by this change. However I still need to find out how large range is required.

appserver/web/weld-integration/src/main/java/org/glassfish/weld/ACLSingletonProvider.java
pzygielo: >> Could the parameters be renamed to express this intention?

> Nope, the Singleton class is Weld's interface. 

But Weld does not have to know what is done to the parameter name in this implementation.
I find it also a little disturbing that it's logged like `id` values matter, while they don't.

dmatej: So what is your idea? I don't like it too, it smells, I see in logs it is probably always the application name and maybe it could be used as a key instead of classloader, but ... I can 

* delete all these logs (but then I cannot read anything useful in logs)
* change `id=` to `application=` (but that can change one day without warning, it is rather guessing what it really is)
* switch from CL to the id (but that would be quite risky experiment now)
* keep it with grinding teeth (confusing as you wrote)
* add `", but using thread classloader as an id instead"`

I did this because it was the simplest way to get at least something.

pzygielo: Do not consider my comment to be blocking :) (But I've not reviewed everything yet.).

I was only thinking about change `id` -> `ignoredId` - and keep everything else as-is. Just for me, reading this somewhere in the future.

I'm for marking this one detail as `resolved`, and leave it on _to think about list_.

appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/util/DOLUtils.java
```diff
@@ -370,12 +370,12 @@ public static ArchiveType getModuleType(String moduleType) {
     // returns true if GF DD should have higher precedence over
     // WLS DD when both present in the same archive
     public static boolean isGFDDOverWLSDD() {
-        return Boolean.valueOf(System.getProperty(GFDD_OVER_WLSDD));
+        return Boolean.parseBoolean(System.getProperty(GFDD_OVER_WLSDD));
```
avpinchuk, line 373: Why not use `Boolean.getBoolean()` instead in all places?

appserver/web/weld-integration/src/main/java/org/glassfish/weld/ACLSingletonProvider.java
dmatej: Finally I added at least the comment to the log.

arjantijms (approved):

nucleus/common/internal-api/src/main/java/org/glassfish/internal/data/ApplicationInfo.java
```diff
@@ -112,12 +112,12 @@ private void createServiceLocator() {
 
         appServiceLocator = slf.create(locatorName);
         deploymentFailedListener = new DeploymentFailedListener(source);
-        events.register(deploymentFailedListener);
+        unregisterEventListener(deploymentFailedListener);
```
avpinchuk, line 115: Maybe we should **register** listener here instead of *unregister*?

avpinchuk (changes requested):

nucleus/common/internal-api/src/main/java/org/glassfish/internal/data/ApplicationInfo.java
dmatej: Do you have any idea why it did not break any test? Whole TCK passed, I am just locally running just few remaining which failed because of read timeouts caused by my slow network.
That is really stupid mistake I made. I do wonder how I did that ... thank you for noticing that!

avpinchuk: `DeploymentFailedListener` calls `disposeServiceLocator()` method. This method unregister the listener itself and destroy the application service locator in case of a deployment failure. I think, we just have a memory leak in this case.

avpinchuk (approved):

hs536 (approved):

PR #24964, opened 2024-05-26
Allow resource references in persistence xml datasource

OndroMih:
Allows to use a resource reference in persistence.xml, as in the provided test. This reference can be changed using resource-ref in `glassfish-web.xml` with alternative runtime descriptor or in a deployment plan during deployment.


appserver/tests/application/src/test/java/org/glassfish/main/test/app/synchronize/SynchronizeStandaloneInstanceTest.java
```diff
@@ -69,10 +71,8 @@ public void testSynchronization() {
 
     private static File createDeployment() {
         WebArchive webArchive = ShrinkWrap.create(WebArchive.class)
-            .addAsWebInfResource(
-                new File("src/main/resources/org/glassfish/main/test/app/synchronize/sample.json"),
-                "resources/sample.json")
-            .addAsWebResource(new File("src/main/resources/org/glassfish/main/test/app/synchronize/index.html"));
+            .addAsWebInfResource(getTestResourceStatic("sample.json"), "resources/sample.json")
```
dmatej, line 74: You don§t need to create a method for that, you can even shorten it to the resource reference. It is also better if you want to filter the file by maven. Perhaps just add json files to resources, like xml files are already there. It is always wrong to touch src/main/resources from the test in runtime (also Eclipse uses different basedir than Maven).

appserver/persistence/common/src/main/java/org/glassfish/persistence/common/PersistenceHelper.java
```diff
@@ -44,10 +45,25 @@ public static DataSource lookupPMResource(ConnectorRuntime connectorRuntime, Dep
...
     }
+
+    private static SimpleJndiName translateResourceReference(DeploymentContext ctx, SimpleJndiName dataSourceName) {
+        final BundleDescriptor currentBundle = DOLUtils.getCurrentBundleForContext(ctx);
+        if (currentBundle instanceof ResourceReferenceContainer) {
+            ResourceReferenceContainer referenceContainer = (ResourceReferenceContainer)currentBundle;
+            try {
+                return referenceContainer.getResourceReferenceByName(dataSourceName.toString()).getJndiName();
+            } catch (IllegalArgumentException e) {
+                return dataSourceName;
```
dmatej, line 63: It would be good to log the exception as a warning at least.

OndroMih: Not here. This is a valid condition - if the datasource name is a proper JNDI and not a reference, just use it without any change. It would only pollute logs if it was a warning. I can add a debug logging.

The `getResourceReferenceByName` gives an exception if resource not found. Ideally, I would rather use `findResourceReferenceByName`, which returns a `null`, but that one is `protected` and only in `WebBundleDescriptor`, so I rather used a method on the interface `ResourceReferenceContainer`.

appserver/tests/application/src/test/java/org/glassfish/main/test/app/synchronize/SynchronizeStandaloneInstanceTest.java
OndroMih: I think I got you. Tests should get resources from the classpath instead of from the source directory. I updated the code.

appserver/persistence/common/src/main/java/org/glassfish/persistence/common/PersistenceHelper.java
dmatej: Ok, debug/trace is good then. Sometimes I am hunting things around DOL and XML/annotations and any such log is useful.

dmatej:
Seems something is still missing: 
```
Expected: asadmin succeeded
     but: was <Command deploy failed.
remote failure: Error occurred during deployment: Exception while deploying the app [JtaDataSourceResourceRefTestWebApp] : Exception [EclipseLink-4002] (Eclipse Persistence Services - 4.0.2.v202306161219): org.eclipse.persistence.exceptions.DatabaseException
Internal Exception: java.sql.SQLException: Failed to obtain/create connection from connection pool [DerbyPool]. Reason: Connection creation failed for 1 times.
Error Code: 0. Please see server.log for more details.

>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
	at org.glassfish.main.test.app.persistence.resourceref.JtaDataSourceResourceRefTest.testDeploy(JtaDataSourceResourceRefTest.java:66)
```

OndroMih:
Hm, obviously, the Derby DB is not running, it's not started by the test suite. I'll try to set up the defalt datasource to point to an embedded database.

OndroMih:
Some EJB test failed in CI with no info about it. @dmatej , can you please have a look and see what's wrong with it?

OndroMih:
@arjantijms , I ran locally after merged from master, and I get 10 failed tests when I run `./runtests.sh ejb_group_1`

From the `test_results_junit.xml` file:

```
<testsuite tests="103" failures="10" errors="0" name="ejb_group_1">
```
I'll try to look what's happening.

It's weird that when I ran the test without merging master, all tests pass but there are only 110 tests.

The latest Jenkins build for this PR reports 111 tests:

```
<testsuite tests="111" failures="0" errors="1" name="ejb_group_1">
```
And the one that is failing is missing in the 110 tests when I run my branch.

So, my branch is based on master, which only has 110 tests. The latest Jenkins run for this PR has 111 tests. And now the master branch has 113 tests. Were any tests added to the `ejb_group_1` test suite in the past few weeks? Or why do I get different number of tests?

OndroMih:
I got more familiar with how the tests are set up, and then I found the culprit. In the appclient context, there's a null value for context where I didn't expect null on server.

OndroMih:
@dmatej , @arjantijms , please review.

arjantijms (approved):

arjantijms:
Looks good to me now!

PR #24973, opened 2024-05-27
Update docs of the set command

OndroMih:
To describe how multiple properties can be set with the same command, and how to delete a property.

### DIFF
--- a/docs/reference-manual/src/main/asciidoc/set.adoc
+++ b/docs/reference-manual/src/main/asciidoc/set.adoc
@@ -13,20 +13,19 @@ set
 set
 ---
 
-Sets the values of configurable attributes
+Sets or deletes the values of configurable attributes
 
 === Synopsis
 
 [source]
 ----
-asadmin [asadmin-options] set [--help]
-attribute-name=value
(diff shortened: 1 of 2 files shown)


docs/reference-manual/src/main/asciidoc/set.adoc
```diff
@@ -70,8 +69,10 @@ asadmin-options::
 
 === Operands
 
-attribute-name=value::
-  Identifies the full dotted name of the attribute name and its value.
+attribute-name=value [attribute-name=value]::
+  Identifies the full dotted name of the attribute name and its value. 
+  Multiple attribute=value pairs can be provided, separated by a space (,).
```
hs536, line 74: > by a space (,).

Does this indicate that a comma can replace a space?

OndroMih: This is my mistake, I'll fix it. I first thought that values should be separated by commas, but then realized it's by space. And I forgot to remove the comma in parenthesis.

hs536 (approved):

PR #24976, opened 2024-06-02
The catch is not required to stop GF instance

dmatej:
- There is already the shutdown hook registered few lines above even with more "radical" parameters (kill, force).

### DIFF
--- a/appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/GlassFishTestEnvironment.java
+++ b/appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/GlassFishTestEnvironment.java
@@ -82,13 +82,7 @@ public class GlassFishTestEnvironment {
             asadmin.withEnv("AS_START_TIMEOUT", Integer.toString(ASADMIN_START_DOMAIN_TIMEOUT - 5000));
         }
         // This is the absolutely first start - if it fails, all other starts will fail too.
-        try {
-            assertThat(asadmin.exec(ASADMIN_START_DOMAIN_TIMEOUT, "start-domain", "--debug"), asadminOK());
-        } catch (AssertionError e) {
-            // stop domain if the domain is running from a previous test
-            asadmin.exec(ASADMIN_START_DOMAIN_TIMEOUT, "stop-domain");
-            throw e;
-        }
+        assertThat(asadmin.exec(ASADMIN_START_DOMAIN_TIMEOUT, "start-domain", "--debug"), asadminOK());
     }
 
 


pzygielo:
- #24965

OndroMih:
Exactly as @avpinchuk wrote - if a test is killed, the server keeps running. This help to stop the server easily on the nect test run.

dmatej:
> But if we kill test JVM with `kill -9`, shutdown hooks will not be executed. And test GF server still running. Or not?

And final blocks would be reliably executed? I don't think so. For sure I can try ... give me a while ... 
Btw for the first time I found answer on LinkeIn. :-D
https://www.linkedin.com/pulse/finally-clause-java-vinay-avasthi/

> When the virtual machine is working normally, the finally is guaranteed to be called unless System.exit is called in the normal flow of the execution.
> When System.exit is called, since the function never returns, finally is not called.
> If the JVM is terminated through SIGKILL or through hardware reset, the finally is not called.

dmatej:
Ok, I did this:
```
    public static void main(String[] args) throws Exception {
        Runtime.getRuntime().addShutdownHook(new Thread() {

            @Override
            public void run() {
                System.out.println("Shutdown hook is called.");
            }
        });
        try {
            System.out.println("Sleeping 60 s ...");
            Thread.currentThread();
            Thread.sleep(60_000);
            System.out.println("Woke up.");
        } catch (InterruptedException e) {
            System.out.println("Interrupted.");
        } finally {
            System.out.println("FINALLY");
        }
}
```

First test: CTRL+C from Konsole - to my surprise the shutdown hook was called, but finally not. Same result has SIGTERM.
```
Sleeping 60 s ...
^CShutdown hook is called.
```

Second test: SIGKILL - both finally and shutdown hooks are ineffective
```
Sleeping 60 s ...
Killed
```

Third test - timeout - both finally and shutdown hook are executed.
```
Sleeping 60 s ...
Woke up.
FINALLY
Shutdown hook is called.
```

avpinchuk:
I thought about this too.

As I think the main idea was next:

```
mvn clean test

kill -9 ...
```

Here GF is running.

Next `mvn clean test` should stop GF. Hmm, why not restart?

But it doesn't work in this case, because `clean` goal will remove test GF and unpack new one (which doesn't contains pid file). The `stop-domain` command can't stop the running GF.

If we run the second `mvn test` (without `clean` goal), the server will stop, but won't start again (because server started in static block, not in tests itself).

avpinchuk (approved):

avpinchuk:
If we need this functionatity, we need a realiable way to identify the running staled test GF and kill it (if it's possible at all).

OndroMih:
@avpinchuk , there's no reliable way to kill GlassFish when you run `mvn clean test`. It's not sure whether the GlassFish instance was started by a previous test or it was started manually outside of the test and it's just a port conflict. 

This situation probably doesn't have a good solution. I created this solution to help me in my specific case and I thought it could help others when they debug tests and decide to kill the test and start it again. When I need it next time, I'll try another solution and maybe I'll find some better approach. For now, it's OK to merge this PR and remove my previous changes.

hs536 (approved):

PR #24980, opened 2024-06-05
Fixed deadlock when rolling the file under a heavy load

### LINKED ISSUE #24978: Detected deadlock in logging - rolling the file
Blocks whole server after a while because of this "recursion". I thought it is ok when I was rewriting that, but I forgot on one special case: the handler doesn't accept new records because it is rolling the file AND the buffer is full. Then the record saying that the file is rolled effectively blocks its own thread.

It is randomly reproducible when you enable FINEST level on some logger with high traffic (ie. CDI) that is able to fill the buffer. Any new message has to wait.

Possible solution: Do not use logging from log file manager until we are able to use the file again.

```
"GlassFishLogHandler log pump" #67 [112445] daemon prio=10 os_prio=0 cpu=3112,83ms elapsed=199,27s tid=0x00007fb614a92760 nid=112445 waiting on condition  [0x00007fb598bf2000]
java.lang.Thread.State: WAITING (parking)
at jdk.internal.misc.Unsafe.park(java.base@21.0.1/Native Method)
- parking to wait for  <0x00000000e14f2f08> (a java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject)
at java.util.concurrent.locks.LockSupport.park(java.base@21.0.1/LockSupport.java:371)
at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionNode.block(java.base@21.0.1/AbstractQueuedSynchronizer.java:519)
at java.util.concurrent.ForkJoinPool.unmanagedBlock(java.base@21.0.1/ForkJoinPool.java:3780)
at java.util.concurrent.ForkJoinPool.managedBlock(java.base@21.0.1/ForkJoinPool.java:3725)
at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(java.base@21.0.1/AbstractQueuedSynchronizer.java:1707)
at java.util.concurrent.ArrayBlockingQueue.put(java.base@21.0.1/ArrayBlockingQueue.java:370)
at org.glassfish.main.jul.handler.LogRecordBuffer.addWithUnlimitedWaiting(LogRecordBuffer.java:160)
at org.glassfish.main.jul.handler.LogRecordBuffer.add(LogRecordBuffer.java:120)
at org.glassfish.main.jul.handler.GlassFishLogHandler.publish(GlassFishLogHandler.java:234)
at org.glassfish.main.jul.GlassFishLogger.lambda$checkAndLog$7(GlassFishLogger.java:459)
at org.glassfish.main.jul.GlassFishLogger$$Lambda/0x00007fb5a01e7c18.accept(Unknown Source)
at java.util.Spliterators$1Adapter.forEachRemaining(java.base@21.0.1/Spliterators.java:704)
at org.glassfish.main.jul.GlassFishLogger.checkAndLog(GlassFishLogger.java:459)
at org.glassfish.main.jul.GlassFishLogger.logOrQueue(GlassFishLogger.java:502)
at org.glassfish.main.jul.GlassFishLogger.log(GlassFishLogger.java:197)
at sun.util.logging.internal.LoggingProviderImpl$JULWrapper.log(java.logging@21.0.1/LoggingProviderImpl.java:274)
at org.glassfish.main.jul.rotation.LogFileManager.roll(LogFileManager.java:160)
- locked <0x00000000ffab9d30> (a org.glassfish.main.jul.rotation.LogFileManager)
at org.glassfish.main.jul.rotation.LogFileManager.rollIfFileTooBig(LogFileManager.java:132)
- locked <0x00000000ffab9d30> (a org.glassfish.main.jul.rotation.LogFileManager)
at org.glassfish.main.jul.rotation.LogFileManager.flush(LogFileManager.java:114)
- locked <0x00000000ffab9d30> (a org.glassfish.main.jul.rotation.LogFileManager)
at org.glassfish.main.jul.handler.GlassFishLogHandler.flush(GlassFishLogHandler.java:249)
at org.glassfish.main.jul.handler.GlassFishLogHandler$LoggingPump.flushOutput(GlassFishLogHandler.java:480)
at org.glassfish.main.jul.handler.LoggingPumpThread.publishBatchFromBuffer(LoggingPumpThread.java:97)
at org.glassfish.main.jul.handler.LoggingPumpThread.run(LoggingPumpThread.java:73)
```

## Environment Details
* GlassFish Version (and build number): 7.0.15
* JDK version: 17

dmatej:
* Fixed LogRecordBuffer configuration - it always used defaults
  *  That caused overload without any possibility to configure maxWaitTime or increase the buffer size
* Fixed deadlock - don't use loggers in critical section.
  * Deadlock can be caused when we are rolling the file with the full log record buffer AND we log anything from this thread.
  * Solution: if you really need to log something, use asynchronous thread.
  * Under a heavy load (1 record every 3 microseconds) such log record can occur even after another 5000 other records in the buffer or trying to be added to the buffer. It is just another one. Usually it will be the first one.
* Moving from synchronized to reentrant locks + fixed deadlock

### DIFF
--- a/nucleus/glassfish-jul-extension/src/main/java/org/glassfish/main/jul/handler/LogRecordBuffer.java
+++ b/nucleus/glassfish-jul-extension/src/main/java/org/glassfish/main/jul/handler/LogRecordBuffer.java
@@ -59,7 +59,7 @@ class LogRecordBuffer {
      *
      * @param capacity capacity of the buffer.
      */
-    public LogRecordBuffer(final int capacity) {
+    LogRecordBuffer(final int capacity) {
         this(capacity, 0);
     }
 
@@ -84,7 +84,7 @@ class LogRecordBuffer {
      * @param maxWait maximal time in seconds to wait for the free capacity. If &lt; 1, can wait
      *            forever.
      */
-    public LogRecordBuffer(final int capacity, final int maxWait) {
+    LogRecordBuffer(final int capacity, final int maxWait) {
         this.capacity = capacity;
         this.maxWait = maxWait;
         this.pendingRecords = new ArrayBlockingQueue<>(capacity, true);
@@ -101,8 +101,13 @@ class LogRecordBuffer {
     // R/W locks - only this method locks everything, because it has to recreate the queue using
     // the original pending records.
     public void reconfigure(final int newCapacity, final int newMaxWait) {
-        if (this.capacity == newCapacity && this.maxWait == newMaxWait) {
-            return;
+        this.lock.readLock().lock();
+        try {
+            if (this.capacity == newCapacity && this.maxWait == newMaxWait) {
+                return;
+            }
+        } finally {
+            this.lock.readLock().unlock();
         }
         this.lock.writeLock().lock();
         try {
--- a/nucleus/glassfish-jul-extension/src/main/java/org/glassfish/main/jul/handler/LogRecordBuffer.java
+++ b/nucleus/glassfish-jul-extension/src/main/java/org/glassfish/main/jul/handler/LogRecordBuffer.java
@@ -18,18 +18,35 @@ package org.glassfish.main.jul.handler;
 
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.logging.Level;
 
 import org.glassfish.main.jul.record.GlassFishLogRecord;
 
 /**
+ * The buffer for log records.
+ * <p>
+ * If it is full and another record is comming to the buffer, the record will wait until the
+ * buffer would have a free capacity, but only for a maxWait seconds.
+ * <p>
+ * If the buffer would not have free capacity even after the maxWait time, the buffer will be
+ * automatically cleared, the incomming record will be lost and there will be a stacktrace in
+ * standard error output - but that may be redirected to JUL again, so this must be reliable.
+ * <ul>
+ * <li>After this error handling procedure the logging will be available again in full capacity
+ * but it's previous unprocessed log records would be lost.
+ * <li>If the maxWait is lower than 1, the calling thread would be blocked until some records would
+ * be processed. It may remain blocked forever.
+ * </ul>
+ *
  * @author David Matejcek
  */
 class LogRecordBuffer {
 
-    private final int capacity;
-    private final int maxWait;
-    private final ArrayBlockingQueue<GlassFishLogRecord> pendingRecords;
+    private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock(true);
+    private int capacity;
+    private int maxWait;
+    private ArrayBlockingQueue<GlassFishLogRecord> pendingRecords;
 
 
     /**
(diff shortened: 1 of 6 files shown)


arjantijms (approved):

nucleus/glassfish-jul-extension/src/main/java/org/glassfish/main/jul/handler/LogRecordBuffer.java
```diff
@@ -70,25 +87,70 @@ public LogRecordBuffer(final int capacity) {
...
+     * Reconfigures the buffer.
+     *
+     * @param newCapacity capacity of the buffer.
+     * @param newMaxWait maximal time in seconds to wait for the free capacity. If &lt; 1, can wait
+     *            forever.
+     */
+    // R/W locks - only this method locks everything, because it has to recreate the queue using
+    // the original pending records.
+    public void reconfigure(final int newCapacity, final int newMaxWait) {
+        if (this.capacity == newCapacity && this.maxWait == newMaxWait) {
```
avpinchuk, line 104: This check is not thread-safe...

dmatej: I am thinking why I did that this way ... and maybe it is ok - those values can be changed just by this method in the locked part. Then if those values were changed by first thread, if the second brings same values it just leaves. If it has another, it will wait on the lock until the first finishes.
If the second thread would have original values, then it can return or can continue, BUT the same situation is with moving lock before the condition because the order of thread execution is not defined. The second can be processed first.

The only disadvantage would be if both first and second thread would have same values and both would pass the check, the first would do the reconfiguration, and the second would do that redundantly again.

But yeah, I can move it up, the method is not called frequently so there is no need to hurry with fast way out. The second thread can wait for the lock first; also the reconfiguration is very quick here.

avpinchuk (approved):
The old code did not guarantee visibility of variables. Each new thread could see the old values (or mix from old and new values) and do reconfiguration.

This code is now safe, although somewhat slower :)

avpinchuk:
Hmmm, if we move check under write lock, we have deadlock...

dmatej:
Good lesson, I read the javadoc again, and also - I should never push without running tests of the module at least.

dmatej:
Well, now I passed tests, but I did some manual testing and I have another deadlock. Also I see too many logging pumps for STDOUT+STDERR, seems after every reconfiguration it creates another ...

nucleus/glassfish-jul-extension/src/main/java/org/glassfish/main/jul/handler/LogRecordBuffer.java
```diff
@@ -43,10 +45,12 @@
  */
 class LogRecordBuffer {
 
-    private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock(true);
-    private int capacity;
-    private int maxWait;
-    private ArrayBlockingQueue<GlassFishLogRecord> pendingRecords;
+    private final BlockingQueue<GlassFishLogRecord> pendingRecords = new LinkedBlockingQueue<>();
```
dmatej, line 48: Doesn't have linked blocking queue significantly lower performance than ArrayBlockingQueue? I have to try ...

dmatej: Seems it is the opposite, hmmm!

> It has a higher throughput than array-based queues.

nucleus/glassfish-jul-extension/src/main/java/org/glassfish/main/jul/handler/LogRecordBuffer.java
```diff
@@ -128,36 +199,60 @@ public void add(final GlassFishLogRecord record) {
...
-        // note: the record is not meaningful for the message. The cause is in another place.
-        this.pendingRecords.offer(new GlassFishLogRecord(Level.SEVERE, //
-            this + ": The buffer was forcibly cleared after " + maxWait + " s timeout for adding another log record." //
-                + " Log records were lost." //
-                + " It might be caused by a recursive deadlock," //
-                + " you can increase the capacity or the timeout to avoid this.", false));
+        lock.lock();
+        try {
+            try {
+                if (availableCapacity.tryAcquire(0, TimeUnit.SECONDS)) {
```
dmatej, line 213: Why this attempt?

avpinchuk: > Why this attempt?

Anothe thread may have already drops all waiting records and we have space in the buffer. Some minor optimization.

avpinchuk (approved):

PR #24984, opened 2024-06-08
Context root doesn't include a random string

OndroMih:
For apps deployed via Admin Console, the uploaded file was copied to a temp file,  which contained a random string. Now, the temp directory has a random string,  but the file in it has the same name.
Context root is generated from the file name and doesn't include any random string with this commit.

Before, when an app was deployed with default values:
image

Now:
image

### DIFF
--- a/appserver/admingui/core/src/main/java/org/glassfish/admingui/handlers/WoodstockHandler.java
+++ b/appserver/admingui/core/src/main/java/org/glassfish/admingui/handlers/WoodstockHandler.java
@@ -47,7 +47,6 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Map;
-import java.security.SecureRandom;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


hs536 (approved):

appserver/admingui/core/src/main/java/org/glassfish/admingui/handlers/WoodstockHandler.java
```diff
@@ -193,6 +185,16 @@ public static void uploadFileToTempDir(HandlerContext handlerCtx) {
         handlerCtx.setOutputValue("uploadedTempFile", uploadTmpFile);
     }
 
+    private static File createTempFileWithOriginalFileName(String name) throws IOException {
+        final Path dir = Files.createTempDirectory(name);
+        dir.toFile().deleteOnExit();
```
dmatej, line 190: Shouldn't we delete the directory too? 
Those deleteOnExit usages will be reported by SonarQube because deleteOnExit is done just by shutdown hooks and is not something to rely on. Maybe we should target it later. Also ignoring file.delete() result can be reported by SpotBugs (which is not your code, it was already there).

- https://rules.sonarsource.com/java/type/Code%20Smell/RSPEC-2308/?search=deleteOn

dmatej (approved):

appserver/admingui/core/src/main/java/org/glassfish/admingui/handlers/WoodstockHandler.java
OndroMih: I updated the code:
* directory is also removed in the `deleteFileFromTempDir` handler
* `Files.deleteIfExists` used instead of `file.delete()`, shouldn't be reported by SpotBugs
* On IO exception, the temp files are removed if possible and an error is reported in GUI - I don't know why, but before it continued with a file, which might have been corrupted or empty

appserver/admingui/core/src/main/java/org/glassfish/admingui/handlers/WoodstockHandler.java
```diff
@@ -178,11 +171,13 @@ public static void uploadFileToTempDir(HandlerContext handlerCtx) {
             } catch (IOException ioex) {
                 try {
                     if (tmpFile != null) {
-                        uploadTmpFile = tmpFile.getAbsolutePath();
+                        Files.deleteIfExists(tmpFile.toPath());
+                        Files.deleteIfExists(tmpFile.toPath().getParent());
                     }
                 } catch (Exception ex) {
-                    //Handle AbsolutePathException here
+                    // ignore nested exception, handle the original exception only
```
dmatej, line 178: You can use ioex.addSuppressed here.

PR #24985, opened 2024-06-10
Dev profile to update modules without building the whole GlassFish distribution

OndroMih:
This will allow to build a single module and quickly update the GlassFish distribution in `appserver/distributions/glassfish/target/stage/glassfish7/glassfish` so that GlassFish can be started quickly with the code changes applied.

To apply changes in a specific module that suppots the `dev` profile:

* stop GlassFish if running
* cd /path/to/module
* `mvn -Pdev install`
* start GlassFish in `appserver/distributions/glassfish/target/stage/glassfish7/glassfish`

GlassFish distribution module must be first built by building the GlassFish project as usual.

To add support for the `dev` profile in other modules:

* in most cases, just set `copy.modules.to.distribution.skip` property to `false` (if the module is 3 nested directories below root, i.e. the root directory is at `../../..`)
* if the root directory is at a different location, set `copy.modules.to.distribution.path.to.root` property to the root directory (e.g. `copy.modules.to.distribution.path.to.root=../../../..`). An example is in the `nucleus/admin/rest/rest-service/pom.xml` file.
* if the destination file in GlassFish installation is different from `modules/${project.build.finalName}.jar`, then set `copy.modules.to.distribution.destFile` to point to the destination file, e.g. `copy.modules.to.distribution.destFile=${basedir}/../../..//${glassfish.distribution.dir}/lib/somefileinlib.jar`. You can use the `${glassfish.distribution.dir}` property which points to the `glassfish7/glassfish` folder relative to the project root. An example is in the `appserver/admingui/core/pom.xml` file

Note: This depends on the new 4.1.0 version of the `glassfish-build-maven-plugin` with this: https://github.com/eclipse-ee4j/glassfish-build-maven-plugin/pull/204.

### DIFF
--- a/appserver/admingui/cluster/pom.xml
+++ b/appserver/admingui/cluster/pom.xml
@@ -33,10 +33,6 @@
     <name>Admin Console Clustering Support Plugin</name>
     <description>Clustering support plugin bundle for GlassFish Admin Console</description>
 
-    <properties>
-        <copy.modules.to.distribution.skip>false</copy.modules.to.distribution.skip>
-    </properties>
-
     <dependencies>
         <dependency>
             <groupId>org.glassfish.main.admingui</groupId>
@@ -51,4 +47,13 @@
             <scope>provided</scope>
         </dependency>
     </dependencies>
+
+    <profiles>
+        <profile>
+            <id>dev</id>
+            <properties>
+                <copy.modules.to.distribution.skip>false</copy.modules.to.distribution.skip>
+            </properties>
+        </profile>
+    </profiles>
 </project>
--- a/appserver/admingui/cluster/pom.xml
+++ b/appserver/admingui/cluster/pom.xml
@@ -33,6 +33,10 @@
     <name>Admin Console Clustering Support Plugin</name>
     <description>Clustering support plugin bundle for GlassFish Admin Console</description>
 
+    <properties>
+        <copy.modules.to.distribution.skip>false</copy.modules.to.distribution.skip>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>org.glassfish.main.admingui</groupId>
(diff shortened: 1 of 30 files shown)


dmatej:
Shouldn't it be already possible just with 
```
mvn clean install -pl :security-ee -amd -Pfastest -T4C
```
It is quite standard without additional custom magic which we want to avoid in the future.

OndroMih:
> Shouldn't it be already possible just with

> `mvn clean install -pl :security-ee -amd -Pfastest -T4C`

No. With -amd, maven will also build tests and everything that somehow depends on the module, not just GlassFish distribution. I tried it and the build fails on a missing dependency, because some test module dependency isn't built with this command. Even if it's fixed, this command takes 2 minutes on my computer.

With this new `dev` profile, building, for example, `cluster-common` module and refreshing it in the GF distribution, takes around 10 seconds:

```
mvn -T4C clean install -Pfastest -Dcyclonedx.skip=true -pl :cluster-common -Pdev
[INFO] Total time:  8.578 s (Wall Clock)
```

The fastest I can get with a standard maven build is in 2 steps with:

```
mvn -T4C clean install -Pfastest -Dcyclonedx.skip=true -pl :cluster-common
[INFO] Total time:  8.718 s (Wall Clock)

mvn -T4C clean install -Pfastest -Dcyclonedx.skip=true -am -pl appserver/distributions/glassfish -rf appserver/featuresets
[INFO] Total time:  01:12 min (Wall Clock)
```
It still takes around 1:20 minutes compared to 10 seconds with the `dev` profile.

OndroMih:
I released the 4.1.0 version of glassfish-build-maven-plugin that provides the copy-file goal. I added configurations for a few more modules that I worked on in the past and used this dev mode on them locally.

Please review, and if OK, we can merge. The intention is to raise other PRs later to add support for more modules as we work on them.

arjantijms (approved):

PR #24993, opened 2024-06-22
New feature: Admin command logger

### LINKED ISSUE #24992: New feature: Admin command logger
Usecase 1:

* configure GlassFish using Admin Console and find out how to configure the same using the Asadmin CLI so that the configuration can be scripted

Usecase 2:

* log all admin operations via Admin Console, REST interface, Admin CLI for auditing purposes

Solution:

Provide a mechanism that intercepts Admin commands and logs them into server.log. It's possible to configure which category of commands should be logged or completely disable logging for all commands. Categories of commands: read operations (do not modify the configuration), write operations (modify the configuration), internal operations (commands not intended to be invoked directly).

Initially, enable the functionality and configure the category of logged commands via a system property set on DAS. Later can be configured as a standard domain.xml configuration via Asadmin CLI and Admin Console but that's not in scope of this issue.

OndroMih:
Fixes #24992

Provides a mechanism that logs admin commands that are invoked by the Admin Console, Admin CLI or the REST admin interface.

Demo of the functionality:

https://github.com/eclipse-ee4j/glassfish/assets/2195988/cde17e80-d649-4f21-b928-174fb4fc965c

Configuration as described in the docs in the PR. Copy of the docs here:

It's possible to log admin commands executed either using the asadmin CLI, Admin Console, or the REST interface.

To enable this logging, set the system property `glassfish.commandlogger.logmode` on the DAS server to one of the following values:

* `ALL_COMMANDS` - will log all commands
* `WRITE_COMMANDS` - will log all commands that modify server configuration
* `INTERNAL_COMMANDS` - will log all commands that modify server configuration and also those that start with `_` (internal commands)
* `READ_WRITE_COMMANDS` - will log all commands that modify server configuration or return info about the configuration
* `NO_COMMAND` - do not log any command, this is the default value if the property not specified

The log message will be logged using `org.glassfish.extras.commandlogger.AdminCommandLogger` logger with `INFO` level, and will contain:

* name of the admin user that executed the command
* name of the command
* arguments to the command

The name and arguments can be copied and pasted to the asadmin CLI to repeat running the same command from the command line.

Note: [SimpleTopicDistributionService.java](https://github.com/eclipse-ee4j/glassfish/pull/24993/files#diff-c33f51ea094f80d07bef1f9472e39ee1bf1a27b5e84125c134986e16d6f77d6b) is a temporary solution until https://github.com/eclipse-ee4j/glassfish-hk2/pull/1058 is released and a new version of HK2 is integrated.


nucleus/parent/pom.xml
```diff
@@ -879,6 +881,7 @@
                             <java.util.logging.manager>${test.logManager}</java.util.logging.manager>
                             <java.util.logging.config.useDefaults>true</java.util.logging.config.useDefaults>
                             <java.util.logging.config.defaultLevel>${test.logLevel}</java.util.logging.config.defaultLevel>
+                            <glassfish.suspend>${glassfish.suspend}</glassfish.suspend>
```
arjantijms, line 884: Just wondering where this is for. I mean, I know of course what suspend does, but what does defining is as a property do here over just using ${glassfish.suspend} elsewhere?

nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/adapter/JerseyContainerCommandService.java
```diff
@@ -138,8 +140,15 @@ private JerseyContainer getJerseyContainer(ResourceConfig config) {
             ServiceFinder.setIteratorProvider(iteratorProvider);
             RestLogging.restLogger.log(Level.FINEST,
                 () -> this + ": Creating Jersey container for " + HttpHandler.class + " and " + config);
-            final HttpHandler httpHandler = ContainerFactory.createContainer(HttpHandler.class, config);
-            return new JerseyContainer() {
+            final GrizzlyHttpContainer httpHandler = ContainerFactory.createContainer(GrizzlyHttpContainer.class, config);
+            final ServiceLocator jerseyLocator = httpHandler.getApplicationHandler().getInjectionManager().getInstance(ServiceLocator.class);
+        /*
+            We enable a temporary distribution service until the HK2 Extras package is fixed so that we can enable
```
arjantijms, line 146: What do we exactt need to fix in HK2 Extras?

Maybe good idea to create an issue here: https://github.com/eclipse-ee4j/glassfish-hk2-extra/issues ?

arjantijms: Interesting btw that HK2 has the hierarchy of bean archives just as CDI has.

avpinchuk: > What do we exactt need to fix in HK2 Extras?

@arjantijms, PR already in description, eclipse-ee4j/glassfish-hk2#1058 ))

nucleus/parent/pom.xml
OndroMih: The surefire plugin runs all tests in a separate JVM. System properties are not passed to this new JVM. So it's necessary to have this to pass the `glassfish.suspend` sys prop to tests.

nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/adapter/JerseyContainerCommandService.java
arjantijms: Thanks! Was looking at HK2 Extras itself, but it was already there ;)

arjantijms (approved):

nucleus/parent/pom.xml
arjantijms: I have been using glassfish.suspend with specifically the surefire plug-in a lot of times. No sure what is different then, but okay.

avpinchuk (approved):

hs536 (approved):

nucleus/parent/pom.xml
```diff
@@ -94,6 +94,8 @@
         <jakarta.inject-api.version>2.0.1.MR</jakarta.inject-api.version>
         <hk2.version>3.1.0</hk2.version>
         <hk2.plugin.version>3.1.0</hk2.plugin.version>
+<!--        <hk2.version>3.1.1-SNAPSHOT</hk2.version>
+        <hk2.plugin.version>3.1.1-SNAPSHOT</hk2.plugin.version>-->
```
dmatej, line 98: Now master depends on HK2 3.1.1 SNAPSHOT. We have to release it asap.

OndroMih: The snapshot version is commented out, so no harm. I forgot to remove this. I think it's OK to remove this comment when we upgrade hk2 in the future.

dmatej: Aha, I did not notice the comment goes through two lines :-)

PR #24997, opened 2024-06-22
Maven config templates

OndroMih:
This adds Maven config templates that can be used in `.mvn/maven.config` (see https://maven.apache.org/configure.html).

The idea is that people can copy and modify these templates in their `.mvn/maven.config` file, which wouldn't be versioned. Then it's easy to speed up the build by selecting the target artifact (uncomment that artifact in the `maven.config` file), as an alternative to #24534.

For example, instead of executing `mvn clean install -Pfastest -pl 'org.glassfish.main.distributions:glassfish' -am -T4C` on command line to build only the GlassFish Server distribution with the fasted profile, it's possible to copy `.mvn/maven.config.template` into `.mvn/maven.config` and uncomment the following:

```
-Pfastest
-T4C
-am
-pl
org.glassfish.main.distributions:glassfish
```

Then just execute `mvn clean install` to clean and build the GlassFish Server distribution. Or just `mvn` to build it without clean.

I think this is simpler and cleaner solution than which I proposed earlier in #24534. If this PR as merged, I'm going to close #24534.

In this PR, I also added the `-Dcyclonedx.skip=true` property to the `fastest` profile. The CycloneDX SBOM check, which is not necessary and slows down the build by a few seconds: https://cyclonedx.org/.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -119,6 +119,9 @@
          -->
         <profile>
             <id>fastest</id>
+            <properties>
+                <cyclonedx.skip>true</cyclonedx.skip>
+            </properties>
             <modules>
                 <module>qa</module>
                 <module>nucleus</module>
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.mvn/maven.config
 *.ipr
 *.iml
 *.iws
--- a/.mvn/maven.config.template
+++ b/.mvn/maven.config.template
@@ -1,0 +1,81 @@
+############################################
+# Build GlassFish Server Full Distribution #
+############################################
+
+#-Pfastest
+#-T4C
+#-am
+#-pl
+#org.glassfish.main.distributions:glassfish
+
+###########################################
+# Build GlassFish Server Web Distribution #
+###########################################
+
+#-Pfastest 
+#-T4C
+#-am
+#-pl
+#org.glassfish.main.distributions:web
+
+##############################################
+# Build GlassFish Embedded Full Distribution #
+##############################################
+
+#-Pfastest
+#-T4C
+#-am
+#-pl
+#org.glassfish.main.extras:glassfish-embedded-all
+
+##############################################
+# Build GlassFish Embedded Full Distribution #
+##############################################
+
+#-Pfastest
+#-T4C
+#-am
+#-pl
+#org.glassfish.main.extras:glassfish-embedded-web
+
+###########################################################################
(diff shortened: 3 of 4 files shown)


pom.xml
```diff
@@ -119,6 +119,9 @@
          -->
         <profile>
             <id>fastest</id>
+            <properties>
+                <cyclonedx.skip>true</cyclonedx.skip>
```
pzygielo, line 123: Maybe set this property:
https://github.com/eclipse-ee4j/ee4j/blob/bd303c4e8badb547d505db7272f3319a1f4d16a0/parent/pom.xml#L200: https://github.com/eclipse-ee4j/glassfish/blob/bd303c4e8badb547d505db7272f3319a1f4d16a0/parent/pom.xml#L200

This would not add plugin to the build at all and avoid its execution (not only skipping it).

OndroMih: The `skipSBOM` property deactivates a profile, therefore it has to be set on maven's command line. If it's set in the `fastest` profile, it's too late to deactivate the sbom. It's not possible to activate or deactive a profile from another profile.

pzygielo: Could the sbom profile be deactivated in maven.config?

pzygielo: Having `<cyclonedx.skip>true</cyclonedx.skip>` as above in `fastest` is perfect on its own and worth to have it in that profile.

pzygielo (approved):

dmatej:
The cyclonedx plugin produces files which are then deployed to maven central. So now the `fastest` profile must not be used to release the version. However for some reason the cyclonedx runs differently in master and in 8.0 and I don't like it too, we should spend some time with that.

OndroMih:
I thought that the fastest profile was only for development, and the fast profile is used in CI. We can revert the change in the profile and move the property to maven.config template, would that be better

arjantijms:
The main build for the deployment to staging on the CI doesn't use the fast profile, it's:

```
mvn -U -B 
     -Poss-release,release-phase2 
     ${MVN_STAGING} 
     -DskipTests -DstagingProfileId=1c0c18a0fc339 -Ddoclint=none -Dmaven.javadoc.failOnError=false 
     clean deploy
```

dmatej:
@arjantijms Check the whole script, it is used. I did not read it line by line, but it is there. Soon we will have migrate from the nexus staging plugin, maybe that would be an opportunity to simplify and standardize things, all those bash scripts are quite hard to read (and reproduce issues locally).

PR #25027, opened 2024-07-05
Fix the startserv script if server is already running

OndroMih:
Also report the error immediately, don't run the command again

Adds tests for the startserv script. For that, improves the process manager to allow waiting until a text it output instead of waiting for the process to complete.

### DIFF
--- a/nucleus/common/common-util/src/test/java/com/sun/enterprise/universal/process/ProcessManagerTest.java
+++ b/nucleus/common/common-util/src/test/java/com/sun/enterprise/universal/process/ProcessManagerTest.java
@@ -17,6 +17,8 @@
 
 package com.sun.enterprise.universal.process;
 
+import static java.time.Duration.ofSeconds;
+
 import com.sun.enterprise.util.OS;
 
 import java.io.File;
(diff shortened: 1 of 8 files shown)


nucleus/common/common-util/src/test/java/com/sun/enterprise/universal/process/ProcessManagerTest.java
```diff
@@ -132,11 +131,12 @@ void testWaitUntilTextInStdErr() {
     @Timeout(value = 5, unit = TimeUnit.SECONDS)
     @DisabledOnOs(WINDOWS)
     void testDetectTextInStdOutIfProcessStops() throws InterruptedException {
+        int sleepTimeSeconds = 1;
         ProcessManager pm = new ProcessManager("sh", "-c", "echo \"start\nhello\ncontinue\"");
         pm.setEcho(false);
         pm.setTextToWaitFor("hello");
         int exitCode = assertDoesNotThrow(pm::execute);
-        Thread.sleep(ofSeconds(1));
+        Thread.sleep(sleepTimeSeconds * 1000);
```
dmatej, line 139: You can write also 1_000L, no need to create a variable if it doesn't have any other usage.

OndroMih: I wanted to do that first, but then I saw that other test method follow this pattern so I did it this way. If we used Java 21, I would rather use Duration, but with millis I’m not sure which approach is better.

dmatej (approved):

PR #24657, opened 2023-10-26
Update pom.xml EE4J parent to 1.0.9

arjantijms:
Changes: https://github.com/eclipse-ee4j/ee4j/compare/1.0.8...1.0.9

Issue #25026, opened 2024-07-04
The master release doesn't deploy the CycloneDX files


dmatej:
See
* https://repo1.maven.org/maven2/org/glassfish/main/common/glassfish-api/8.0.0-M6/ - CycloneDX files are there
* https://repo1.maven.org/maven2/org/glassfish/main/common/glassfish-api/7.0.15/ - Here they don't

Note that in #24997 is another change that might be related in the future. Also I am not sure what are standard rules for this, we should read some docs about it.

avpinchuk:
One difference between 7.x and 8.x:

7.x:
https://github.com/eclipse-ee4j/glassfish/blob/1d3efd38c4ba219fa445ed32025f4cdc5c17fc69/nucleus/parent/pom.xml#L27

8.x:
https://github.com/eclipse-ee4j/glassfish/blob/59fdd9b172310c2423be2126b4df030c1776dd71/nucleus/parent/pom.xml#L27

Therefore, the assembly of the GF8 gives us a lot of CycloneDX warnings.

dmatej:
The fastest profile is used in the release process 
* Because of the parent from Eclipse (1.0.8 vs 1.0.9) it differs between master and 8.0. 
* The CycloneDX significantly slows down depeloper builds with fastest; developers usually don't care about CycloneDX.
* The target of fastest is to create all deployables asap, so here we have some controversy.

Also developers can have own profiles in settings.xml. Maybe there is a better place for the cyclone. Or we need another profile for the release process. However at some point it would be useful to have the OWASP check locally too, but that is different plugin.

Hmmm, they have different targets. Probably the CycloneDX could be executed really just for releases + explicitly.
* https://github.com/CycloneDX/cyclonedx-maven-plugin
* https://jeremylong.github.io/DependencyCheck/dependency-check-maven/index.html

dmatej:
Yet one idea - we can also push the cyclonedx plugin execution to the deploy phase.

pzygielo:
Maybe in release job, where `fastest` profile is activated, the `sbom` profile from parent could also be activated explicitly?

PR #25028, opened 2024-07-05
Update EE4J/parent to 1.0.9

pzygielo:
- Like #24657 but also in `nucleus/parent`

### DIFF
--- a/appserver/tests/quicklook/pom.xml
+++ b/appserver/tests/quicklook/pom.xml
@@ -24,7 +24,7 @@
     <parent>
         <groupId>org.eclipse.ee4j</groupId>
         <artifactId>project</artifactId>
-        <version>1.0.8</version>
+        <version>1.0.9</version>
         <relativePath/>
     </parent>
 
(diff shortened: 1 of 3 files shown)


avpinchuk:
Parent update give us a lot of warnings (in GF 8 too):

```
[2024-07-05T12:13:49.009Z] [WARNING] *****************************************************************
[2024-07-05T12:13:49.009Z] [WARNING] * An aggregator mojo is already being executed in this          *
[2024-07-05T12:13:49.009Z] [WARNING] * parallel build, those kind of mojos require exclusive access  *
[2024-07-05T12:13:49.009Z] [WARNING] * to reactor to prevent race conditions. This mojo execution    *
[2024-07-05T12:13:49.009Z] [WARNING] * will be blocked until the aggregator mojo is done.            *
[2024-07-05T12:13:49.009Z] [WARNING] *****************************************************************
```
Is it correct?

pzygielo:
> Parent update give us a lot of warnings (in GF 8 too):
> Is it correct?

Not sure why maven does that as warning [MNG-7476](https://issues.apache.org/jira/browse/MNG-7476), if it's completely handled on its side:

https://github.com/eclipse-ee4j/glassfish/blob/bc0240f3c744dd6b6ec2920b3cd08dcc295161ae/maven-core/src/main/java/org/apache/maven/lifecycle/internal/MojoExecutor.java#L237-L241

But now we know, that that happens.

dmatej:
It is ok, we can see the same in the 8.0 branch. Also this update slows down the build, that is why we made #24997 and then #25026. I will play with the CycloneDX in next few days (it depends on how much tired I am).

dmatej (approved):

avpinchuk (approved):

Issue #24141, opened 2022-10-25
check-distribution warnings and errors reported for glassfish build


pzygielo:
The build of the `glassfish` distribution produces spec errors. 

Reproducible with  `mvn clean install -Pfastest -pl org.glassfish.main.distributions:glassfish`.

```
[ jakarta.authentication:jakarta.authentication-api:3.0.0 ]
{ groupIdPrefix=jakarta. apiPackage=jakarta.security.auth.message API final spec-impl-version=3.0.0 }
- ERROR: missing configuration ( spec-version )


[ jakarta.mvc:jakarta.mvc-api:2.1.0 ]
{ groupIdPrefix=jakarta. API final spec-impl-version=null }
- ERROR: missing configuration ( spec-version api-package spec-impl-version )


[ jakarta.validation:jakarta.validation-api:3.0.2 ]
{ groupIdPrefix=jakarta. API final spec-impl-version=null }
- ERROR: Extension-Name not found in MANIFEST
- ERROR: Specification-Version not found in MANIFEST
- ERROR: Implementation-Version not found in MANIFEST
- ERROR: missing configuration ( spec-version api-package spec-impl-version )


[ jakarta.json:jakarta.json-api:2.1.1 ]
{ groupIdPrefix=jakarta. API final spec-impl-version=null }
- ERROR: missing configuration ( spec-version api-package spec-impl-version )


[ jakarta.activation:jakarta.activation-api:2.1.0 ]
{ groupIdPrefix=jakarta. spec-version=2.1 apiPackage=jakarta.activation API final spec-impl-version=2.1.0 }
- ERROR: Implementation-Version not found in MANIFEST


[ jakarta.json.bind:jakarta.json.bind-api:3.0.0 ]
{ groupIdPrefix=jakarta. API final spec-impl-version=null }
- ERROR: missing configuration ( spec-version api-package spec-impl-version )


[ org.glassfish.main.core:jakartaee-kernel:7.0.0-SNAPSHOT ]
{ groupIdPrefix=jakarta. API final spec-impl-version=null }
- ERROR: Extension-Name not found in MANIFEST
- ERROR: Specification-Version not found in MANIFEST
- ERROR: Implementation-Version not found in MANIFEST
- ERROR: missing configuration ( spec-version api-package spec-impl-version )


[ jakarta.batch:jakarta.batch-api:2.1.1 ]
[...]

PR #25032, opened 2024-07-07
Fixes Jakarta Servlet API version

avpinchuk:
Fixes Jakarta Servlet API version in various POMs.

If we build GlassFish with the empty maven cache:

```sh
mvn clean install -Pstaging
```
the local maven repo contains strange artifact:

image

### DIFF
--- a/appserver/pom.xml
+++ b/appserver/pom.xml
@@ -887,7 +887,7 @@
                             </artifact>
                             <nonFinal>false</nonFinal>
                             <jarType>api</jarType>
-                            <specVersion>${jakarta.servlet-api.version}</specVersion>
+                            <specVersion>6.0</specVersion>
                             <specImplVersion>${jakarta.servlet-api.version}</specImplVersion>
                             <apiPackage>jakarta.servlet</apiPackage>
                         </spec>
--- a/appserver/pom.xml
+++ b/appserver/pom.xml
@@ -883,12 +883,12 @@
                             <artifact>
                                 <groupId>jakarta.servlet</groupId>
                                 <artifactId>jakarta.servlet-api</artifactId>
-                                <version>${servlet-api.version}</version>
+                                <version>${jakarta.servlet-api.version}</version>
                             </artifact>
                             <nonFinal>false</nonFinal>
                             <jarType>api</jarType>
-                            <specVersion>3.1</specVersion>
-                            <specImplVersion>${servlet-api.version}</specImplVersion>
+                            <specVersion>${jakarta.servlet-api.version}</specVersion>
+                            <specImplVersion>${jakarta.servlet-api.version}</specImplVersion>
                             <apiPackage>jakarta.servlet</apiPackage>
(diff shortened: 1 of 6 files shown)


appserver/pom.xml
```diff
@@ -883,12 +883,12 @@
                             <artifact>
                                 <groupId>jakarta.servlet</groupId>
                                 <artifactId>jakarta.servlet-api</artifactId>
-                                <version>${servlet-api.version}</version>
+                                <version>${jakarta.servlet-api.version}</version>
                             </artifact>
                             <nonFinal>false</nonFinal>
                             <jarType>api</jarType>
-                            <specVersion>3.1</specVersion>
```
pzygielo, line 890: However - the spec is only versioned to 2 components (like 6.0 in JakartaEE 10), while the `jakarta.servlet-api.version` property holds the API artifact version (with bugfix component). I see below for annotation it's kept in-sync with the same value/property, but I'm not sure what is the role of `<specVersion>` and if it should match the implementation to the point... I imagine it would not be here to keep duplicate...

But `3.1` is of course not good here as well :grin:

appserver/pom.xml
```diff
@@ -883,12 +883,12 @@
                             <artifact>
                                 <groupId>jakarta.servlet</groupId>
                                 <artifactId>jakarta.servlet-api</artifactId>
-                                <version>${servlet-api.version}</version>
```
pzygielo, line 886: I wonder why this missing property didn't bother the plugin...

pzygielo: Seems it does [master:1488](https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/job/master/1488/):
```
[2024-07-07T10:38:50.796Z] [ jakarta.servlet:jakarta.servlet-api:6.0.0 ]
[2024-07-07T10:38:50.796Z] { groupIdPrefix=jakarta. API final spec-impl-version=null }
[2024-07-07T10:38:50.796Z] - ERROR: missing configuration ( spec-version api-package spec-impl-version )
```
but not enough to cause failure.

pzygielo:
> If we build GlassFish with the empty maven cache:
> 
> ```shell
> mvn clean install -Pstaging
> ```
> 
> the local maven repo contains strange artifact:
> 

I suppose: maven resolver tries to download specified version (https://github.com/eclipse-ee4j/glassfish/pull/25032#discussion_r1667698392) but there is no such property to be interpolated so it stays as is.

pzygielo (approved):

arjantijms (approved):

dmatej:
Aha, I got it, dependencies are specified correctly (except tests, where it is reported as an error by Eclipse IDE; maybe those tests fail to compile but bash script ignores the result? We should check ...), just the spec plugin is wrong and unfortunately doesn't crash the build even with unresolved properties.

Good catch!

pzygielo:
- #24141

PR #25033, opened 2024-07-07
Synchronize expectations for spec versions

pzygielo:
- Additional findings after #25032
- #24141

### DIFF
--- a/appserver/pom.xml
+++ b/appserver/pom.xml
@@ -790,7 +790,7 @@
                             </artifact>
                             <nonFinal>false</nonFinal>
                             <jarType>impl</jarType>
-                            <specVersion>3.0</specVersion>
+                            <specVersion>4.0</specVersion>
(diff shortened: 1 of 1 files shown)


appserver/pom.xml
```diff
@@ -971,8 +971,8 @@
                             </artifact>
                             <nonFinal>false</nonFinal>
                             <jarType>api</jarType>
-                            <specVersion>2.3</specVersion>
-                            <specImplVersion>2.3.1</specImplVersion>
+                            <specVersion>3.1</specVersion>
+                            <specImplVersion>3.1.1</specImplVersion>
```
avpinchuk, line 975: Why not use `${jakarta.pages-api.version}` instead of hard-coded value?

pzygielo: > Why not use `${jakarta.pages-api.version}` instead of hard-coded value?

I will update.

avpinchuk (approved):

PR #25038, opened 2024-07-10
Optimize GJULE log record source autodetection

avpinchuk:
The `StackWalker` is used to autodetect source class and method.

### DIFF
--- a/nucleus/glassfish-jul-extension/src/main/java/org/glassfish/main/jul/record/GlassFishLogRecord.java
+++ b/nucleus/glassfish-jul-extension/src/main/java/org/glassfish/main/jul/record/GlassFishLogRecord.java
@@ -39,20 +39,6 @@ public class GlassFishLogRecord extends LogRecord {
 
     private static final ZoneId TIME_ZONE = ZoneId.systemDefault();
 
-    private static final StackWalker STACK_WALKER = StackWalker.getInstance(StackWalker.Option.SHOW_REFLECT_FRAMES);
-
-    private static final Set<String> IGNORED_CLASSES = Set.of(
-        "org.glassfish.main.jul.GlassFishLogger",
-        "org.glassfish.main.jul.GlassFishLoggerWrapper",
-        // see LogDomains in GlassFish sources
-        "com.sun.logging.LogDomainsLogger",
-        // remaining classes are in the JDK
-        "java.util.logging.Logger",
-        "java.util.logging.LoggingProxyImpl",
-        // see LoggingPrintStream
-        "java.lang.Throwable"
-    );
-
     private final LogRecord record;
     private final String threadName;
     private String messageKey;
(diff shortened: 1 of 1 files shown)


nucleus/glassfish-jul-extension/src/main/java/org/glassfish/main/jul/record/GlassFishLogRecord.java
```diff
@@ -289,37 +306,24 @@ public String toString() {
...
-            if (!isIgnoredStackTraceElement(className)) {
-                wrappedRecord.setSourceClassName(className);
-                wrappedRecord.setSourceMethodName(element.getMethodName());
-                return;
-            }
-        }
+        final Optional<StackWalker.StackFrame> stackFrame = STACK_WALKER.walk(stackFrames ->
+            stackFrames.dropWhile(frame -> !isIgnoredStackFrame(frame.getClassName()))
+                .filter(frame -> !isIgnoredStackFrame(frame.getClassName()))
+                .findFirst());
```
pzygielo, line 312: I'd drop local variable `stackFrame`, continue with `.ifPresent`, and remove import for `Optional`.

dmatej (approved):

nucleus/glassfish-jul-extension/src/main/java/org/glassfish/main/jul/record/GlassFishLogRecord.java
```diff
@@ -287,41 +289,43 @@ public String toString() {
...
-            || sourceClassName.startsWith("sun.reflect.");
+            "java.lang.Throwable"
+        );
+
+        private static final StackWalker STACK_WALKER = StackWalker.getInstance(StackWalker.Option.SHOW_REFLECT_FRAMES);
+
+        static void detectClassAndMethod(final LogRecord wrappedRecord) {
+            STACK_WALKER
+                .walk(stackFrames ->
+                    stackFrames.dropWhile(frame -> !isIgnoredStackFrame(frame.getClassName()))
```
pzygielo, line 311: Just for fun - if there only could be `isNotIgnoredStackFrame(StackWalker.StackFrame frame)` method (`::`-referenced in both `dropWhile` and `filter`)...

arjantijms (approved):

nucleus/glassfish-jul-extension/src/main/java/org/glassfish/main/jul/record/GlassFishLogRecord.java
dmatej: @avpinchuk @pzygielo Can you do that? It is a good idea, unfortunately @arjantijms already merged the PR, so you have to create new one.

avpinchuk: Sure. But as a separate PR ))

PR #25042, opened 2024-07-11
Better cyclonedx handling + release process + small improvements

### LINKED ISSUE #25026: The master release doesn't deploy the CycloneDX files

dmatej:
*  Closes #25026 
* cyclonedx from eclipse's parent overridden
   * We don't deploy the aggregator pom, no need to run it there
   * The execution was moved to the oss-release profile, which is used for deployment
* Renamed release-phase2 to oss-release
   * Both profiles are activated in the single execution; first was in GlassFish, second is in parent.
   * There is no release-phase1 or other phases
* Fixed maven XSD locations - http to https
* Added skipping javadoc errors to the oss-release profile; it is explicitly set on CI and build fails without it.
* Try locally
   * `mvn clean install -Pfastest,oss-release `# don't use parallel builds with T4C, seems gpg plugin is not thread safe
   * `mvn clean validate -Ptck` # to check that all POM files are okay after the last commit.

### Note For Reviewers

* Please read individual commit - each contains individual change. 
* Otherwise I would create 5 individual PRs which would send you 5 individual notifications and you would review 5 individual PRs. Every PR would be built by CI (maybe several times).
* All commits are quite simple and related to the build and the release process.

### DIFF
--- a/appserver/common/amx-jakartaee/pom.xml
+++ b/appserver/common/amx-jakartaee/pom.xml
@@ -17,7 +17,7 @@
 
 -->
 
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsdd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
--- a/appserver/common/amx-jakartaee/pom.xml
+++ b/appserver/common/amx-jakartaee/pom.xml
@@ -17,7 +17,7 @@
 
 -->
 
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsdd">
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
--- a/appserver/admin/admin-core/pom.xml
+++ b/appserver/admin/admin-core/pom.xml
@@ -17,7 +17,7 @@
 
 -->
 
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
--- a/appserver/admin/admin-core/pom.xml
+++ b/appserver/admin/admin-core/pom.xml
@@ -80,7 +80,7 @@
 
     <profiles>
         <profile>
-            <id>release-phase2</id>
+            <id>oss-release</id>
             <activation>
                 <activeByDefault>false</activeByDefault>
             </activation>
--- a/appserver/admin/backup/pom.xml
+++ b/appserver/admin/backup/pom.xml
@@ -17,7 +17,7 @@
 
 -->
 
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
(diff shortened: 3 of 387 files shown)


OndroMih:
Could you move the commit [Fixed Maven XSD location - http changed to https](https://github.com/eclipse-ee4j/glassfish/pull/25042/commits/1a6622ec17106b8c78b2938bb5fbf4d1954c7685) into a separate PR? It touches a lot of files, it's not related to the other changes and makes it hard to review the important changes in other commits.

dmatej:
> Could you move the commit [Fixed Maven XSD location - http changed to https](https://github.com/eclipse-ee4j/glassfish/pull/25042/commits/1a6622ec17106b8c78b2938bb5fbf4d1954c7685) into a separate PR? It touches a lot of files, it's not related to the other changes and makes it hard to review the important changes in other commits.

It also saves CI resources and you can view commits one by one.

OndroMih:
> It also saves CI resources and you can view commits one by one.

I’m sorry, but I disagree with this approach. Checking commits one by one is tedious, especially if there are multiple commits. I would either squash all other commits to one, or mve the big commit to another PR. Otherwise it’s an unnecessary burden on reviewers. (I did review commits one by one indeed, and it’s much harder than reviewing a single changeset)

dmatej:
> > It also saves CI resources and you can view commits one by one.
> 
> I’m sorry, but I disagree with this approach. Checking commits one by one is tedious
* You mean that reviewing 5 standalone PRs built by 5 standalone CI builds is more effective? I don't think so.

>, especially if there are multiple commits. I would either squash all other commits to one
* That is absolutely wrong - every commit contains standalone change, just first two are related to each other, but still first just disables cyclone in aggregator, which is not deployed at all, second aligns the cyclone with our release process.

> or mve the big commit to another PR. 
* That would make sense except unnecessary load on CI, because all these changes are just trivial http -> https replacement in XML header.

> Otherwise it’s an unnecessary burden on reviewers
* Valid only if you are reading diff of the whole PR and call it a review. 

> (I did review commits one by one indeed, and it’s much harder than reviewing a single changeset)
* How it can be harder here? Maybe I miss something, but how reading these individual commits can be hard?
   * https://github.com/eclipse-ee4j/glassfish/pull/25042/commits/c86d088640cfb9b0d45ab2b87f9b3f4dde18355e
   * https://github.com/eclipse-ee4j/glassfish/pull/25042/commits/ab22a6101f74a5ae63d5c53815ce888d8c965e47
   * https://github.com/eclipse-ee4j/glassfish/pull/25042/commits/c4cc2c74296fdf11e17169ab74799c271a5a30ce
   * https://github.com/eclipse-ee4j/glassfish/pull/25042/commits/1a6622ec17106b8c78b2938bb5fbf4d1954c7685
   * https://github.com/eclipse-ee4j/glassfish/pull/25042/commits/6552b8cf4036ad893db19af35c331c0f6e4fc03f
   * Also - when reviewing diff of the whole PR, you don't see descriptions of individual commits.

OndroMih:
OK, I get it now, @dmatej. If the commits are more or less separate and don't depend on each other, then it makes sense to add them into a single PR and reviewers review them separately. It wasn't clear to me until you clarified it and also added more info into the description. Thank you for that.

OndroMih (changes requested):
I found a glitch in the commit [Fixed Maven XSD location - http changed to https](https://github.com/eclipse-ee4j/glassfish/pull/25042/commits/1a6622ec17106b8c78b2938bb5fbf4d1954c7685) - several files have `.xsdd">` instead of `.xsd">`.

For example:

```
diff --git a/appserver/itest-tools/pom.xml b/appserver/itest-tools/pom.xml
index f78fbc9f8cd..3fcc6244651 100644
--- a/appserver/itest-tools/pom.xml
+++ b/appserver/itest-tools/pom.xml
@@ -1,5 +1,5 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsdd">
     <modelVersion>4.0.0</modelVersion>
```

appserver/common/amx-jakartaee/pom.xml
```diff
@@ -17,7 +17,7 @@
 
 -->
 
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsdd">
```
dmatej, line 20: Oh god, really! Thanks, that must be another bug in Eclipse IDE, sad ☹️
I will fix it when I get to my laptop, and I am happy we agreed.

arjantijms: FYI, tidy:1.2.0:pom does this transformation too.

dmatej: So it is really a bug in Eclipse ... now it swallowed the `"` character and then again duplicated the last one. Meh!!!

dmatej: Aha, the problem occurs when it is a subproject once imported to Eclipse IDE and once seen as a subfolder of another project. Eclipse now does the conversion twice and that has some unpredictable results. The workaround is to exclude duplicities, or to mark those subfolders manually as "derived".

OndroMih:
Looks OK now. For those, who would like to review the 2 commits related to the XSD schema location, here's the diff of both commits combined: https://github.com/OndroMih/glassfish/pull/2/files

OndroMih (approved):

avpinchuk (approved):

PR #25044, opened 2024-07-12
Enhancement: Respect scheme from the serverName listener property in redirect URLs

### LINKED ISSUE #24918: Enhancement: Allow to specify protocol for URLs the server sends to the client
## Environment Details
* GlassFish 7.0.14
* JDK version: Any
* OS: Any
* Database: N/A

----------

## Background (motivation)

URLs generated by GlassFish take into consideration the value of the `Server Name` configuration value in `config -> Network Config -> Protocols -> (listener) -> HTTP`. When `Server Name` is set, its value (including port number if specified) will be used instead of localhost in generated URLs. The generated URLs now always retain the protocol (HTTP or HTTPS) used by the listener.

In case when GlassFish is behind a proxy server, where the proxy server communicates with the outside world via HTTPS, and GlassFish communicates with the proxy via plain HTTP inside an internal network, GlassFish should generate URLs to respect how outside world communicates with the proxy, including using HTTPS in the URL. This doesn't happen now, it's only possible to create URLs with a custom domain and port, but not with a custom protocol. The URLs always start with `http://` and there's no way to turn them into `https://` in GlassFish.

An example:
* Web browser uses https://mydomain.com/myapp to access the proxy server
* Proxy server forwards the request to GlassFish via a request to http://192.168.0.5:8080/myapp
* If Server Name is empty, GlassFish now issues a redirect, with "Location" header "http://192.168.0.5:8080/myapp/index.xhtml"
* If Server Name is set to mydomain.com:80, GlassFish now issues a redirect, with "Location" header "http://mydomain.com/myapp/index.xhtml"
* Ideally, the "Location" header in the redirect should contain HTTPS, such as "https://mydomain.com/myapp/index.xhtml"

## Enhancement Description

Add a configuration option for the HTTP protocol, e.g. `Server Protocol` (to be in line with `Server Name`), which allows to change the protocol used in the generated URLs. So that if HTTPS is specified, the URL will use the https:// protocol.

Alternatively, allow using the protocol in the `Server Name` value, so that it could be `https://mydomain.com:443". Currently, the protocol is stripped from the value of "Server Name" and the value is changed to just `mydomain.com:443`.

## Steps to reproduce

1. Deploy a simple application, e.g. with a html page. Set context root to /myapp.
2. Configure Server Name for the http-listener-1 to `mydomain.com:443`
3. Open http://localhost:8080/myapp in a browser (8080 is the port of the http-listener-1)
4. The browser will redirect to `http://mydomain.com:443/myapp/` (which will end with "The connection was reset" which is expected)

image

With the enhancement and proper configuration, in point 4, the browser would redirect to `https://mydomain.com/myapp/`

## Impact of Issue
Simpler and safer configuration of a proxy server as a load balancer for a GlassFish cluster.

Type: Enhancement
Component: Network
Priority: Normal

OndroMih:
ServerName property on the HTTP listener can contain scheme prefix (e.g. https://), which will be respected in redirects created by GlassFish.

For example, if GlassFish is behind a proxy server, connections to the proxy server go through HTTPS but connections from proxy server to GlassFish go through HTTP. If the proxy server is exposed at https://my.domain.com, then it's possible to set the serverName to https://my.domain.com:443 to use https://my.domain.com/ in redirects. This URL doesn't have to be translated by the proxy server because it's the correct URL where the proxy server listens to incoming requests.

This resolves #24918.

### DIFF
--- a/appserver/web/web-core/src/main/java/org/apache/catalina/connector/Connector.java
+++ b/appserver/web/web-core/src/main/java/org/apache/catalina/connector/Connector.java
@@ -197,17 +197,23 @@ public class Connector implements org.apache.catalina.Connector, Lifecycle {
 
     /**
      * The server name to which we should pretend requests to this Connector were directed. This is useful when operating
-     * Tomcat behind a proxy server, so that redirects get constructed accurately. If not specified, the server name
+     * GlassFish behind a proxy server, so that redirects get constructed accurately. If not specified, the server name
      * included in the <code>Host</code> header is used.
      */
     private String proxyName;
 
     /**
      * The server port to which we should pretend requests to this Connector were directed. This is useful when operating
-     * Tomcat behind a proxy server, so that redirects get constructed accurately. If not specified, the port number
+     * GlassFish behind a proxy server, so that redirects get constructed accurately. If not specified, the port number
      * specified by the <code>port</code> property is used.
      */
     private int proxyPort = 0;
+
+    /**
+     * The HTTP scheme to which we should pretend requests to this Connector were directed. This is useful when operating
+     * GlassFish behind a proxy server, so that redirects get constructed accurately.
+     */
+    private String proxyScheme = null;
 
     /**
      * The redirect port for non-SSL to SSL redirects.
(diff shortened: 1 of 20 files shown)


appserver/web/web-core/src/main/java/org/apache/catalina/connector/Connector.java
```diff
@@ -197,18 +197,24 @@ public class Connector implements org.apache.catalina.Connector, Lifecycle {
...
+     * GlassFish behind a proxy server, so that redirects get constructed accurately. If not specified, the port number
      * specified by the <code>port</code> property is used.
      */
     private int proxyPort = 0;
 
+    /**
+     * The HTTP scheme to which we should pretend requests to this Connector were directed. This is useful when operating
+     * GlassFish behind a proxy server, so that redirects get constructed accurately.
+     */
+    private String proxyScheme = null;
```
arjantijms, line 216: It's extremely minor, but null is already the default and can be left out.

arjantijms (approved):

PR #25045, opened 2024-07-13
Parsson now supports Annotations 3.0, no need for patch anymore.

arjantijms:
Also removed some stale other patch folders.

See also https://github.com/eclipse-ee4j/parsson/issues/118


arjantijms:
p.s. Parsson updated the major dependency to Annotations 3 in a small patch release, which is why that release doesn't work on GF 7 anymore.

avpinchuk (approved):

PR #25053, opened 2024-07-16
Update various maven plug-in versions

arjantijms:


OndroMih (approved):

avpinchuk (approved):

dmatej (approved):

appserver/tests/application/pom.xml
```diff
@@ -157,7 +157,7 @@
             <plugin>
                 <groupId>com.sun.xml.ws</groupId>
                 <artifactId>jaxws-maven-plugin</artifactId>
-                <version>3.0.0</version>
+                <version>4.0.3</version>
```
pzygielo, line 160: Who could release it to central?

dmatej: Btw 7.0.16 is blocked by Metro now - 4.0.4 is in staging, but @lukasj has to push it to the Central. I expect this will step to 4.0.4 too then ...

dmatej: Hmmm, aha, maybe not: https://mvnrepository.com/artifact/com.sun.xml.ws/jaxws-maven-plugin
4.0.3 is in staging too but not in central, it seems we have another blocker ... https://github.com/eclipse-ee4j/metro-jax-ws/tags

dmatej: Ok, I have a message from Lukáš - both artifacts were released, Central doesn't have them in the website cache yet, but they are already available for Maven.

PR #25056, opened 2024-07-19
Refactoring in security (TLS)

dmatej:
* Simplified
* Fixed many NPEs and swallowed exceptions
* Fixed HK2 related issues (missing `@Contract`, inaccessible classes in a context)
* Narrowed dependencies (a bit)
* I believe it is not worse than before :-) , but the progress is very slow and painful. Usual mistakes:
  * Typo in reflection. GlassFish vs. Glassfish took 3 hours.
  * Missing `@Contract` was the cause of all that original boilerplate code
  * Swallowed exceptions -> misleading logs reporting "something broken" when it was already broken three steps before.
  * Added dependency just for one import = bad architecture, something should be on another place.
* All tests passed on OmniFish's Jenkins including TCK

### Review

* Better per commit. All changes are related - usually I broke something then noticed the error is not logged or is shadowed by another error, so I had to fix that first. Finally I reached some "safe point", committed, and then I split that commit to several smaller commits to at least partially meaningful blocks of the work.


nucleus/grizzly/config/src/main/java/org/glassfish/grizzly/config/ssl/JSSE14Support.java
```diff
@@ -20,61 +21,42 @@
...
+import static java.lang.System.Logger.Level.ERROR;
 
-   This will only work with JDK 1.2 and up since it
-   depends on JDK 1.2's certificate support
-
-   @author EKR
-   @author Craig R. McClanahan
-   Parts cribbed from JSSECertCompat
-   Parts cribbed from CertificatesValve
-*/
```
avpinchuk, line 40: Should we remove this comment?

dmatej: The comment was copy pasted from JSSESupport which came from Tomcat - which now has completely different sources. In our JSSESupport is this comment too. Those two mentioned classes, JSSECertCompat and CertificatesValve are not on classpath and - probably they existed around year 2000 somewhere (aha, CertificatesValve is in tomcat4, 2001, I did not find the second one), but without package names it doesn't even make sense to have them here.
I think the JSSE14Support and JSSESupport could be merged to one, probably in next PR. We don't support Java 1.2 and 1.4 any more ...

avpinchuk (approved):

nucleus/common/common-util/src/test/java/com/sun/enterprise/universal/process/ProcessManagerTest.java
```diff
@@ -131,12 +131,15 @@ void testWaitUntilTextInStdErr() {
     @Timeout(value = 5, unit = TimeUnit.SECONDS)
     @DisabledOnOs(WINDOWS)
     void testDetectTextInStdOutIfProcessStops() throws InterruptedException {
```
pzygielo, line 133: Could `IEx` be undeclared now here?

nucleus/common/mbeanserver/src/main/java/org/glassfish/admin/mbeanserver/ssl/SecureRMIServerSocketFactory.java
```diff
@@ -45,7 +46,7 @@
 public class SecureRMIServerSocketFactory extends SslRMIServerSocketFactory {
 
     private final InetAddress mAddress;
-    private final ServiceLocator habitat;
+    private final ServiceLocator locator;
```
arjantijms, line 49: Absolutely tiny remark, and no need to change it, but I think "serviceLocator" as variable name would be more clear (various other locations use that too)

dmatej: We use both, I prefer shorter version as it is clear in the context what it is.

PR #25079, opened 2024-08-07
Faster autodeploy at startup - do not wait before first autodeployment scan

### LINKED ISSUE #25078: Automatic deployment waits for pollingInterval time before it scans the autodeploy directory for the first time

## Environment Details
* GlassFish Version (and build number): 7.0.16 (and older)

----------

## Problem Description

When an application is dropped into the autodeploy directory while GlassFish is not started, GlassFish waits for the polling interval time before it deploys the application. 

## Impact of Issue
This prolongs the time from GlassFish launch to application deployment on startup. it takes longer until the application is deployed at startup and serves requests. This also prolongs the startup time of an application in Docker. By default, the pollingInterval is 2 seconds, so app startup takes 2 seconds longer than needed. Polling interval could be reduced to 1 second

Type: improvement
Component: AutoDeploy

OndroMih:
Fixes #25078.

Before, delay time was the same as the poll interval time.  Now there's no delay. Autodeployment is started immediately and then in polling intervals.

### DIFF
--- a/nucleus/deployment/autodeploy/src/main/java/org/glassfish/deployment/autodeploy/AutoDeployService.java
+++ b/nucleus/deployment/autodeploy/src/main/java/org/glassfish/deployment/autodeploy/AutoDeployService.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
  * Copyright (c) 2008, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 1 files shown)


dmatej (approved):

OndroMih:
I don't know why but it seems that all the 3 tests failed on "Provided hostname could not be resolved to an IP address". It doesn't look like related with changes in this PR.

hs536 (approved):

dmatej (approved):
Retested once more locally and on CI, both passed.

PR #25081, opened 2024-08-08
Add profiles to Faces TCK to run subsets

arjantijms:
This defines 2 main subsets:

* only old tests
* only new tests

Old can be divided into two further sets:

* old1
* old2

### DIFF
--- a/appserver/tests/tck/faces/pom.xml
+++ b/appserver/tests/tck/faces/pom.xml
@@ -36,6 +36,12 @@
         <tck.root>${project.build.directory}</tck.root>
         <maven.executable>${maven.home}/bin/mvn</maven.executable>
         <maven.settings.xml>${user.home}/.m2/settings.xml</maven.settings.xml>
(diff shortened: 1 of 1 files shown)


dmatej:
Bit confusing what means old1 and what means old2 ... and new ... bad names? :-)

dmatej:
I would have to go through tests, but I see they are separated by some `component/html` part. So perhaps `old-html-component` and `old-...`?

arjantijms:
> but I see they are separated by some component/html part. So perhaps old-html-component and old-...?

That's the split for now, but maybe we can find better splits? The parts are still rather big.

dmatej:
> > but I see they are separated by some component/html part. So perhaps old-html-component and old-...?
> 
> That's the split for now, but maybe we can find better splits? The parts are still rather big.

Those should have good names too, but now I will never know if I want `old1` or `old2` or `new`.

arjantijms:
>  but now I will never know if I want old1 or old2 or new.

It's a bit like the other tests where we have "...group1", "...group2":

groups

dmatej:
Yeah, but I call these an army of zombies, its 23 years old set of tests, btw I group will be "cured" before the end of the year! ;-)

PR #25088, opened 2024-08-12
Improvements and fixes in the SQL trace listener functionality

OndroMih:
This fixes a few issues and improves the SQL trace listener feature.

Just to remind, the SQL listener feature allows to implement the `SQLTraceListener` from GlassFish API, package it into a JAR, drop the JAR to domain/lib directory, and configure a connection pool to trigger this listener when SQL is executed. This allows to execute a custom Java code which gets info about SQL executed by the connection pool.

Fixes:

* `SQLTraceRecord` received by a listener always returned `null` from `getPoolName()`

Improvements:

* add `getSqlQuery()` method to `SQLTraceRecord` - returns the executed SQL query if a query was executed (some JDBC driver methods don't execute a query)
* add `getCallingApplicationMethod()` method to `SQLTraceRecord` - returns the info about the last method in the application that triggered the query

Adds test that verify the behavior and also demonstrate usage of the SQL trace listener feature.


nucleus/common/glassfish-api/src/main/java/org/glassfish/api/jdbc/SQLTraceRecord.java
```diff
@@ -194,21 +214,111 @@ public void setParams(Object[] params) {
...
+    /**
+     * Get the stack trace element for the method call in the application that triggered SQL execution.
+     *
+     * This call analyzes the stacktrace of the current thread and finds the first element that
+     * represents a call in the application (the class is loaded by the application classloader).
+     *
+     * @return Stack trace element that represents a call to a server component that triggered SQL execution
+     */
+    public Optional<StackTraceElement> getCallingApplicationMethod() {
+        return Arrays.stream(Thread.currentThread().getStackTrace())
```
avpinchuk, line 281: Why not use `StackWalker` API with the `RETAIN_CLASS_REFERENCE`?

OndroMih: Thank you for the suggestion, I forgot about this API. I redesigned my solution using StackWalker and moved the computation outside of the SQLTraceListener. It means that it will be computed even if the listener doesn't need it. I think it's OK because now the computation is more efficient. It's still computed only if there is at least 1 listener.

nucleus/common/glassfish-api/src/main/java/org/glassfish/api/jdbc/SQLTraceRecord.java
```diff
@@ -194,21 +218,107 @@ public void setParams(Object[] params) {
...
         if (params != null && params.length > 0) {
             int index = 0;
             for (Object param : params) {
-                sb.append("arg[" + index++ + "]=" + (param != null ? param.toString() : "null") + " | ");
+                sb.append("arg[").append(index++).append("]=")
+                        .append(param != null ? param.toString() : "null").append(" | ");
             }
         }
-        // TODO add poolNames and other fields of this record.
+        if (callingApplicationMethod != null) {
```
dmatej, line 317: Just invert the if, so the condition would be without negation. It is easier to read.

appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/connectors/ActiveResourceAdapterImpl.java
```diff
@@ -385,17 +389,22 @@ private void setLogWriter(ManagedConnectionFactory mcf) {
...
 
 
-    protected ManagedConnectionFactory instantiateMCF(String mcfClass, ClassLoader loader) throws Exception {
+    protected ManagedConnectionFactory instantiateMCF(String mcfClassName, ClassLoader loader) throws Exception {
         ManagedConnectionFactory mcf = null;
+        Class<?> mcfClass;
         if (jcl_ != null) {
-            mcf = (ManagedConnectionFactory) jcl_.loadClass(mcfClass).getDeclaredConstructor().newInstance();
+            mcfClass = jcl_.loadClass(mcfClassName);
+//            mcf = (ManagedConnectionFactory) mcfClass.getDeclaredConstructor().newInstance();
```
dmatej, line 397: Don't forget to remove the commented out code.

dmatej (commented):
Seems good, just those two comments, thanks!

dmatej (approved):

avpinchuk (approved):

OndroMih:
The build is failing because of some failures unrelated to the changes in this PR.

hs536 (approved):

PR #25095, opened 2024-08-14
Syncing with master

dmatej:
* Done also some changes which will be different between 7 and 8 (and 8.jdk17) - checks for the JVM version.

PR #25097, opened 2024-08-14
Sync with 8.0 branch 

dmatej:
* Based on #25095 , waiting for that


dmatej:
The test fails because it enables SecurityManager, which is not supported in Jakarta EE 11. So I will remove the test additonally in the 8.0 branch.

PR #25100, opened 2024-08-15
Integrate PFL 5.0.0

arjantijms:
Required patching of MANIFEST.MF of GMBAL and ORB to wide version protection.


dmatej:
I quite hate this patching - we then forget to remove it after updates. I would prefer to align those dependencies and then wait for their release. Also just changing manifest doesn't ensure it would work with JDK23.

arjantijms:
> we then forget to remove it after updates.

That's why there is now a comment next to the place where updating is done. Indeed, we forgot that before.

PR #25102, opened 2024-08-16
Apply proc:none/proc:only for test compile too.

arjantijms:
This is to prepare for the default changing in JDK 23.

See https://inside.java/2024/06/18/quality-heads-up/


appserver/distributions/web/pom.xml
```diff
@@ -71,6 +71,11 @@
             <scope>test</scope>
             <optional>true</optional>
         </dependency>
+        <dependency>
```
dmatej, line 74: We already abandoned Junit4 ;-)

nucleus/parent/pom.xml
```diff
@@ -829,6 +829,26 @@
                                 <proc>only</proc>
                             </configuration>
                         </execution>
+                        <execution>
```
dmatej, line 832: Wouldn't it work with full in default-testcoimpile?
In standard build we have it because it follows some preprocessing if I remember well. Maybe we should try it to see - something should fail ... maybe it can be set even more simple way.

appserver/distributions/web/pom.xml
dmatej: However this is glassfish-distribution packaging and even when you compile those classes, they will not be executed because surefire/failsafe plugins are not in its lifecycle ... or am I wrong?
Still would be better to migrate these tests tu JUnit5 than poisoning build with the obsoleted version.

dmatej: I rechecked - the failsafe plugin is excuted, so just rename the test class to *IT,m use junit5-jupiter-engine as a dependency and use @BeforeAll + @AfterAll etc. from JUnit5. But really don't bring Junit4 back to the project, pleeeease! ;-)
```
        <dependency>
            <groupId>org.junit.jupiter</groupId>
            <artifactId>junit-jupiter-engine</artifactId>
        </dependency>

```

dmatej (changes requested):

nucleus/parent/pom.xml
arjantijms: The default changed in JDK 23, so that when you specify nothing the annotations are not processed.

Therefore the option to do "only"; so the second execution does the annotations. Optionally we can do "full", but then we can only compile on the very latest JDK versions.

appserver/distributions/web/pom.xml
arjantijms: > But really don't bring Junit4 back to the project, pleeeease! ;-)

It was essentially already there :P

appserver/distributions/web/src/test/java/org/glassfish/distributions/test/EmbeddedTest.java
```diff
@@ -65,14 +70,14 @@ public static void setup() throws GlassFishException {
...
-        Assert.assertTrue(listeners.size() == 1);
+        Assertions.assertTrue(listeners.size() == 1);
         for (WebListener listener : listeners) {
             System.out.println("Network listener " + listener.getPort());
         }
 
     }
 
-    @Test
+    @org.junit.jupiter.api.Test
```
dmatej, line 80: You already have it in imports ;)

nucleus/parent/pom.xml
dmatej: You are right here, I did not read the link carefully.

appserver/distributions/web/src/test/java/org/glassfish/distributions/test/EmbeddedTest.java
```diff
@@ -36,12 +37,16 @@
 import org.glassfish.embeddable.web.HttpListener;
 import org.glassfish.embeddable.web.WebContainer;
 import org.glassfish.embeddable.web.WebListener;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
 
 public class EmbeddedTest {
```
dmatej, line 45: Yet don't forget to rename both test classes to *IT.java - then failsafe will find and execute them automatically.

arjantijms: Maybe better to delete the tests. They haven't been executed for like forever, and didn't even compile before.

dmatej (approved):
Ok, I will approve this because we already have tests for embedded in standalone modules and this was quite ugly code.

avpinchuk (approved):

PR #25107, opened 2024-08-20
Fix for a ClassCastException when a filter over a WebSocket endpoint wraps the request

### LINKED ISSUE #24712: WebsocketFilterWrapper casts the wrapped request as a RequestFace, causes classCastException
## Environment Details
* GlassFish Version: Eclipse GlassFish 7.0.11 (commit: ad98c44e896c7a78cb4eb2a84ca9fac450fb3a10)
* JDK version: OpenJDK Runtime Environment Temurin-17.0.7+7 (build 17.0.7+7)
* OS: Windows
* Database: none

----------

## Problem Description
My application uses a programmatically declared Filter to add overrides to a Websocket request before it is processed.
The filter preforms some operation, then creates a wrappedRequest:

MyApplicationFilter.java
```
        public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException, ServletException {
...
		ServletRequest wrappedRequest = new HttpServletRequestWrapper((HttpServletRequest) request) {
                     ... //overrides in here, omitted due to not being relevant
		};
		chain.doFilter(wrappedRequest, response);
        }
...
```

When reaching org.apache.catalina.core.WebSocketFilterWrapper.java further down the Filter chain:
https://github.com/eclipse-ee4j/glassfish/blob/e1f0c9409bdf0c538733de35d6a5336457a3862c/appserver/web/web-core/src/main/java/org/apache/catalina/core/WebSocketFilterWrapper.java#L72

Filter attempts to cast the wrappedRequest as a FaceRequest, causes ClassCastException then fails silently (no error in logs)

## Steps to reproduce
Currently no easy-to-run code sample, but if required I can build something.
Let me know in comment if this is required.

## Impact of Issue
Causes websocket channel initialization to fail silently. No error log, no error message on client-side (just that connection had failed, without an error code or error message)

In order to make the wrappedRequest created in my application filter compatible with WebsocketFilterWrapper, I would need to use FacadeRequest instead of HttpServletRequestWrapper in application code, which I think is unexpected, as RequestFacade is specific to the server implementation, instead of the generic API HttpServletRequestWrapper which should be supported regardless of which environement the application is executed into.

OndroMih:
Fixes https://github.com/eclipse-ee4j/glassfish/issues/24712.

Makes the cast to RequestFacade more robust - only if cast is allowed, searches through the chain of wrapped requests.

Added tests for the case fixed in #24691 and also for the case described in #24712.

### DIFF
--- a/appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/GlassFishTestEnvironment.java
+++ b/appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/GlassFishTestEnvironment.java
@@ -149,8 +149,7 @@ public class GlassFishTestEnvironment {
 
 
     public static int getPort(HttpListenerType listenerType) {
-        return Integer.parseInt(System.getenv()
-                .getOrDefault(listenerType.getVariableName(), String.valueOf(listenerType.getDefaultPort())));
+        return listenerType.getPort();
     }
     /**
      * Creates a {@link Client} instance for the domain administrator.
--- a/appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/GlassFishTestEnvironment.java
+++ b/appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/GlassFishTestEnvironment.java
@@ -24,6 +24,8 @@ import java.io.InputStream;
 import java.net.Authenticator;
 import java.net.HttpURLConnection;
 import java.net.PasswordAuthentication;
+import java.net.URI;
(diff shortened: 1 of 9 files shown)


hs536 (approved):
LGTM

dmatej (approved):

dmatej:
I have rebased the PR and resolved the conflict. Also note that there are two tests failing on JDK23 and this is one of them. It is not important for the PR, just to know ...

PR #25132, opened 2024-09-06
Proxy creation improvements for JPMS, remove the need for --add-opens in some cases

OndroMih:
Raising as a draft, because this improves only a few places. Hopefully all tests pass and we can add more improvements gradually.

In short, use JPMS compatible MethodHandles.Lookup instead of calling ClassLoader.defineClass by reflection. This approach has some limitations. We need a class to be used as a template (the generated class will have the same classloader and protection domain derived from it). And the generated class must have the same package name as the template class. 

For Weld proxies, I decided to generate a class with a different name than Weld requests, because the generated class must be in the same package as the anchor class and Weld requests a different package name for the generated classes. It looks like Weld doesn't care and all worked on my sample app with EJBs.

Ideally, all methods in `ClassGenerator` that I marked as deprecated, will be removed and callers will be refactored to use the non-deprecated methods. The deprecated methods use reflection incompatible with JPMS, the other ones use the JPMS compatible MethodHandles. There are still a lot of places, though, where the deprecated methods are used, so it will take some time.

Yet to refactor:
- `ProxyServicesImpl.classPoolMap` should be moved to a global service rather than be a static field. It must be global because `ProxyServicesImpl` is created for each app deployment, while the classloader where the classes are generated, is shared. So the cache needs to be preserved, otherwise we'd attempt to create a new class that already exists in the classloader.


OndroMih:
Hi @dmatej, @avpinchuk, @pzygielo, could you please review this and tell me whether this is a viable approach to make the code compatible with JPMS?

avpinchuk:
I did not any experiments yet how GF behaves with your changes (but will do soon ;) ).

But I have a question. What happens when we redeploy application many times? Generated classes unloaded or stay loaded (as a garbage, actually) in the bootstrap or server CL?

mz1999:
My understanding is that the ideal goal here is to eliminate the need for `--add-opens=java.base/java.lang=ALL-UNNAMED` altogether.  And if all defineClass calls were using the anchor-class based lookup.defineClass – the JPMS-compatible way – that would indeed be the case.

However, as it stands, there are still many places relying on ClassLoader.defineClass. So, I see this PR introduces a pragmatic compromise with lazy loading.  This is smart, as it means that if ClassLoader.defineClass isn't actually triggered at runtime, we won't immediately hit errors even without --add-opens.  It's a nice improvement in that users might not need --add-opens from the get-go, and only add it if they actually run into problems.

My question is, is the eventual plan to migrate all usages of ClassLoader.defineClass to lookup.defineClass? Or are there scenarios where switching to lookup.defineClass isn't feasible?  Just curious about the roadmap here.

Anyway, this looks like valuable progress.

OndroMih:
Hi @mz1999 ,

> are there scenarios where switching to lookup.defineClass isn't feasible?

I don’t know. Currently, there are some failed tests related to remote EJBs, which our team at OmniFish haven’t time to look into. This is more like a proof of concept than a plan. We try to replace usages of ClassLoader.defineClass with lookup.defineClass continually, with other improvements. This PR is an effort to replace rhem everywhere but it’s not completed yet.

The problem is that ClassLoader.defineClass, as it is used in many places in GlassFish, has different semantics compared to lookup.defineClass. In most cases, it defines a class in a different package than the original class it wraps. This is not allowed with lookup.defineClass, which requires that the new class is in the same package. And this is probably what dails the remote EJB tests. This also changes packages of CDI proxy classes. We need to investigate what is the impact of the package change and how to fix all issues with it.

mz1999:
Thanks for the update!  Sounds like a big job, lots to do.

I'd be happy to contribute and have some time available to assist if you think it would be helpful.

About the failed remote EJB tests, are they unit tests or TCK?  Knowing that would help me try to reproduce it locally and investigate.

OndroMih:
> About the failed remote EJB tests, are they unit tests or TCK?

Well, they are integration tests in the GlassFish project. You can run the failed test in the GlassFish source root with:

To build GlassFish without running tests:

```
mvn clean install -Pfastest -T4C
```

To build common test artifacts:

```
mvn -pl :tests-embedded -am clean install -DskipTests
```

To run the failed test:

```
mvn -pl :tests-embedded -B -e clean install -Pstaging,qa -Dit.test=org.glassfish.main.extras.embedded.test.all.deployment.webapp.RemoteWebAppDeploymentITest
```

mz1999:
I ran the test case `org.glassfish.main.extras.embedded.test.all.deployment.webapp.RemoteWebAppDeploymentITest` using the branch [OndroMih:ondromih-jpms-defineClass](https://github.com/OndroMih/glassfish/tree/ondromih-jpms-defineClass), and encountered a `ClassCastException`. This exception is thrown by `PortableRemoteObject.narrow` in the [EJBUtils.lookupRemote30BusinessObject](https://github.com/OndroMih/glassfish/blob/3e8556316990229f141e628eaaf742a8a10bce74/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/EJBUtils.java#L283) method:

```java
public static Object lookupRemote30BusinessObject(Object jndiObj, String businessInterface) throws NamingException {
		...
        Class<?> genericEJBHome = loadGeneratedGenericEJBHomeClass(loader, jndiObj.getClass());
        final Object genericHomeObj = PortableRemoteObject.narrow(jndiObj, genericEJBHome);
        ...
}
```

Let's briefly analyze how this `ClassCastException` occurs:

1. The internal rule of the `loadGeneratedGenericEJBHomeClass` method is to generate `genericEJBHome` based on the package name of `jndiObj.getClass()`. This means both classes belong to the same package. Since `jndiObj.getClass()` returns `com.sun.corba.ee.impl.corba.CORBAObjectImpl`, the generated `genericEJBHome` becomes `com.sun.corba.ee.impl.corba.GenericEJBHome_Generated`.

2. Subsequently, `PortableRemoteObject.narrow` attempts to cast `jndiObj` to the type `genericEJBHome`. However, the actual type that `jndiObj` "represents" is `org.glassfish.main.extras.embedded.test.app.ejb.GenericEJBHome_Generated`. Although the class names are the same as `genericEJBHome`, they belong to different packages. This package mismatch causes the casting to fail, resulting in the `ClassCastException`.

The root cause is that `genericEJBHome` is generated based on `jndiObj.getClass()`, and they end up in the same package, which is `com.sun.corba.ee.impl.corba`.  Meanwhile, the type `jndiObj` actually "represents" is `org.glassfish.main.extras.embedded.test.app.ejb.GenericEJBHome_Generated`, which is in the same package as the application's EJB.  Furthermore, `jndiObj`'s package can vary depending on the deployed application.

Based on this error, the idea for a fix is to change the `anchorClass` for `genericEJBHome`. Since `genericEJBHome` needs to be generated based on an `anchorClass`, we can use a class related to the application's EJB as the `anchorClass`. This way, the generated `genericEJBHome` will be in the same package as the EJB and thus in the same package as the type `jndiObj` actually "represents".  This should allow `narrow` to work correctly.

In the `lookupRemote30BusinessObject` method, the `businessInterface` class is also loaded. Therefore, we can use this class as the `anchorClass`. The modification would be:

```java
public static Object lookupRemote30BusinessObject(Object jndiObj, String businessInterface) throws NamingException {

	Class<?> businessInterfaceClass = loadGeneratedRemoteBusinessClasses(businessInterface);
	...
        Class<?> genericEJBHome = loadGeneratedGenericEJBHomeClass(loader, businessInterfaceClass);
        final Object genericHomeObj = PortableRemoteObject.narrow(jndiObj, genericEJBHome);
        ...
}
```

I've verified that this modification allows the test case to pass. However, the EJB part is quite complex, and I don't fully understand it. I'm just making guesses and modifications based on debugging information, so I'm not sure if this modification is entirely complete or correct.

PR #25134, opened 2024-09-06
Refactoring around the `glassfish` module

dmatej:
* Removed unused code
* JDepend is happy - no cyclic dependencies between packages
* `static` renamed to `embedded`, however static had to remain in code for the embedded-glassfish-maven-plugin which uses hardcoded `static` value; it is marked as deprecated now.
* `*Static*` classes renamed to Embedded
* Anonymous or private classes moved to own files

### Tests
* Passed SpotBugs rules (without exclusions)
* TCK passed for glassfish which came out of build https://github.com/eclipse-ee4j/glassfish/issues/1:
  * https://ci.eclipse.org/jakartaee-tck/job/10/job/eftl-jakartaeetck-run-100/245/parameters/
  * https://ci.eclipse.org/jakartaee-tck/job/jakartaee-tck/job/10.0.x/143/

### Review 
* basically diff makes nearly no sense, rather open the implementation and try to find issues. It should be much easier than before.


nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/osgi/Jar.java
```diff
@@ -85,47 +83,56 @@ public Jar(URI uri) {
...
         return uri == null ? 0 : uri.hashCode();
     }
 
-    public boolean equals(Object obj)
-    {
-        if (obj == null || !(obj instanceof Jar)) return false;
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == null || !(obj instanceof Jar)) {
```
avpinchuk, line 121: `obj == null` check redundant here

nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/cfg/AsenvConf.java
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * The asenv.conf file content.
+ */
+public class AsenvConf {
+
+    private static final Pattern LINE = Pattern.compile("(?i)(set +)?([^=]*)=\"?([^\"]*)\"?");
+
+    private final Properties properties;
+
+    public AsenvConf(File asenv, File configDir) {
```
avpinchuk, line 40: Maybe `private` or package-private scope for ctor?

nucleus/core/bootstrap/pom.xml
```diff
@@ -129,6 +129,15 @@
                     </archive>
                 </configuration>
             </plugin>
```
avpinchuk, line 131: Missing license header update )

appserver/extras/embedded/common/bootstrap/src/main/java/org/glassfish/uberjar/bootstrap/UberJarOSGiGlassFishRuntimeBuilder.java
```diff
@@ -161,24 +161,15 @@ public GlassFishRuntime build(BootstrapProperties bsOptions) throws GlassFishExc
         bsOptions.setProperty(AUTO_START_BUNDLES_PROP, autoStartBundleLocation);
         System.setProperty(AUTO_START_BUNDLES_PROP, autoStartBundleLocation);
 
-        System.setProperty(Constants.INSTALL_ROOT_PROP_NAME, installRoot);
-        System.setProperty(Constants.INSTANCE_ROOT_PROP_NAME, instanceRoot);
+        System.setProperty(BootstrapKeys.INSTALL_ROOT_PROP_NAME, installRoot);
```
avpinchuk, line 164: Maybe better use static import for consistency with other code?

nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/cfg/ServerFiles.java
```diff
@@ -0,0 +1,88 @@
...
+     * @param envKey Key to use with {@link System#getenv(String)}
+     * @param sysPropsKey Key to use with {@link System#getProperty(String)}
+     * @param defaultSubdir Relative path to {@link #getInstanceRoot()}
+     * @return first configured directory.
+     * @throws IllegalArgumentException if the directory does not exist.
+     */
+    public File getOsgiHome(String envKey, String sysPropsKey, Path defaultSubdir) {
+        final String envProperty = System.getenv(envKey);
+        if (envProperty != null) {
+            return ServerFiles.toExistingFile(envProperty);
```
avpinchuk, line 64: Redundant `ServerFiles` prefix

nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/cfg/ServerFiles.java
```diff
@@ -0,0 +1,88 @@
...
+        }
+        // try system property, which comes from asenv.conf
+        final String sysProperty = System.getProperty(sysPropsKey);
+        if (sysProperty != null) {
+            return ServerFiles.toExistingFile(sysProperty);
```
avpinchuk, line 69: Here too

avpinchuk (commented):
This is a quick look ;)

nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/GlassFishImpl.java
```diff
@@ -30,31 +32,41 @@
...
-
 public class GlassFishImpl implements GlassFish {
 
+    private static final String CONFIG_PROP_PREFIX = "embedded-glassfish-config.";
+
     private ModuleStartup gfKernel;
-    private ServiceLocator habitat;
-    volatile Status status = Status.INIT;
+    private ServiceLocator serviceLocator;
+    private volatile Status status = Status.INIT;
```
avpinchuk, line 41: Please move this initialization into ctor after `serviceLocator` assignment. This will guarantee thread visibility of the `serviceLocator` and `gfKernel` values.

nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/cfg/StartupContextUtil.java
```diff
@@ -36,46 +39,70 @@ public final class StartupContextUtil {
...
+        // glassfish/modules/glassfish.jar
+        File bootstrapFile = findBootstrapFile();
+        // glassfish/
+        return bootstrapFile.getParentFile().getParentFile();
+    }
+
+
+    private static File findBootstrapFile() {
+        try {
+            return Which.jarFile(StartupContextUtil.class);
```
avpinchuk, line 56: In the original code used `ASMain.class`. Is not this a mistake or typo?

nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/embedded/EmbeddedGlassFishRuntimeBuilder.java
```diff
@@ -42,12 +47,12 @@
...
+// Note: Used in a service file!
+public class EmbeddedGlassFishRuntimeBuilder implements RuntimeBuilder {
 
-public class StaticGlassFishRuntimeBuilder implements RuntimeBuilder {
-
-    private static Logger logger = Util.getLogger();
+    private static final Logger LOG = LogFacade.BOOTSTRAP_LOGGER;
     private static final String JAR_EXT = ".jar";
-    final List<String> moduleExcludes = Arrays.asList("jsftemplating.jar", "gf-client-module.jar");
+    private static final List<String> MODULE_EXCLUDES = Arrays.asList("jsftemplating.jar", "gf-client-module.jar");
```
avpinchuk, line 55: `List.of(...)`?

nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/embedded/JarUtil.java
```diff
@@ -38,41 +40,41 @@
...
      * JMS datasource  system resource adapter name.
      */
-    public static final String DEFAULT_JMS_ADAPTER = "jmsra";
+    private static final String DEFAULT_JMS_ADAPTER = "jmsra";
 
-    public static final String RAR_EXTENSION = ".rar";
+    private static final String RAR_EXTENSION = ".rar";
 
-    public static final List<String> systemRarNames = Collections.unmodifiableList(
+    private static final List<String> SYSTEM_RAR_NAMES = Collections.unmodifiableList(
```
avpinchuk, line 77: `List.of(...)` instead of `Collections.unmodifiableList(...)`?

avpinchuk:
LGTM except for the above shortcomings.

nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/GlassFishImpl.java
dmatej: I moved that to the constructor, but as the first line. It doesn't matter, if constructor fails, its fields will not be used.

nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/cfg/StartupContextUtil.java
dmatej: It uses any class from the jar file to locate the jar file, so I used the closest one instead of making dependency cycle between packages. I will add javadoc for sure.

nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/GlassFishImpl.java
avpinchuk: In this case `start()` method may throw NPE if called from another thread. Field initialization done after `super()` but before fields assignment in constructor. We have next sequence: init `gfKernel` with default value (null) -> volatile write `state`-> `gfKernel` assignment -> volatile read `state` in `start()` method -> volatile write `state` -> plain read `gfKernel` (at the line 75). This assignment does not cover by the happens-before relation. And another thread can see null. Synchronized on `start` does not matters because `gfKernel` initialized without synchronization.

hs536 (approved):

nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/GlassFishImpl.java
dmatej: I am not sure if we understand each other, so maybe silly question - does it apply for the current code, when I moved it to the constructor? I believe then all three fields should be initialized in order.
I understand and absolutely agree with you about initialization outside constructor. Btw on other my projects tools prohibit calling nonstatic methods of the same instance from constructor - for similar reasons.

dmatej: Is it possible to use an object with unfinished constructor in another thread? How could that happen? The only such scenario I can imagine is if we would start thread using `this` in constructor.
I really want to get it.

dmatej: Aha, seems I forgot something important again ... you are right, thanks!
https://stackoverflow.com/a/10528614/375449

avpinchuk (approved):

PR #25145, opened 2024-09-15
Executable GlassFish Embedded in GlassFish Server installation

OndroMih:
This makes GlassFish Embedded static shell in the GlassFish Server installation, located at `glassfish/lib/embedded/glassfish-embedded-static-shell.jar`, equal to the classical GlassFish Embedded JAR. The UberMain class is the main class of the JAR and the GlassFish bootstrap JARs are added to the classpath.

The GlassFish Embedded static shell JAR can be treated as glassfish-embedded-all.jar - it can be added to the classpath and run GlassFish Embedded apps, or can be started as an executable JAR with `java -jar glassfish/lib/embedded/glassfish-embedded-static-shell.jar`.

This is a preparation for running GlassFish Embedded from the GlassFish Server Docker image. I'm working on adding more features to the executable JAR to allow configuring it and running applications from command line.


OndroMih:
Closing. I will raise this together with the new runnable GF Embedded which I'm working on, there will be some changes.

PR #25147, opened 2024-09-17
Fixed insecure XML processing - all XSDs and DTD have to be local

dmatej:
* See also
  *  https://find-sec-bugs.github.io/bugs.htm#XXE_DTD_TRANSFORM_FACTORY
  * https://docs.stackhawk.com/vulnerabilities/90023/
  * https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html
* We use just local XSD and DTD files here 

Additional changes:
* It is better to close the stream where it was open, symmetrically (try-with)

### DIFF
--- a/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/stringsubs/impl/StringSubstitutionParser.java
+++ b/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/stringsubs/impl/StringSubstitutionParser.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation.
  * Copyright (c) 2013, 2018-2021 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -68,6 +69,9 @@ public class StringSubstitutionParser {
             JAXBContext context = JAXBContext.newInstance(StringsubsDefinition.class.getPackage().getName());
             Unmarshaller unmarshaller = context.createUnmarshaller();
             SchemaFactory schemaFactory = SchemaFactory.newInstance(XMLConstants.W3C_XML_SCHEMA_NS_URI);
+            schemaFactory.setProperty(XMLConstants.ACCESS_EXTERNAL_DTD, "");
+            schemaFactory.setProperty(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "file");
+            schemaFactory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
             Schema schema = schemaFactory.newSchema(schemaUrl);
             unmarshaller.setSchema(schema);
             InputSource is = new InputSource(configStream);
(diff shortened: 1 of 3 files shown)


nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/template/TemplateInfoHolder.java
```diff
@@ -67,29 +67,22 @@ public String getLocation() {
...
-            } catch (IOException e) {
-                /** Ignore */
-            }
+            throw new DomainException("Invalid stream: null");
         }
+        URL schemaUrl = TemplateInfoHolder.class.getClassLoader().getResource(TEMPLATE_INFO_SCHEMA_PATH);
+        JAXBContext context = JAXBContext.newInstance(TemplateInfo.class.getPackage().getName());
+        Unmarshaller unmarshaller = context.createUnmarshaller();
+        SchemaFactory schemaFactory = SchemaFactory.newInstance(XMLConstants.W3C_XML_SCHEMA_NS_URI);
+        schemaFactory.setProperty(XMLConstants.ACCESS_EXTERNAL_DTD, "");
```
avpinchuk, line 78: External entities not allowed ony here. Is it correct?

dmatej: I remember that also disabling some protocols helps, and GlassFish here should have all dependencies locally available. I remember some old versions were missing something and we discovered it on production which did not have access to the internet. This should cause that some tests would fail.
External entities are disabled by the "secure processing feature".
I think SpotBugs will report also other places, but I am going per module.
I will add more references.

dmatej: Hmm, however it seems that in this case we can disable both. I will recheck the second change.

nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/stringsubs/impl/StringSubstitutionParser.java
```diff
@@ -68,6 +69,9 @@ public static StringsubsDefinition parse(InputStream configStream) throws String
             JAXBContext context = JAXBContext.newInstance(StringsubsDefinition.class.getPackage().getName());
             Unmarshaller unmarshaller = context.createUnmarshaller();
             SchemaFactory schemaFactory = SchemaFactory.newInstance(XMLConstants.W3C_XML_SCHEMA_NS_URI);
+            schemaFactory.setProperty(XMLConstants.ACCESS_EXTERNAL_DTD, "");
+            schemaFactory.setProperty(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "file");
```
dmatej, line 73: Here we need the file protocol, because `stringsubs.xsd` includes `nons_stringsubs.xsd` using `xsd:include`

avpinchuk (approved):

PR #25149, opened 2024-09-17
SpotBugs based cleanup in nucleus-admin and surroundings

dmatej:
### Review

I recommend to review each commit, 
Just one note before you notice - the commit "Fixed parsing charset from ContentType..." fixes mistake I made in "SpotBugs-based cleanup in security module" where I tried to parse Charset from the whole content type header, not just the charset parameter.

### Changes

* Mostly explicitly declaring Charse.defaultCharset where using STDIN, STDOUT or STDERR, or respecting contentType header or using UTF-8 as a default (not defaultCharset as we are communicating with other servers with other settings and even local can change between users - or even user can switch his own, encoded files then might stop working).
* Charset is a feature of a file, not of a user.
* Refactoring - when I was fixing some incorrect copy pasted code for fifth time, I decided to move it somewhere.
* When I noticed 1K buffer, I checked how large data it usually processes; mostly I increased the size of the buffer to 8K. Btw Files.readAllBytes uses 8K as a default.
* Some trivial changes were made automatically by an editor - generics, added finals. Sometimes there could be done more, but I did not want to increase the size of this PR too much. Maybe later with adding Checkstyle rules.
* **Bugs** - I noticed several bugs and I fixed them, commits with their fixes are very small, so you will see in the review.


arjantijms (approved):

arjantijms:
Just spot checked, but lgtm

hs536 (approved):

PR #25153, opened 2024-09-22
Dynamically update classloader after adding a library with the add-library command

OndroMih:
Dynamically updates the common classloader if a common library is added. It means that restart is not needed to add the library to the classpath.

### DIFF
--- a/nucleus/core/kernel/pom.xml
+++ b/nucleus/core/kernel/pom.xml
@@ -260,6 +260,56 @@
                     </archive>
                 </configuration>
             </plugin>
+            <plugin>
+                <artifactId>maven-assembly-plugin</artifactId>
+                <version>3.7.1</version>
+                <executions>
+                    <execution>
+                        <id>commonClassLoader tests</id>
+                        <phase>process-test-classes</phase>
+                        <goals>
+                            <goal>single</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <inlineDescriptors>
+                        <inlineDescriptor>
+                            <id>domain</id>
+                            <includeBaseDirectory>false</includeBaseDirectory>
+                            <formats>
+                                <format>dir</format>
+                            </formats>
+                            <fileSets>
+                                <fileSet>
+                                    <outputDirectory>lib/classes</outputDirectory>
+                                    <includes>
+                                        <include>**/CommonClassLoaderServiceImplTestDomainClass.class</include>
+                                    </includes>
+                                    <directory>target/test-classes/</directory>
+                                </fileSet>
+                            </fileSets>
+                        </inlineDescriptor>
+                        <inlineDescriptor>
+                            <id>additional-classes</id>
+                            <includeBaseDirectory>false</includeBaseDirectory>
+                            <formats>
+                                <format>dir</format>
+                            </formats>
+                            <fileSets>
+                                <fileSet>
+                                    <outputDirectory>/</outputDirectory>
+                                    <includes>
+                                        <include>**/CommonClassLoaderServiceImplTestAdditionalClass.class</include>
+                                    </includes>
+                                    <directory>target/test-classes/</directory>
+                                </fileSet>
+                            </fileSets>
+                        </inlineDescriptor>
+                    </inlineDescriptors>
+                    <finalName>test</finalName>
+                </configuration>
+            </plugin>
         </plugins>
     </build>
 </project>
--- a/nucleus/core/kernel/pom.xml
+++ b/nucleus/core/kernel/pom.xml
@@ -31,6 +31,10 @@
     <packaging>glassfish-jar</packaging>
 
     <name>Kernel Classes</name>
+    
+    <properties>
+        <commonClassLoader.tests.skip>false</commonClassLoader.tests.skip>
+    </properties>
 
     <developers>
         <developer>
@@ -273,6 +277,7 @@
                     </execution>
                 </executions>
                 <configuration>
+                    <skipAssembly>${commonClassLoader.tests.skip}</skipAssembly>
                     <inlineDescriptors>
                         <inlineDescriptor>
                             <id>domain</id>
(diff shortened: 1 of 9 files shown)


nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/CommonClassLoaderServiceImpl.java
```diff
@@ -155,10 +153,32 @@ public ClassLoader getCommonClassLoader() {
         return commonClassLoader != null ? commonClassLoader : APIClassLoader;
     }
 
+    /**
+     * Adds a classpath element to the common classloader if the classloader supports it.
+     * @param url URL of the classpath element to add
+     * @throws UnsupportedOperationException If adding not supported by the classloader
+     */
+    public void addToClassPath(URL url) {
+        if (commonClassLoader != null) {
```
dmatej, line 162: Just invert the condition, `== null` should be first.

dmatej:
Is it possible also to remove the library? I see there is the `asadmin remove-library` command too, however I have no idea how it works if the application is already running and using the library.

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/CommonClassLoaderServiceImpl.java
```diff
@@ -155,10 +153,32 @@ public ClassLoader getCommonClassLoader() {
...
+            commonClassLoader = new GlassfishUrlClassLoader(new URL[] {url}, APIClassLoader);
+        }
+        commonClassPath = urlsToClassPath(Arrays.stream(commonClassLoader.getURLs()));
+    }
+
     public String getCommonClassPath() {
         return commonClassPath;
     }
 
+    private static String urlsToClassPath(Stream<URL> urls) {
```
pzygielo, line 174: What a great candidate to have unit test for!

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/CommonClassLoaderServiceImpl.java
```diff
@@ -155,10 +153,32 @@ public ClassLoader getCommonClassLoader() {
...
+        return urls
+                .map(URL::getFile)
+                .filter(Predicate.not(String::isBlank))
+                .map(file -> new File(file).getAbsolutePath())
```
pzygielo, line 178: Could the lambda be replaced with additional `.map`?

avpinchuk:
> Is it possible also to remove the library? I see there is the `asadmin remove-library` command too, however I have no idea how it works if the application is already running and using the library.

Perhaps, on Linux (and Unix generally) library file will be removed, but library will be accessible until server stop/restart. On Windows `remove-library` command should be failed due to file locking.

OndroMih:
The remove-library command just deletes the JAR on the file system. It could remove the url in the classloader but it’s not necessary. The CL will just ignore the url if the file doesn’t exist. So no need to remove the url from the CL. Classes loaded from the removed JAR will keep being loaded even after the JAR is removed.

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/CommonClassLoaderServiceImpl.java
OndroMih: You mean have 2 map calls instead of just one? It could but I thought it’s an unnecessary overhead to call map twice.

avpinchuk:
> The CL will just ignore the url if the file doesn’t exist

If JAR was already opened by CL, CL will not ignore this JAR. Application will load classes and resources from this JAR.

dmatej:
We already have this command, so it should probably do more, ie. refuse to remove the file if it is used by any classloader and report that classloader to the user. We can try to do that in another PR, but for this reason I always pushed libraries inside ear/war files, because they have lifecycle while this is quite fragile.

Could it be enough to disable the application and then remove the library? Maybe we can move this to discussions.

avpinchuk:
> Maybe we can move this to discussions

Good idea ;)

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/CommonClassLoaderServiceImpl.java
pzygielo: > You mean have 2 map calls instead of just one? It could but I thought it’s an unnecessary overhead to call map twice.

Yes, yes.

2 map calls instead of one, and no lambda instead of one.

OndroMih:
I used a wrong phase for the assembly plugin. It's supposed to copy test classes before classloader tests are executed. Before the fix, it was executed before test classes were compiled, it tried to move non-existent classes and failed.

OndroMih:
OK, the fastest profile disabled test compilation, so I needed to disable the assembly plugin in the fastest profile too.

nucleus/core/kernel/src/test/java/com/sun/enterprise/v3/server/CommonClassLoaderServiceImplTest.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    @Test
+    public void testAddingUrlWithNoInitialUrls() throws MalformedURLException, ClassNotFoundException {
+        commonCLService.postConstruct();
+
+        final String classesPath = "target/test-additional-classes/";
+        commonCLService.addToClassPath(new File(classesPath).toURI().toURL());
+
+// we need to retrieve the classloader after adding URLs, otherwise we
```
dmatej, line 63: Is this aligned intentionally this way?

nucleus/core/kernel/pom.xml
```diff
@@ -260,6 +264,66 @@
                     </archive>
                 </configuration>
             </plugin>
+            <plugin>
+                <artifactId>maven-assembly-plugin</artifactId>
+                <version>3.7.1</version>
```
dmatej, line 269: Just remove the version, it is already set in parents.

dmatej (approved):
If we want to hurry to merge it to 7.0.18, those two comments can be resolved in another PR (or I can add it to mine as I am messing with classloaders too)

PR #25160, opened 2024-09-25
Fallback to basic file attributes if sun.nio.fs cannot be used

OndroMih:
setAccessible throws InaccessibleObjectException if not allowed.  The generic catch block is just in case - no other exception should be thrown. if it is thrown, we should revert to fallback rather than blow up.

[//]: For general information, please read <"https://github.com/eclipse-ee4j/glassfish/blob/master/CONTRIBUTING.md">.

[//]: DO NOT create a PR for the gh-pages branch. Instead, please read the docs directory README file <"https://github.com/eclipse-ee4j/glassfish/blob/master/docs/README.md"> and create a PR for the appropriate docs subdirectory in the master branch.

### DIFF
--- a/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/AppLibClassLoaderServiceImpl.java
+++ b/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/AppLibClassLoaderServiceImpl.java
@@ -52,6 +52,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Supplier;
 import java.util.jar.JarFile;
 
 import org.glassfish.api.admin.ServerEnvironment;
@@ -423,11 +424,17 @@ public class AppLibClassLoaderServiceImpl implements EventListener {
                 }
                 method.setAccessible(true);
                 field.setAccessible(true);
-            } catch (ReflectiveOperationException | InaccessibleObjectException e) {
+            } catch (InaccessibleObjectException e) {
+                LOG.log(WARNING, () -> "Could not access specialized 'sun.nio.fs' APIs. Try opening the package with the JVM argument '--add-opens=java.base/sun.nio.fs=ALL-UNNAMED' to the JVM options. We will now fallback to using basic file attributes. Error: " + e.getMessage());
                 method = null;
                 field = null;
             } catch (Exception e) {
-                LOG.log(WARNING, () -> "Could not initialize sun.nio.fs. We will fallback to basic file attributes. Error: " + e.getMessage(), e);
+                Supplier<String> messageSupplier = () -> "Could not initialize sun.nio.fs. We will now fallback to using basic file attributes. Error: " + e.getMessage();
+                if (e instanceof ReflectiveOperationException) {
+                    LOG.log(WARNING, messageSupplier);
+                } else {
+                    LOG.log(WARNING, messageSupplier, e);
+                }
                 method = null;
                 field = null;
             }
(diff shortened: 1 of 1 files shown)


nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/AppLibClassLoaderServiceImpl.java
```diff
@@ -422,7 +423,11 @@ private static class Library {
                 }
                 method.setAccessible(true);
                 field.setAccessible(true);
-            } catch (ReflectiveOperationException e) {
+            } catch (ReflectiveOperationException | InaccessibleObjectException e) {
```
avpinchuk, line 426: I think, this catch block is redundant in this case.

OndroMih: It’s definitely needed. The setAccessible method throws InaccessibleObjectException if sun.nio.fs is not opened in Java17+. 

And InaccessibleObjectException doesn’t extend ReflectiveOperationException.

avpinchuk: Both `InaccessibleObjectException` and `ReflectiveOperationException` extends `Exception`. This:

```java
catch (Exception e) {
    LOG.log(WARNING, () -> "Could not initialize sun.nio.fs. We will fallback to basic file attributes. Error: " + e.getMessage(), e);
    method = null;
    field = null;
```

covers all cases and is sufficient. Or not?

avpinchuk: > However, I'm not sure whether we need to call the sun nio methods? The readAttributes method is called only from the equals method. Is it needed to have all available attributes? Are basic attributes not enough to compare libraries?

Main purpose of sun.nio.* methods - work with descriptor of the open file if possible. This prevents inconsistency if library was replaced between `readAttributes` and snapshot creation.

avpinchuk: > InaccessibleObjectException happens because of the JPMS, and is kind of expected if add-opens is missing. The add-opens for sun nio can be missing simply because people upgrade to a newer version of GlassFish and keep their domain. I thought it's not appropriate to log the stacktrace.

Perhaps, we should give the user some hint about misconfiguration.

OndroMih: OK, I have a new versoin now:
* `InaccessibleObjectException` logs a WARNING with instructions to add --add-opens, no stacktrace - it's clear how to fix this, no need for garbage in logs
* `ReflectiveOperationException` logs a WARNING, no stacktrace, can only happen if the underlying JDK doesn't contain the expected classes - shouldn't happen, if it does, it's clear what is the issue
* another exception - also print stacktrace

avpinchuk (approved):

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/AppLibClassLoaderServiceImpl.java
```diff
@@ -422,7 +424,17 @@ private static class Library {
...
                 field.setAccessible(true);
-            } catch (ReflectiveOperationException e) {
+            } catch (InaccessibleObjectException e) {
+                LOG.log(WARNING, () -> "Could not access specialized 'sun.nio.fs' APIs. Try opening the package with the JVM argument '--add-opens=java.base/sun.nio.fs=ALL-UNNAMED' to the JVM options. We will now fallback to using basic file attributes. Error: " + e.getMessage());
+                method = null;
+                field = null;
+            } catch (Exception e) {
+                Supplier<String> messageSupplier = () -> "Could not initialize sun.nio.fs. We will now fallback to using basic file attributes. Error: " + e.getMessage();
+                if (e instanceof ReflectiveOperationException) {
+                    LOG.log(WARNING, messageSupplier);
```
pzygielo, line 434: What would be logged in that case?

hs536 (approved):

PR #25161, opened 2024-09-25
Enhance start-db command - initdb and suspend

arjantijms:
This adds a feature to the start-database command to initialize the DB using a file with SQL statements. This is roughly equivalent to the feature in Jakarta Persistence to initialize a datasource.

It also adds a feature to suspend the process running Derby, so, if needed, starting the DB and the SQL execution can be debugged.

### DIFF
--- a/appserver/admin/cli-optional/src/main/java/com/sun/enterprise/admin/cli/optional/DerbyControl.java
+++ b/appserver/admin/cli-optional/src/main/java/com/sun/enterprise/admin/cli/optional/DerbyControl.java
@@ -55,7 +55,7 @@ public final class DerbyControl {
     public static void main(String[] args) {
         if (args.length < 3) {
             System.out.println("paramters not specified.");
-            System.out.println("DerbyControl <derby command> <derby host> <derby port> <redirect> [<derby home> | <derby user> <derby password>");
+            System.out.println("DerbyControl <derby command> <derby host> <derby port> <redirect> [<derby home> | <derby user> <derby password>]");
             System.exit(1);
         }
 
--- a/appserver/admin/cli-optional/src/main/java/com/sun/enterprise/admin/cli/optional/DerbyControl.java
+++ b/appserver/admin/cli-optional/src/main/java/com/sun/enterprise/admin/cli/optional/DerbyControl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Contributors to the Eclipse Foundation
+ * Copyright (c) 2021, 2024 Contributors to the Eclipse Foundation
  * Copyright (c) 1997, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -55,7 +55,7 @@ public final class DerbyControl {
     public static void main(String[] args) {
         if (args.length < 3) {
             System.out.println("paramters not specified.");
-            System.out.println("DerbyControl <derby command> <derby host> <derby port> <derby home> <redirect output>");
+            System.out.println("DerbyControl <derby command> <derby host> <derby port> <redirect> [<derby home> | <derby user> <derby password>");
             System.exit(1);
         }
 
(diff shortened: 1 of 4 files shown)


appserver/admin/cli-optional/src/main/java/com/sun/enterprise/admin/cli/optional/DerbyExecuteSQL.java
```diff
@@ -0,0 +1,104 @@
...
+                                connection.prepareStatement(statement).execute();
+                            }
+                        } catch (SQLException s) {
+                            s.printStackTrace();
+                        }
+                    }
+                }
+            }
+        } catch (Throwable e) {
+            e.printStackTrace();
```
dmatej, line 99: Wouldn't be better to throw it out? It should crash when it has broken script ...

arjantijms: It should not crash, although we can add an option to crash or not.

Scripts very often do a "drop table", and don't have logic to do "drop table if not exist".

appserver/admin/cli-optional/src/main/java/com/sun/enterprise/admin/cli/optional/DerbyExecuteSQL.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    private void executeCommand() {
+        try {
+            Driver driver = (Driver)
+                Class.forName("org.apache.derby.jdbc.ClientDriver")
+                     .getDeclaredConstructor()
+                     .newInstance();
+
+            Properties properties = new Properties();
```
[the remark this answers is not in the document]
arjantijms: That's a pretty good question indeed...

OndroMih (approved):

appserver/admin/cli-optional/src/main/java/com/sun/enterprise/admin/cli/optional/DerbyExecuteSQL.java
arjantijms: The SQL ANT task (which the TCK often uses to init data) behaves the same way. I recently checked it out. It has an option to fail right away, and at the end it prints the number of failed / succeeded statements.

It has a ton of more options, but see: https://ant.apache.org/manual/Tasks/sql.html

arjantijms:
Still working on this, and will postpone to 7.0.19.

appserver/admin/cli-optional/src/main/java/com/sun/enterprise/admin/cli/optional/DerbyControl.java
```diff
@@ -55,38 +55,55 @@ public final class DerbyControl {
     public static void main(String[] args) {
         if (args.length < 3) {
             System.out.println("paramters not specified.");
-            System.out.println("DerbyControl <derby command> <derby host> <derby port> <derby home> <redirect output>");
+            System.out.println("DerbyControl <derby command> <derby host> <derby port> <redirect> [<derby home> | <derby user> <derby password>");
```
hs536, line 58: This message has an opening '[' but is missing a closing ']'.

appserver/admin/cli-optional/src/main/java/com/sun/enterprise/admin/cli/optional/DerbyExecuteSQL.java
dmatej: In this case is quite harmful to ignore the error instead of escalating it. Especially in TCK you simply don't want to search logs for a stacktrace when it should crash immediately. That is a good practice, if TCK doesn't follow that, we can fix it - I submitted fixes for JEE10 of this kind ;)
These hidden crashes took a lot of time.

arjantijms: In the case of the TCK it's not a real error. Derby should use

"create if not exists table ..." or "drop if exists table..."

But that syntax is not there.

So it must do:

"drop table"
"create table ..."

And then ignore the "drop table" if it fails.

PR #25171, opened 2024-10-07
Added parameter for "terse" to Asadmin used in tests

dmatej:
It follows usual scenarios:
- Usually we process the output by a software, most often we just need to know if it succeeded (which is signaled by the exit code 0), so there is no need to parse the Command succeeded message.
- But sometimes we do exactly opposite.
- Both ways usually differ by test class, sometimes by a test method.

The second change is about creating a backup of domain properties set by `asadmin set` and retrievable by `asadmin get`. (they can be manipulated also by other commands, like `create-jdbc-pool`, etc).

@OndroMih recently created this feature, this PR brings refactored and simplified version using the `terse` output and minimalistic api.


appserver/tests/application/src/test/java/org/glassfish/main/test/app/connpool/SQLTraceListenerTest.java
```diff
@@ -80,8 +80,8 @@ public static void deployAll() throws IOException {
         assertThat(result, asadminOK());
         // add-library shouldn't require a restart anymore
 
-        DOMAIN_SETTINGS.backupDerbyPoolSettings();
-        DOMAIN_SETTINGS.setDerbyPoolEmbededded();
+        DERBYPOOL_BACKUP.restore();
```
avpinchuk, line 83: This call seems redundant.

dmatej: Aha, you are right, and this was my mistake, mhm ...

avpinchuk (approved):

OndroMih (approved):

PR #25174, opened 2024-10-08
Reduced number of builds to keep from 10 to 2 as requested by Jenkins ops

dmatej:
- Currently Eclipse GlassFish projects uses 139 GB of CI disk space.
- There is a plan for migration to new infrastructure: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5072

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022, 2024 Contributors to the Eclipse Foundation
  * Copyright (c) 2018, 2021 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -161,7 +161,7 @@ spec:
   }
 
   options {
-    buildDiscarder(logRotator(numToKeepStr: '10'))
+    buildDiscarder(logRotator(numToKeepStr: '2'))
```
fredg02, line 164: You can keep 10 builds, but you should only keep the build artifacts of 2 builds:

```
buildDiscarder(logRotator(numToKeepStr: '10', artifactNumToKeepStr: '2'))
```

dmatej: Lets try to limit both, if we would have issues with that, we can always reconfigure it again. The thing is that logs (server and ant logs) are also artifacts here, and without them the build result is not much useful. However as all tests are usually reproducible locally, we don't need to keep so much data on Jenkins.

hs536 (approved):

OndroMih (approved):

Issue #24900, opened 2024-04-09
Document and refactor some connectors-runtime module code


escay:
## Environment Details
* GlassFish Version (and build number): 7.0.14
* JDK version: jdk-17.0.8.1_1-openjdk-adoptium
* OS: Windows 10
* Database: Postgresql

## Problem description

- Document (javadoc) some connection pool interfaces in connectors-runtime module to better understand issue #24805. With focus on the ResourceHandle fields to understand all fields:
   - com.sun.appserv.connectors.internal.api.ResourceHandle
   - com.sun.enterprise.resource.ResourceHandle
   - com.sun.enterprise.resource.ResourceState
- Refactor some code: remove unused methods, parameters and fields from interfaces.
- Write new unit test to test and understand `com.sun.enterprise.resource.pool.PoolManagerImpl` mainly regarding state isBusy and state enlisted behaviour of ResourceHandles.
 
Refactor items like:
```
ResourceHandle interface: remove destroyResource();
ConnectorConnectionPool: remove resourceSelectionStrategyClass
AssocWithThreadResourceHandle: remove ClientSecurityInfo from constructor
ResourceHandle: change resource and getResource() type from Object to ManagedConnection
 ## (risk if someone made their own implementation)
ResourceHandle: remove ClientSecurityInfo from constructor and field and getter/setter 
 ## (risk if someone made their own implementation)
ResourceHandle: remove subject field and getter/setter
ResourceHandle: remove partition field and getter/setter
ResourceHandle: remove setState setter, set state field immediately so it is never null
ResourceHandle: rename index to rwLockDataStructureResourceIndex
ResourceHandle: rename spec to resourceSpec
ResourceHandle: rename xaRes to xaResource
ResourceHandle: rename supportsLazyEnlistment_ to supportsLazyEnlistment
ResourceHandle: rename supportsLazyAssoc_ to supportsLazyAssociation

ResourceHandle / ResourceState: make ResourceState calls nullpointer safe.

ConnectionPool: remove resourceSelectionStrategyClass
ConnectionPool: rename validation to connectionValidationRequired
ConnectionPool: rename timer to resizerTaskTimer

DataStructureFactory: change parameters of class loading (risk if someone made their own implementation)

ListDataStructure: remove strategyClass from constructor
RWLockDataStructure: remove strategyClass from constructor	
```

escay:
Keep issue open for another while

PR #25176, opened 2024-10-08
Issue #24900 add a few unit tests for ManagedConnectionImpl and ConnectionHolder to understand associateConnection logic

escay:
I wrote a few unit tests to start to understand associateConnection logic which is used by ConnectorXAResource in connector-runtime. 
Added some javadoc to understand ManagedConnectionImpl parameters.
Added a listener null check and added unregister listener logic which seemed to be missing.
Removed a boolean parameter from setActive(true) which was always true.


dmatej:
Sorry, we had a broken metadata file in maven cache, that crashed your build, I have restarted the job now

appserver/jdbc/jdbc-ra/jdbc-core/src/main/java/com/sun/gjc/spi/JdbcObjectsFactory.java
```diff
@@ -52,13 +52,14 @@ public abstract class JdbcObjectsFactory implements Serializable {
...
+        String className = "com.sun.gjc.spi.jdbc40.Jdbc40ObjectsFactory";
         try {
             return(JdbcObjectsFactory)
-                Class.forName("com.sun.gjc.spi.jdbc40.Jdbc40ObjectsFactory")
+            Class.forName(className)
                      .getDeclaredConstructor()
                      .newInstance();
         } catch (Exception e) {
-            _logger.log(WARNING, "jdbc.jdbc_factory_class_load_exception", e);
+            _logger.log(WARNING, "jdbc.jdbc_factory_class_load_exception", className);
```
dmatej, line 62: As I am fighting with classloaders now, could you log the exception too? Sometimes it has an important cause.

escay: Yes no problem, I understand.
Is it ok to make the change and squash the commit?

dmatej (approved):

PR #25177, opened 2024-10-08
Reduced number of builds to keep from 10 to 2 as requested by Jenkins ops

dmatej:
- Currently Eclipse GlassFish projects uses 139 GB of CI disk space.


arjantijms (approved):

dmatej:
Seems the Jenkinsfile is bit obsoleted here, I will try to update it based on master ...

dmatej:
Seems there is some issue in admin tests, but I will merge it now.

PR #25180, opened 2024-10-09
CLI Environment - more useful log file

dmatej:
- load formatter just once
- more robust log file creation
- don't ignore exceptions, crash if logging is not possible. Basic requirement if used for some cmd line audit.
- Uses UTF-8, not default charset
- Uses ISO 8061 timestamp format - breaking change as original code used java.util.Date formatted as `MM/dd/yyyy HH:mm:ss` which is not a standard today. Despite the file is still trivial, it can be parsed automatically.

### DIFF
--- a/nucleus/admin/cli/src/main/java/com/sun/enterprise/admin/cli/Environment.java
+++ b/nucleus/admin/cli/src/main/java/com/sun/enterprise/admin/cli/Environment.java
@@ -109,23 +109,7 @@ public final class Environment {
             fname = System.getenv(logProp);
         }
         if (fname != null) {
-            File f = new File(fname);
-            try {
-                if (f.isFile() && f.canWrite()) {
-                    logfile = f;
-                } else {
-                    File dir = f.getParentFile();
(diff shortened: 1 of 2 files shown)


nucleus/admin/cli/src/main/java/com/sun/enterprise/admin/cli/Environment.java
```diff
@@ -109,13 +110,29 @@ public Environment(boolean ignoreEnvironment) {
...
             File f = new File(fname);
-
             try {
-                if ((f.exists() || f.createNewFile()) && f.isFile() && f.canWrite()) {
+                if (f.isFile() && f.canWrite()) {
+                    logfile = f;
+                } else {
+                    File dir = f.getParentFile();
+                    if (!dir.isDirectory() && !dir.mkdirs()) {
+                        throw new IllegalStateException("Could not create a log file: " + f);
```
avpinchuk, line 119: Maybe it's better adding a reason to the exception message? Something like "log directory not found"?

avpinchuk (approved):

PR #25182, opened 2024-10-11
Updated logging-annotation-processor with JPMS.

dmatej:
This PR will be followed by other that depends on this.


dmatej:
Maven Central is somehow faster than it was before few months, wow!
https://central.sonatype.com/artifact/org.glassfish.annotations/logging-annotation-processor/1.10/versions

pzygielo (approved):

avpinchuk (approved):

PR #25183, opened 2024-10-12
Redesign of GlassFish bootstrap

### LINKED ISSUE #25133: JDK loggers can initialize the JDK LogManager before GlassFishLogManager is set
[//]: # "If you are reporting a security issue or vulnerability, please contact the Eclipse Foundation Security Team instead of creating a GitHub issue."
[//]: # "  - Security policy: https://github.com/eclipse-ee4j/glassfish/security/policy"

## Environment Details
* GlassFish Version (and build number): 7.0.17
* JDK version: 21
* OS: Red Hat
----------

## Problem Description
After "start-domain" and executing 

`$ASADMIN multimode --file calypso.config7.txt` (attached)

Glassfish stops logging.

Notes:

- I've tried separated set-log-attributes with the same effect
- I've tried "only" set-log-attributes configuration with the same effect

[calypso.config7.txt](https://github.com/user-attachments/files/16907699/calypso.config7.txt)

## Steps to reproduce
Execute attached configuration

## Impact of Issue
Glassfish stops logging

dmatej:
## Main feature changes

* Fixes #25133 - we use the standard JVM option, however complete initialization is finished by HK2 service.
* ClassLoaders reflect boot phases
* asenv.conf respects existing environment
* Server's logging manager is set by JVM option
* Asadmin and nadmin and startserv scripts use JPMS
* appclient had to be refactored 

## Bootstrap libs
* glassfish.jar split to non-osgi and osgi part
* glassfish-jdk-extensions split from common-util and doesn't support osgi.
* simple-glassfish-api goes to command-line cp and doesn't support osgi.
* 4 libs support JPMS:  glassfish.jar, glassfish-jdk-extensions, glassfish-jul-extension, simple-glassfish-api
* these 4 libs are used as JPMS modules.
* `*.cli` jars were moved to install root as they are simply executable jars.

## Bootstrap design
* Basically still the same, but more comprehensible.
* Executable JAR (`admin-cli.jar`) started by `asadmin` script, or `appserver-cli.jar` started by `startserv` script
* That runs `GlassFishMainLauncher` which prepares command line to run the server, runs `GlassFishMain` and stops itself.
* `GlassFishMain` does some basic stuff and initializes OSGi and HK2 environment.
* Server started.

## Appclient changes
* JPMS broke many Ant tests, to fix them I had to reimplement also parts of appclient.
* Removed my 2 years old fix to run single class, because it fixed tests, but was not correct
* Created new classloaders as replacements of classpath in the appclient script
* Moved classes, created interfaces to disconnect `gf-client` and `gf-client-module`
   * Now they are not connected via classpath on command line and manifest any more
   * gf-client serves just to run `CliBootstrap` which generates command for the appclient script
   * The script then runs gf-client yet once with the Java agent to enable class instrumentation, to create the client container and set the classloader hierarchy.
   * `*Facade` classes then use the thread classloader to get the container and launch it.
* JWS did not and will not work for now
* There is few other ways to run the client, fortunately the test coverage is good here.

## Related changes 
* Were mostly pushed in own PRs this week, but some are here, some are waiting, but this should be a safe point.
* asenv.conf on linux respects external env variables. I don't know how to do that on Windows, so there it is as it was before.
* Added reproducer test for #25133 
* Classpath is not added to manifests. It caused issues with classloading, ie defaul-web.xml filter was ignored, required classes were still found by parent classloader. The manifest works just if files are when expected which is the case of the server, but then something else was not working - responsibility of classloader hierarchy. See #25179 
* Changes in appclient led to a discovery of several other changes - I will create PRs individually.

## Future PRs
* Next PR will disassemble SystemPropertyConstants and add enum making initialization of env and sys props easier.
* glassfish-jdk-extensions now contains just two classes, we will go through common-util and move here more things. Basically it should contain implementations which might be implemented in future JDK versions. It will never have dependencies on other GF libs.
* common-util will remain dependent on HK2, OSGI, whatever. It is used quite often, we should optimize it.
* we will get rid of felix-framework.jar because it overlaps with osgi-core.jar, or we have to repackage it. We have to try which solution is the best. JPMS hates it.
* Might be possible to start the logging immediately after start, I am not sure with that now, I need to prototype it first.

## Review

Probably is better to go through commits.
```
git remote add dmatej git@github.com:dmatej/glassfish.git
git checkout -b gjule-vs-gc dmatej/gjule-vs.gc
git diff -w -M git show -M -w dmatej/gjule-vs-gc^^^^^^^^^
```

## Notes

* Startup speed is nearly the same as with previous version - measured around some 50 times, this version has the best result, but we are talking about change in less then 10 ms of around 1500 - 3000 millis (the time including CLI).
* Breaking changes - breaking in meaning of jar file api compatibility, but even if someone would copy older domain.xml, it should work without issues (I did not try it). I don't expect anyone using GlassFish jar files directly except realms and login modules.


dmatej:
Wow, it took me a while to reproduce it - the reason is that JDK17 and JDK21 resolve Manifest classpath in different way! I have to find where and why it changed ...

dmatej:
Ok, another round, seems we don't have any mavenized tests for gfclient, tbd soon ...

dmatej:
TODO: Recommit and explain every step.

arjantijms (approved):

arjantijms:
> Will it be merged to 8.0?

Yes, it should be merged to 8.0 for sure. Hopefully we can to this for 8.0 M13, but certainly it's a blocking condition for 8.0 Final.

pzygielo:
I'm getting

```
$ java -jar glassfish7/glassfish/appserver-cli.jar
Error: Could not find or load main class org.glassfish.admin.cli.AsadminMain
Caused by: java.lang.NoClassDefFoundError: com/sun/enterprise/admin/cli/AdminMain
```

The manifest defines `Main-Class`, but no `Class-Path`.

pzygielo:
- In 0002084b13a: adding parameter to `KernelLoggerInfo.cantFindDerby` and not considering it in `CommonClassLoaderServiceImpl.findDerbyJars` caused #25650.

PR #25186, opened 2024-10-16
Add Microprofile Health

Thihup:
IMPORTANT: Depends on Java 17, should be merged to 7.1.0, not earlier.

This change supports the deployment of MicroProfile Health 4.0 compatible applications, and modifies GF full profile to pass the MP Health 4.0.1 TCK by implementing the specification from scratch.

The MicroProfile Health uses the MicroProfile Config to check for two properties, so instead of requiring an explicit dependency,it handles the absence of the MP Config by returning a default value.

Currently this implementation does not add any default procedure. 
Each application now contains four new endpoints: `/health`, `/health/live`, `/health/ready` and `/health/started`.

The TCK requires a single property (`mp.health.disable-default-procedures`) to be set (although currently unused by the implementation)


arjantijms:
Wow, super cool! Just curious, how many of the TCK tests are passing at the moment?

Maybe the manifest.mf file can be generated? It usually is, but I didn't check yet if you're doing anything special. Just a quick glance.

Thihup:
Tests passed/Failed/Skipped: | 27/1/0
The only test failing is the [testEmptyReadinessWithConfig](https://github.com/eclipse/microprofile-health/blob/72caa1983a77167264efa99b231c44531cbaf81c/tck/src/main/java/org/eclipse/microprofile/health/tck/ConfigTest.java#L55C17-L55C45).

The main issue related to this failure is use of the Microprofile Config. As I'm creating a new `MpConfigProviderResolver`, it is not reading the `microprofile-config.properties`.

As I understand, the correct use should be `ConfigProvider.getConfig()`, but using it fails with `No ConfigProviderResolver implementation found!`. I couldn't understand why it was happening as I required the implementation (Import-Package: io.helidon.config.mp;).

Related to the manifest, I had to change it to allow the use of a higher CDI version. I followed the pattern of the `microprofile-jwt-auth-api`.

Thihup:
Another issue: right now I'm registering the Servlet as the following `http://localhost:8080/my-app/health`, but the spec requires it to be `http://localhost:8080/my-app/health`.

OndroMih:
@Thihup, regarding "As I understand, the correct use should be ConfigProvider.getConfig(), but using it fails with No ConfigProviderResolver implementation found!", this is probably related to classloading or the order in which MP Health and Config are initialized. Config in GF is initialized in https://github.com/eclipse-ee4j/glassfish/blob/master/appserver/microprofile/config/src/main/java/org/glassfish/microprofile/config/ConfigDeployer.java. Maybe it happens after MP Health is initialized and thus Config is not yet available. If it's the case, we need to ensure that Health initializes after Config.

arjantijms:
> Related to the manifest, I had to change it to allow the use of a higher CDI version. I followed the pattern of the microprofile-jwt-auth-api.

You are absolutely right. I missed it being in a patches folder.

(maybe for the future we can add some extension somewhere in OSGi so that we can more directly specify we allow higher versions, without patching the entire manifest)

appserver/microprofile/health/src/main/java/org/glassfish/microprofile/health/HealthServlet.java
```diff
@@ -0,0 +1,135 @@
...
+    private static HealthCheckResponse buildHealthCheckResponse(String name, Exception e) {
+        return HealthCheckResponse.builder()
+                .down()
+                .name(name)
+                .withData("rootCause", e.getMessage())
+                .build();
+    }
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException {
```
OndroMih, line 64: Can we disable the health check endpoint with a system property? For example, return 404 status if disabled? 
It might be a security issue if the health endpoint is publicly available and wouldn't be possible to disable it.

Later we can add security to allow access to the endpoint only to authorized callers, but now it's enough if it's possible to disable the endpoint.

OndroMih (changes requested):

OndroMih (approved):
Looks awesome!

appserver/microprofile/health-glassfish/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+                            jakarta.enterprise.inject.spi;
+                            jakarta.json.bind;
+                            jakarta.servlet;
+                            jakarta.servlet.http;
+                            org.eclipse.microprofile.health;
+                            org.glassfish.internal.api;
+                            org.glassfish.microprofile.health;
+                            org.glassfish.hk2.api;
+                            org.glassfish.hk2.utilities;
+                        </Import-Package>
```
OndroMih, line 73: The `<Import-Package>` block can be removed. By default, all imported packages are added automatically: https://felix.apache.org/documentation/subprojects/apache-felix-maven-bundle-plugin-bnd.html#_import_package
```suggestion
```

appserver/microprofile/health-glassfish/src/main/java/org/glassfish/microprofile/impl/HealthService.java
```diff
@@ -0,0 +1,31 @@
...
+import org.glassfish.hk2.runlevel.RunLevel;
+import org.glassfish.internal.api.Globals;
+import org.glassfish.internal.data.ApplicationInfo;
+import org.glassfish.internal.deployment.Deployment;
+import org.glassfish.microprofile.health.HealthReporter;
+import org.jvnet.hk2.annotations.Service;
+
+@Service(name = "healthcheck-service")
+@RunLevel(StartupRunLevel.VAL)
+public class HealthService implements EventListener {
```
OndroMih, line 15: It's required to register the event listener, otherwise the event method won't be called:

```
import org.glassfish.api.event.Events;
import org.glassfish.hk2.api.PostConstruct;

public class HealthService implements PostConstruct, EventListener {

    @Inject
    Events events;

    @Override
    public void postConstruct() {
        events.register(this);
    }
```

appserver/microprofile/health-glassfish/src/main/java/org/glassfish/microprofile/impl/HealthService.java
```diff
@@ -0,0 +1,31 @@
...
+
+
+    @Override
+    public void event(Event<?> event) {
+
+        HealthReporter service = Globals.getDefaultHabitat().getService(HealthReporter.class);
+
+        if (event.is(Deployment.APPLICATION_UNLOADED) && event.hook() instanceof ApplicationInfo appInfo) {
+            service.removeAllHealthChecksFrom(appInfo.getName());
```
OndroMih, line 24: The healthchecks are registered with "" context root but here we want to remove them by `appInfo.getName()`, which is not "". As a result, the health checks are not removed on undeployment, remain leaking, and even remain in the output of the `/health` endpoint.

appserver/microprofile/health-glassfish/src/main/java/org/glassfish/microprofile/impl/CollectHealthChecksExtension.java
```diff
@@ -0,0 +1,75 @@
...
+            if (beans.getBean().getTypes().contains(HealthCheck.class)) {
+                healthChecks.add((Bean<HealthCheck>) beans.getBean());
+            }
+        }
+    }
+
+    public void afterDeploymentValidation(@Observes AfterDeploymentValidation adv, BeanManager beanManager) {
+        healthChecks.forEach(bean -> {
+            CreationalContext<HealthCheck> creationalContext = beanManager.createCreationalContext(bean);
+            service.addHealthCheck("", bean.create(creationalContext));
```
OndroMih, line 69: This should bind the health checks to application name, which has to be unique. When undeployed, `HealthService.event` should unbind health checks for the same application name.

The problem with the current approach is that when 2 applications are deployed, they deploy all health checks under the same context name. When one of the applications is undeployed, it would remove health checks also from other applications. We need to add and remove health checks for a single application under a unique context.

appserver/microprofile/health/src/main/java/org/glassfish/microprofile/health/HealthReporter.java
```diff
@@ -0,0 +1,148 @@
...
+        }
+        return new HealthReport(overallStatus, healthCheckResults);
+
+    }
+
+
+    public void addHealthCheck(String contextPath, HealthCheck healthCheck) {
+        applicationHealthChecks.computeIfAbsent(contextPath, k -> new CopyOnWriteArrayList<>())
+                .add(healthCheck);
+//        applicationHealthChecks.add(healthCheck);
```
OndroMih, line 126: remove comment

appserver/microprofile/health/src/main/java/org/glassfish/microprofile/health/HealthReporter.java
```diff
@@ -0,0 +1,148 @@
...
+                    .filter(HealthCheckResponse.Status.DOWN::equals)
+                    .findFirst()
+                    .orElse(HealthCheckResponse.Status.UP);
+        }
+        return new HealthReport(overallStatus, healthCheckResults);
+
+    }
+
+
+    public void addHealthCheck(String contextPath, HealthCheck healthCheck) {
```
OndroMih, line 123: I would rename `contextPath` to `contextName`. It has nothing to do with a "path", it's just a context to group several checks. In GlassFish, it would be application name.

OndroMih (changes requested):

appserver/microprofile/health-glassfish/src/main/java/org/glassfish/microprofile/impl/HealthService.java
```diff
@@ -0,0 +1,31 @@
...
+        }
+
+        if (event.is(Deployment.APPLICATION_STARTED) && event.hook() instanceof ApplicationInfo appInfo) {
```
OndroMih, line 27: Either remove this block, or use it to find out the application name to be used in the CDI extension for the application context name

appserver/microprofile/health-glassfish/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+        <artifactId>microprofile-parent</artifactId>
+        <version>7.0.21-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>health-glassfish</artifactId>
+    <packaging>glassfish-jar</packaging>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <maven.compiler.release>17</maven.compiler.release>
```
OndroMih, line 18: Remove `maven.compiler.release` setting, it's ignored.

appserver/microprofile/health-glassfish/src/main/java/org/glassfish/microprofile/impl/CollectHealthChecksExtension.java
OndroMih: From within an application. e.g. in the CDI extension, you can get the appname from InvocationManager.

There are multiple ways to retrieve it. The most straightforward is to get it from the global HK2 service locator via Globals.get(InvocationManager.class).

On InvocationManager, you get the app name:

```
InvocatoinManager im;
im.getCurrentInvocation().getAppName();
```

OndroMih (approved):

OndroMih:
I tested it, all looks good!

I raised one PR with a fix for GlassFish: https://github.com/eclipse-ee4j/glassfish/pull/25386. Without this, events are not properly triggered if app deployment fails, and then the healthchecks are not cleaned and leak until the app with the same name is deployed again and undeployed.

I raised a PR to this branch with an enhancement to add appname to the healthcheck report: https://github.com/Thihup/glassfish/pull/1

appserver/microprofile/health-glassfish/pom.xml
```diff
@@ -0,0 +1,74 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.glassfish.main.microprofile</groupId>
+        <artifactId>microprofile-parent</artifactId>
+        <version>7.0.23-SNAPSHOT</version>
```
OndroMih:
```suggestion
        <version>7.1.0-SNAPSHOT</version>
```

appserver/microprofile/health/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.glassfish.main.microprofile</groupId>
+        <artifactId>microprofile-parent</artifactId>
+        <version>7.0.23-SNAPSHOT</version>
```
OndroMih:
```suggestion
        <version>7.1.0-SNAPSHOT</version>
```

appserver/tests/tck/microprofile/health/pom.xml
```diff
@@ -0,0 +1,78 @@
...
+
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.glassfish.main.tests.tck</groupId>
+        <artifactId>glassfish-external-tck-microprofile</artifactId>
+        <version>7.0.23-SNAPSHOT</version>
```
OndroMih:
```suggestion
        <version>7.1.0-SNAPSHOT</version>
```

OndroMih (approved):

PR #25211, opened 2024-11-15
Merge from 8.0-JDK17

arjantijms:


Jenkinsfile
```diff
@@ -94,7 +94,7 @@ kind: Pod
 spec:
   containers:
   - name: maven
-    image: maven:3.9.6-eclipse-temurin-21
+    image: maven:3.9.6-eclipse-temurin-17
```
dmatej, line 97: This is wrong. 8 should be compiled by 21.

arjantijms: Wow, that's wrong indeed. Thanks for catching it!

pzygielo:
Removed from 8.0.0 milestone as 8.0 branch was forcibly pushed and c7c660e3d6 is no longer there.

PR #25212, opened 2024-11-16
Fix more merge mistakes

arjantijms:


appserver/tests/admin/tests/src/test/java/org/glassfish/main/admin/test/OSGiCommandsITest.java
```diff
@@ -56,13 +57,15 @@ public class OSGiCommandsITest {
     @BeforeAll
     public static void waitOsgiReady() throws Exception {
         long timeout = System.currentTimeMillis() + 10_000L;
-        while (System.currentTimeMillis() < timeout) {
-            AsadminResult result = ASADMIN.exec("osgi", "lb");
-            if (!result.isError()) {
-                return;
+        for (int i = 0; i < 50; i++) {
```
dmatej, line 60: The while loop should be enough.

dmatej (commented):
I am not sure if this is everything, I would expect we always do merges just from 8.0 to 8.0-JDK17 and never backwards. If there is some PR merged by mistake, I would do cherrypick and again merge in the expected direction.

The problem is that once you invert the direction, you take even changes you never wanted in 8.0.

OndroMih:
I agree with David, jdk17 branch should never be merged to the 8.0 branch.

I suggest that we revert https://github.com/eclipse-ee4j/glassfish/pull/25211, cherry pock relevant commits from the jdk17 branch to 8.0, and then merge 8.0 to jdk17.

arjantijms:
>  8.0 to 8.0-JDK17 and never backwards.

I know, but it happened now, so I'm trying to repair it.

appserver/tests/admin/tests/src/test/java/org/glassfish/main/admin/test/OSGiCommandsITest.java
arjantijms: >The while loop should be enough.

Good comment, but it's not a new thing. It was already there.

PR #25217, opened 2024-11-19
Fix #25216 Redirection does not work after session timeout in Admin Console

### LINKED ISSUE #25216: Redirection does not work after session timeout in Admin Console
## Environment Details

* GlassFish Version (and build number): GlassFish 7.0.19
* JDK version: OpenJDK 21.0.4
* OS: Red Hat Enterprise Linux 8.6
* Database: none

----------

## Problem Description

After session timeout in the Admin Console, the page only containing "class jakarta.faces.application.ViewExpiredException" is shown, and the web browser is not redirected to the home page or the login page.

gui-viewexpiredexception

This behavior is reproduced in GlassFish 7 or later.
It is expected that the web browser is redirected to the home page or the login page as in GlassFish 6 or earlier.

## Steps to reproduce

This is reproduced when a button triggering Ajax communication is clicked after session timeout.

1.  Run the following commands to set the admin session timeout to 1 minute.
    *  `asadmin set server.admin-service.das-config.admin-session-timeout-in-minutes=1`
    *  `asadmin stop-domain`
    *  `asadmin start-domain`
2.  Open the Admin Console in the web browser.
3.  Open the "Clusters" page and wait more than 1 minute.
4.  Click the "New" button.

## Impact of Issue

It is an incompatibility with GlassFish 6 or earlier.

tnagao7:
Fixes #25216 .

This fix adds an exception handler for the Admin Console to handle `jakarta.faces.application.ViewExpiredException`.
For Ajax requests, since forwarding/redirection methods of JSF do not work as expected, this fix uses redirection by `windows.location.href` as workaround.

### DIFF
--- a/appserver/admingui/war/src/main/java/org/glassfish/admingui/common/AdminGuiExceptionHandler.java
+++ b/appserver/admingui/war/src/main/java/org/glassfish/admingui/common/AdminGuiExceptionHandler.java
@@ -19,6 +19,7 @@ import jakarta.faces.FacesException;
 import jakarta.faces.application.ViewExpiredException;
 import jakarta.faces.context.ExceptionHandler;
 import jakarta.faces.context.ExceptionHandlerWrapper;
+import jakarta.faces.context.ExternalContext;
 import jakarta.faces.context.FacesContext;
 import jakarta.faces.event.ExceptionQueuedEvent;
 import jakarta.faces.event.ExceptionQueuedEventContext;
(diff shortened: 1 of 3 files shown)


appserver/admingui/war/src/main/java/org/glassfish/admingui/common/AdminGuiExceptionHandler.java
```diff
@@ -0,0 +1,64 @@
...
+            if (eventContext.getException() instanceof ViewExpiredException) {
+                try {
+                    FacesContext facesContext = FacesContext.getCurrentInstance();
+                    if (facesContext == null) {
+                        continue;
+                    }
+                    if (facesContext.getPartialViewContext().isAjaxRequest()) {
+                        facesContext.getExternalContext().getResponseOutputWriter().write(
+                                "<script type='text/javascript'>"
+                                        + "window.location.href = '" + facesContext.getExternalContext().getRequestContextPath() + "/';"
```
OndroMih, line 49: This can be stored in a variable and reused on line 53.

OndroMih (approved):

appserver/admingui/war/src/main/java/org/glassfish/admingui/common/AdminGuiExceptionHandler.java
tnagao7: I've corrected it.

dmatej (approved):

PR #25224, opened 2024-11-24
Integrate Jackson 2.18.1

arjantijms:
Chanes: https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.18.1

### DIFF
--- a/nucleus/parent/pom.xml
+++ b/nucleus/parent/pom.xml
@@ -144,7 +144,7 @@
         <antlr.version>2.7.8</antlr.version>
         <ant.version>1.10.12</ant.version>
         <ant.external.version>1.10.2</ant.external.version>
-        <jackson.version>2.17.1</jackson.version>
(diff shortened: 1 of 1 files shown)


OndroMih:
By the way, is Jackson needed in GlassFish? It’s bundled and available for apps but I only found one module in GF that depends on it - xmlbind-annotations: https://github.com/eclipse-ee4j/glassfish/blob/1a19eb80f196f13b5ffaf585a0920c241b50ab99/nucleus/xmlbind-annotations/pom.xml#L28

Could we remove Jackson from GF 7.1 or GF 8?

OndroMih (approved):

arjantijms:
> Could we remove Jackson from GF 7.1 or GF 8?

Good question, let's see where xmlbind-annotations uses it for, AND whether there's not some other library that transitively depends on it.

dmatej (approved):

PR #25226, opened 2024-11-24
Fixed RarHandler - non-absolute URL to File conversion

dmatej:
- As the Archive#getURI says, the method can return nearly anything and there no guarantee that the URI can be converted to a File.
- Sometimes it is an absolute file, sometimes relative path, sometimes URI pointing to a file inside jar inside ear.
- This minimal fix avoids an exception, and it is similar to conversion used on other places.

### DIFF
--- a/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/connectors/module/RarHandler.java
+++ b/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/connectors/module/RarHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022, 2024 Contributors to the Eclipse Foundation
  * Copyright (c) 1997, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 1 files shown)


appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/connectors/module/RarHandler.java
```diff
@@ -160,4 +160,13 @@ public List<URI> getClassPathURIs(ReadableArchive archive) {
         }
         return uris;
     }
+
+
+    private static File toFile(URI uri) {
+        try {
+            return new File(uri.getSchemeSpecificPart());
+        } catch (IllegalArgumentException e) {
```
pzygielo, line 168: Previously used [File(URI)](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/io/File.html#%3Cinit%3E(java.net.URI)) declares `IAEx`, but [File(String)](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/io/File.html#%3Cinit%3E(java.lang.String)) does not.
Can you show, the best with test(s) for this `toFile`, case with `IAEx` being thrown here, please?

dmatej: Good catch, I have seen the exception, then did some incorrect fix, then this, but I forgot to remove it.

arjantijms (approved):

PR #25228, opened 2024-11-24
Fixed processing of usesCallerIdentity and runAs

dmatej:
- XML can contain null/false/true, null is not equal to true, but true is used as a default value after processing.
- setElementValue complained about missing descriptor (exception)


arjantijms (approved):

arjantijms:
> Boolean.FALSE.equals

As a small stylistic issue, I think qualifying `FALSE` with `Boolean` doesn't add much to readability or understanding of the code.

PR #25234, opened 2024-11-26
Integrate Felix Webconsole 5.0.8

arjantijms:
Changes: https://github.com/apache/felix-dev/compare/org.apache.felix.webconsole-4.8.12...org.apache.felix.webconsole-5.0.8


OndroMih:
It seems like this is not used in GlassFish at all. I only see it used in felix-webconsole-extension, which is not used anywhere.

OndroMih (approved):

arjantijms:
Another one we may investigate somewhat closer indeed

PR #25246, opened 2024-11-27
Add some more exclusions for EE 11 dependencies

arjantijms:


pzygielo:
Did you mean EE 10?

pzygielo:
> Also exclude MQ 6.5, its using JDK 21

But it's built for JDK 11 I believe.

arjantijms:
> > Also exclude MQ 6.5, its using JDK 21
> 
> But it's built for JDK 11 I believe.

Maybe that's for 6.6 then? I quickly looked at master and saw:

https://github.com/eclipse-ee4j/openmq/blob/master/mq/main/pom.xml#L64

pzygielo (approved):

Issue #25169, opened 2024-10-03
Asadmin should use STDOUT for communication with the user, not STDERR and the logging system


dmatej:
I have to analyze how many changes this means, but basically when somebody changes the formatter or level, user cannot use commands in script which parse the output.

Expectations:
* --terse should produce minimalized parseable output, ie. `osgi --session new` should produce just the session id, no "command successful", no timestamp, no log level, just one new line after the id, done, regardless log format.
* Without --terse, the command can also produce some logs using defined level and formatter. By default, should print INFO+ level messages to STDOUT, but without timestamps, levels, etc., like now; Btw, now prints FINE too, that sometimes creates a mess.
* When using `--terse` and configured logfile, it should use that for all asadmin commands. The current state causes #24065 .
* When instance launcher fails, user should see somewhere why that happened.

Consider also some flexibility:
* Using logging.properties (higher priority, should be possible to combine, partially), might be specified in asenv.conf as ASADMIN_LOGGING_CFG_FILE
* Using just env variables and no logging.properties: 
    * `ASADMIN_LOG_FILE=${HOME}/asadmin.log`, 
    * `ASADMIN_LOG_FILE_FORMATTER=org.glassfish.main.jul.formatter.ODLLogFormatter` , 
    * `ASADMIN_LOG_FILE_FORMAT=`, 
    * `ASADMIN_LOG_FILE_ROLL_POLICY=hour/day/week/month/year/never`, 
    * `ASADMIN_LOG_FILE_MAX_FILES=10`, 
    * `ASADMIN_STDERR_FORMATTER=%m`, 
    * `ASADMIN_STDERR_FORMAT=%m`, 
    * `ASADMIN_STDERR_ENABLED` - last two could be overridden by using `--terse` explicitly

Notes:
* linux can separate STDOUT (for user) and STDERR (for developer), but most people use merged output. That is why we have `--terse`.
* asadmin can work with multiple domains, instances, so the logging directory should not be the domain log dir (also because it can be deleted with the domain). We can use the home dir as a default, because it is on every operating system, however user should specify some well chosen directory.
* About the breaking change label - it can affect some scripts, however probably just to simplify them, mostly they will work still the same way. We have to check the CI, TCK, and all we have if we can find some issues with these changes.
[...]

PR #25262, opened 2024-12-03
Suppress internal processing output on executing asadmin create-local-instance

dmiya3:
Internal processing message "Extracted transferred entry ..." [*1] is output on executing asadmin create-local-instance.

```sh
# /opt/glassfish7-org/glassfish/bin/asadmin create-local-instance --cluster cluster1 cluster1-1
Rendezvoused with DAS on localhost:4848.
Port Assignments for server instance cluster1-1: 
OSGI_SHELL_TELNET_PORT=26666
JAVA_DEBUGGER_PORT=29009
JMS_PROVIDER_PORT=27676
HTTP_LISTENER_PORT=28080
IIOP_SSL_LISTENER_PORT=23820
ASADMIN_LISTENER_PORT=24848
IIOP_SSL_MUTUALAUTH_PORT=23920
JMX_SYSTEM_CONNECTOR_PORT=28686
HTTP_SSL_LISTENER_PORT=28181
IIOP_LISTENER_PORT=23700
Extracted transferred entry config/keystore.jks of size 6,221 B to /opt/glassfish7-org/glassfish/nodes/localhost-domain1/cluster1-1/config/keystore.jks [*1]
Extracted transferred entry config/cacerts.jks of size 2,495 B to /opt/glassfish7-org/glassfish/nodes/localhost-domain1/cluster1-1/config/cacerts.jks [*1]
Command create-local-instance executed successfully.
```

FYI, we can still view the internal processing with environment variables (`AS_TRACE` / `AS_DEBUG`).

```sh
# export AS_TRACE=true
# /opt/glassfish7-org/glassfish/bin/asadmin create-local-instance --cluster cluster1 cluster1-1
...
---- END PAYLOAD ----
Copyied 6,221 bytes to /opt/glassfish7-org/glassfish/nodes/localhost-domain1/cluster1-1/config/keystore.jks
Extracted transferred entry config/keystore.jks of size 6,221 B to /opt/glassfish7-org/glassfish/nodes/localhost-domain1/cluster1-1/config/keystore.jks [*1]
Copyied 2,495 bytes to /opt/glassfish7-org/glassfish/nodes/localhost-domain1/cluster1-1/config/cacerts.jks
Extracted transferred entry config/cacerts.jks of size 2,495 B to /opt/glassfish7-org/glassfish/nodes/localhost-domain1/cluster1-1/config/cacerts.jks [*1]
doHttpCommand succeeds
loadClass(name=com.sun.enterprise.admin.util.CommandModelData$ParamData, resolve=false)
Command create-local-instance executed successfully.
```

### DIFF
--- a/nucleus/common/common-util/src/main/java/org/glassfish/admin/payload/PayloadFilesManager.java
+++ b/nucleus/common/common-util/src/main/java/org/glassfish/admin/payload/PayloadFilesManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022, 2024 Contributors to the Eclipse Foundation
  * Copyright (c) 1997, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -457,7 +457,7 @@ public abstract class PayloadFilesManager {
                 dirTimestamps.put(extractedFile, lastModified);
             }
             postExtract(extractedFile);
-            logger.log(Level.CONFIG, "Extracted transferred entry {0} of size {1} B to {2}",
+            logger.log(Level.FINER, "Extracted transferred entry {0} of size {1} B to {2}",
                 new Object[] {part.getName(), extractedFile.length(), extractedFile});
             reportExtractionSuccess();
             return extractedFile;


dmatej:
Honestly I was missing these logs when we had issues on production and sometimes we were not sure if something changed or not.

What I do believe is more important - **please tell me if this is the real problem for you: asadmin spams user output which is parsed by some script**  - because the asadmin uses STDOUT both for logging and for user communication. I already created an issue for that: https://github.com/eclipse-ee4j/glassfish/issues/25169

The GFLauncherLogger class uses STDOUT instead of STDERR, but even worse, it mixes user communication and logs so we cannot simply switch that as it would create other issues.

For now I will approve this (so it would stop breaking scripts), after #25169 these logs should be config/info level, but in STDERR instead of STDOUT.

dmatej (approved):

PR #25267, opened 2024-12-05
Support injecting entity manager factory, and some utils from it

arjantijms:


pzygielo:
- seems related: #25377

PR #25270, opened 2024-12-05
Improve the connection pool lookup messages

dmiya3:
I would like to change the log level of the following message from WARNING to INFO because this warning is printed in the server log even though the lazy-loading works as expected.

```
[2024-12-05T12:26:39.599471+09:00] [GF 7.0.20] [WARNING] [] [jakarta.enterprise.resource.resourceadapter.com.sun.enterprise.connectors.service] [tid: _ThreadID=104 _ThreadName=pool-18-thread-1] [levelValue: 900] [[
  Probably the pool org.glassfish.resourcebase.resources.api.PoolInfo@fdae7f48[jndiName=MyPool1, applicationName=null, moduleName=null] is not yet initialized (lazy-loading), trying to check ...]]
```

I would also like to make the messages a little better when the lookup ultimately fails, so it's easier for those unfamiliar with the internal work (two lookups, accounting for lazy-loading) to understand.

### DIFF
--- a/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/connectors/service/ConnectorResourceAdminServiceImpl.java
+++ b/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/connectors/service/ConnectorResourceAdminServiceImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022, 2024 Contributors to the Eclipse Foundation
  * Copyright (c) 1997, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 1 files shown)


appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/connectors/service/ConnectorResourceAdminServiceImpl.java
```diff
@@ -66,15 +66,15 @@ public void createConnectorResource(ResourceInfo resourceInfo, PoolInfo poolInfo
...
             ccp = (ConnectorConnectionPool) namingService.lookup(poolInfo, jndiNameForPool);
         } catch (NamingException ne) {
-            _logger.log(Level.WARNING,
+            _logger.log(Level.INFO,
                 "Probably the pool {0} is not yet initialized (lazy-loading), trying to check ...", poolInfo);
             try {
                 checkAndLoadPool(poolInfo);
                 ccp = (ConnectorConnectionPool) namingService.lookup(poolInfo, jndiNameForPool);
             } catch (NamingException e) {
-                _logger.log(Level.SEVERE, "Second lookup failed for {0}: {1}", new Object[] {poolInfo, e});
```
dmatej, line 75: Just a question - isn't the error logged twice? I am usually deleting these logs like this if the exception is thrown immediately after it. 
The only exception is when the thrown exception is somehow processed and then we would not have the stacktrace in logs (also this log doesn't produce stacktrace, just `poolInfo.toString` and `e.toString`) or if it is logged on lower level.

dmiya3: As you said, the error logs are duplicated because the exception message is also printed in:  

https://github.com/eclipse-ee4j/glassfish/blob/7bfd4527f92e1f76115af814c630da880a6c7bc0/nucleus/resources/src/main/java/org/glassfish/resourcebase/resources/listener/ResourceManager.java#L430

I've corrected it.

dmatej (approved):

PR #25271, opened 2024-12-06
Fix CLI031 warning message on executing asadmin ping-connection-pool command

takahashi024:
* Same as #23643

  
The `asadmin ping-connection-pool` command outputs  CLI031 warning message even when executed without the target option, as shown below:

```
> asadmin ping-connection-pool DerbyPool
CLI031: Warning: Option "target" is obsolete and will be ignored.
CLI031: Warning: Option "target" is obsolete and will be ignored.
Command ping-connection-pool executed successfully. 
```


dmatej:
Hmm, seems I caused that here: https://github.com/eclipse-ee4j/glassfish/commit/1c635426081cab3094536f37c40b9e5c749eacbb
Unfortunately it was not visible that it will complain. Thanks!

dmatej (approved):

PR #25283, opened 2024-12-13
Fix versions of X-Powered-By field for empty "product.name"

dmiya3:
* Follow-up: https://github.com/eclipse-ee4j/glassfish/pull/24322

X-Powered-By response field looks broken as follows by setting "product.name" system property to the empty string [*1]:

```sh
# curl -i http://localhost:8080/
HTTP/1.1 200 OK
X-Powered-By: Servlet/6.0 JSP/3.1( Java/Eclipse Adoptium/17)
```

Expected behavior:

```sh
# curl -i http://localhost:8080/
HTTP/1.1 200 OK
X-Powered-By: Servlet/6.0 JSP/3.1(Eclipse GlassFish 7.0.21 Java/Eclipse Adoptium/17)
```

[*1] This use case is described below:

https://github.com/eclipse-ee4j/glassfish/blob/71df878e14aaca628a8c97ef9bff1894751de26a/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/services/impl/GlassfishNetworkListener.java#L357-L371

### DIFF
--- a/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/services/impl/GlassfishNetworkListener.java
+++ b/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/services/impl/GlassfishNetworkListener.java
@@ -373,10 +373,9 @@ public class GlassfishNetworkListener extends GenericGrizzlyListener {
 
             serverVersion = serverInfo == null ? Version.getProductId() : serverInfo;
 
-            if (isXPoweredByEnabled) {
-                String serverVersionForXPoweredBy = (serverInfo == null || serverInfo.isEmpty()) ? Version.getProductId() : serverInfo;
+            if (isXPoweredByEnabled && (serverVersion != null && !serverVersion.isEmpty())) {
                 xPoweredBy = "Servlet/6.0 JSP/3.1"
-                        + "(" + serverVersionForXPoweredBy
+                        + "(" + serverVersion
                         + " Java/"
                         + System.getProperty("java.vm.vendor") + "/"
                         + System.getProperty("java.specification.version") + ")";
--- a/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/services/impl/GlassfishNetworkListener.java
+++ b/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/services/impl/GlassfishNetworkListener.java
@@ -371,9 +371,10 @@ public class GlassfishNetworkListener extends GenericGrizzlyListener {
             */
             String serverInfo = System.getProperty("product.name");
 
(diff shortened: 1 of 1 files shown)


dmatej (changes requested):

OndroMih (changes requested):
I would argue that the expected behavior you described is not really expected. On the contrary. If users set the value of `product.name` to an empty string, they don't want to disclose information about the server. So it shouldn't even appear in the X-Powered-By header.

I would even go that far to disable the X-Powered-By header if `product.name` is set but it's empty. Like this:

```
if (isXPoweredByEnabled || (serverInfo != null && serverInfo.isEmpty()) {
```

The same thing can be achieved by setting the x-powered-by property to false, but this would allow to disable both the `Server` and `X-Powered-By` HTTP headers with a single setting.

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/services/impl/GlassfishNetworkListener.java
```diff
@@ -374,8 +374,9 @@ public GlassfishHttpCodecFilter(
             serverVersion = serverInfo == null ? Version.getProductId() : serverInfo;
 
             if (isXPoweredByEnabled) {
```
OndroMih, line 376: Why not just disable the whole `X-Powered-By` header if `product.name` is empty? I'd expect this is what users would expect if they set it to an empty value.

```suggestion
            if (isXPoweredByEnabled || (serverInfo != null && serverInfo.isEmpty()) {
```

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/services/impl/GlassfishNetworkListener.java
```diff
@@ -374,8 +374,9 @@ public GlassfishHttpCodecFilter(
...
+                String serverVersionForXPoweredBy = (serverInfo == null || serverInfo.isEmpty()) ? Version.getProductId() : serverInfo;
```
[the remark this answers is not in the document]
dmiya3: Thank you for your advice.  
I've refectored it.

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/services/impl/GlassfishNetworkListener.java
dmiya3: Thank you for your suggestion.  
I described the past GlassFish behavior as expected one, but your idea would be better.  
I've corrected it.

OndroMih (approved):

dmatej (approved):

PR #25287, opened 2024-12-15
Fix embedded faces init

arjantijms:
Faces applications fail to init when using embedded, since the service locator for setting the invocation does not match the one inside the servlet context for Faces apps to use, but only when using embedded.

A fix for now is to set the different service locator only for system apps (it is an app scoped locator, only really necessary for the admin ui).

### DIFF
--- a/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebModule.java
+++ b/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebModule.java
@@ -1180,18 +1180,12 @@ public class WebModule extends PwcWebModule implements Context {
         return isStandalone;
     }
 
-    /**
-     * @return the isSystemApplication
-     */
     public boolean isSystemApplication() {
         return isSystemApplication;
     }
 
-    /**
-     * @param isSystemApplication the isSystemApplication to set
-     */
-    public void setSystemApplication(boolean isSystemApp) {
-        this.isSystemApplication = isSystemApp;
+    public void setSystemApplication(boolean isSystemApplication) {
+        this.isSystemApplication = isSystemApplication;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebModuleListener.java
```diff
@@ -189,12 +189,17 @@ private void configureJsp(WebModule webModule) {
...
         ServiceLocator defaultServices = webContainer.getServerContext().getDefaultServices();
-        final String servicesName = webModule.getComponentId();
-        ServiceLocator webAppServices = ServiceLocatorFactory.getInstance().create(servicesName, defaultServices);
-        initializeServicesFromClassLoader(webAppServices, Thread.currentThread().getContextClassLoader());
 
-        // set services for jsf injection
-        servletContext.setAttribute(Constants.HABITAT_ATTRIBUTE, webAppServices);
+        if (webModule.isSystemApplication()) {
+            final String servicesName = webModule.getComponentId();
+            ServiceLocator webAppServices = ServiceLocatorFactory.getInstance().create(servicesName, defaultServices);
```
OndroMih, line 195: Can tou move creation of webAppServices to a separate method for clarity?

OndroMih: I mean, refactor the code, so that the code to create webAppServices is extracted into a new method.

appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebModule.java
```diff
@@ -1178,6 +1180,20 @@ protected boolean isStandaloneModule() {
...
 
+    /**
+     * @return the isSystemApplication
+     */
+    public boolean isSystemApplication() {
+        return isSystemApplication;
+    }
+
+    /**
+     * @param isSystemApplication the isSystemApplication to set
```
pzygielo, line 1191: 1. no such parameter
2. I'd drop both (here and in getter) javadocs. Noone complained about missing one for `getBean`/`setBean` for example.

arjantijms: You are right, those docs don't make sense. My IDE was still configured for a project where useless docs are the norm ;)

OndroMih (approved):

PR #25289, opened 2024-12-17
Fix #25288 asadmin delete-log-levels command fails if the --target option is specified

### LINKED ISSUE #25288: asadmin delete-log-levels command fails if the --target option is specified
## Environment Details
* GlassFish Version (and build number): Eclipse GlassFish 7.0.20
* JDK version: OpenJDK 21.0.3
* OS: Red Hat Enterprise Linux 8.6
* Database: none

----------

## Problem Description

If the `--target` option is specified, the `asadmin delete-log-levels` command fails to delete log levels of the specified target.

```
$ asadmin list-log-levels cluster1 | grep my.logger
 my.logger       <FINE>
    
$ asadmin delete-log-levels --target cluster1 my.logger
remote failure: An error occurred during replication
Failure: Command delete-log-levels failed on server instance instance1: remote failure: Could not delete logger levels for cluster1.
Command delete-log-levels failed.
```

## Steps to reproduce

Execute the following command lines:

```
asadmin create-cluster cluster1
asadmin create-local-instance --cluster cluster1 instance1
asadmin set-log-levels --target cluster1 my.logger=FINE
asadmin start-cluster cluster1
asadmin delete-log-levels --target cluster1 my.logger
```

## Impact of Issue
Cannot delete log levels of the specified target.

tnagao7:
Fixes #25288.

Fixes the `LoggingConfigImpl#deleteLoggingProperties` method to handle the `target` argument.

### DIFF
--- a/nucleus/common/common-util/src/main/java/com/sun/common/util/logging/LoggingConfigImpl.java
+++ b/nucleus/common/common-util/src/main/java/com/sun/common/util/logging/LoggingConfigImpl.java
@@ -232,7 +232,7 @@ public class LoggingConfigImpl implements LoggingConfig, PostConstruct {
             setWebLoggers(propertyValue);
         }
 
-        closePropFile("");
+        closePropFile(null);
         return property;
     }
 
@@ -298,7 +298,7 @@ public class LoggingConfigImpl implements LoggingConfig, PostConstruct {
             m.put(key, property);
 
         }
-        closePropFile("");
+        closePropFile(null);
         return m;
     }
 
@@ -385,7 +385,7 @@ public class LoggingConfigImpl implements LoggingConfig, PostConstruct {
         LOG.log(Level.INFO, "deleteLoggingProperties(properties={0})", properties);
         loadLoggingProperties();
         remove(properties);
-        rewritePropertiesFileAndNotifyMonitoring("");
+        rewritePropertiesFileAndNotifyMonitoring(null);
         return setMissingDefaults(getLoggingProperties());
     }
 
--- a/nucleus/common/common-util/src/main/java/com/sun/common/util/logging/LoggingConfigImpl.java
+++ b/nucleus/common/common-util/src/main/java/com/sun/common/util/logging/LoggingConfigImpl.java
@@ -407,13 +407,13 @@ public class LoggingConfigImpl implements LoggingConfig, PostConstruct {
 
     private void rewritePropertiesFileAndNotifyMonitoring(final String targetConfigName) throws IOException {
         LOG.finest("rewritePropertiesFileAndNotifyMonitoring");
-        File file = null;
+        final File file;
         if (targetConfigName == null || targetConfigName.isEmpty()) {
             file = getLoggingPropertiesFile();
         } else {
             file = getLoggingPropertiesFile(targetConfigName);
         }
-        File parentFile = file.getParentFile();
+        final File parentFile = file.getParentFile();
         if (!parentFile.exists() && !parentFile.mkdirs()) {
             throw new IOException(
                 "Directory '" + parentFile + "' does not exist, cannot create logging.properties file!");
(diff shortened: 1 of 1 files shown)


nucleus/common/common-util/src/main/java/com/sun/common/util/logging/LoggingConfigImpl.java
```diff
@@ -385,7 +385,7 @@ public synchronized Map<String, String> deleteLoggingProperties(final Collection
         LOG.log(Level.INFO, "deleteLoggingProperties(properties={0})", properties);
         loadLoggingProperties();
         remove(properties);
-        rewritePropertiesFileAndNotifyMonitoring();
+        rewritePropertiesFileAndNotifyMonitoring("");
```
dmatej, line 388: `null` would be better

nucleus/common/common-util/src/main/java/com/sun/common/util/logging/LoggingConfigImpl.java
```diff
@@ -394,20 +394,25 @@ public synchronized Map<String, String> deleteLoggingProperties(final Collection
...
     private void remove(final Collection<String> keys) {
         final Consumer<String> toRealKeyAndDelete = k -> props.remove(LoggingXMLNames.xmltoPropsMap.getOrDefault(k, k));
         keys.forEach(toRealKeyAndDelete);
     }
 
-    private void rewritePropertiesFileAndNotifyMonitoring() throws IOException {
+    private void rewritePropertiesFileAndNotifyMonitoring(final String targetConfigName) throws IOException {
         LOG.finest("rewritePropertiesFileAndNotifyMonitoring");
-        File file = getLoggingPropertiesFile();
+        File file = null;
```
dmatej, line 410: You can make it final and remove the redundant null assignment as you always set it in `if` or in `else` block.

tnagao7: I have modified the code.

nucleus/common/common-util/src/main/java/com/sun/common/util/logging/LoggingConfigImpl.java
tnagao7: Thank you for your suggestion. 
But existing code uses  `""` instead of `null` as below, and using `""` and `null` mixedly would not be desirable...

https://github.com/eclipse-ee4j/glassfish/blob/858cfe603f66c1e2c5b4d9cbf10baa9da01ac3e7/nucleus/common/common-util/src/main/java/com/sun/common/util/logging/LoggingConfigImpl.java#L235

https://github.com/eclipse-ee4j/glassfish/blob/858cfe603f66c1e2c5b4d9cbf10baa9da01ac3e7/nucleus/common/common-util/src/main/java/com/sun/common/util/logging/LoggingConfigImpl.java#L301

If we want to refactor this code, method overloading (i.e. defining `rewritePropertiesFileAndNotifyMonitoring()` and `closePropFile()` without arguments) would be better rather than using `""` and `null`.
Since doing such refactoring in the current pull request obscures the essense of the bug fix, I think we can do it in another pull request if needed.

dmatej: I would just replace all those empty strings with `null` (and in the future this class should never get an empty string target, it should be resolved when parsing the user input). Adding overloaded methods doesn't bring a value in private methods. Refactoring is ok if you put it to separate commit from the fix.

Additional PR means more work to each of us and waiting on CI ;-)

dmatej (approved):

nucleus/common/common-util/src/main/java/com/sun/common/util/logging/LoggingConfigImpl.java
tnagao7: Okay, I've modified the code to use `null` instead of `""`.

dmatej (approved):

Issue #25206, opened 2024-11-06
Jakarta Security: SecurityContext roles intermittently lost on page reload or navigation


riyadomf:
## Environment Details
* GlassFish Version (and build number): Eclipse GlassFish 7.0.19 and 7.0.15 platform (tested on both)
* JDK version: 17
* Jakarta EE version: Jakarta EE 10
* OS: Ubuntu 22.04 and Ubuntu 24.04
* Database: PostgreSQL

----------

## Problem Description
When reloading or navigating between pages in a `Jakarta EE` application running on `GlassFish`, `SecurityContext` roles unexpectedly disappear, resulting in inconsistent access control behavior and unexpected `403` Forbidden errors. The user principal (`securityContext.getCallerPrincipal()`) is still present, but roles are sometimes unavailable, causing access restrictions even for authenticated users with valid roles.

## Steps to reproduce
1. Authenticate a user with specific roles.
2. Navigate between pages or refresh the page.
3. Observe that suddenly `securityContext.getCallerPrincipal()` remains non-null, but role-checking methods like `securityContext.isCallerInRole("roleName")` return false for roles that should be active. Similarly `securityContext.getAllDeclaredRoles()` returns only `**` role, while all other roles temporarily disappear. However, The roles reappear and function correctly upon the next reload.

## Impact of Issue
This issue disrupts role-based access control in applications, causing intermittent authorization failures during navigation or page refreshes. As a result, users may lose access to certain features or pages even though they are authenticated, leading to inconsistent user experiences and potential security concerns. This behavior is particularly problematic for applications relying on stable, role-based permissions across sessions.

Type: Bug
Component: Security / Authentication
Priority: High

OndroMih:
Hi @riyadomf, do you have a reproducer application for this? Can you, please, share it with us or prepare a simple application on which you can reproduce the issue? It would greatly help us investigate the issue and if it's caused by a bug, fix it.

OndroMih:
Hi @riyadomf, would you be able to create a simple reproducer that can be easily used to reproduce the issue?

hberton:
> ## Environment Details
> * GlassFish Version (and build number): Eclipse GlassFish 7.0.19 and 7.0.15 platform (tested on both)
> * JDK version: 17
> * Jakarta EE version: Jakarta EE 10
> * OS: Ubuntu 22.04 and Ubuntu 24.04
> * Database: PostgreSQL
> 
> ## Problem Description
[...]

PR #25290, opened 2024-12-17
Fix random 403 responses

### LINKED ISSUE #25141: Sometimes 403 Forbidden server response when websocket message call ajax update with identity store
## Environment Details
* Eclipse GlassFish 7.0.18-SNAPSHOT (commit: b4bd2e2bc141c6744575fe90b4cac561e7368b36)
* JDK version: 21.0.3
* OS: Windows 10

----------

## Problem Description
Sometimes server return 403 forbidden

I add RealmAdapter log:

Normal trace:
  RealmAdapter.findPrincipalWrapper
 principal:jakarta.security.enterprise.CallerPrincipal@31614fef
 sessionPrincipal:jakarta.security.enterprise.CallerPrincipal@31614fef|#]
  RealmAdapter.findPrincipalWrapper
 principal:jakarta.security.enterprise.CallerPrincipal@41e82e26
 sessionPrincipal:jakarta.security.enterprise.CallerPrincipal@41e82e26|#]
  sessionBean.getText(): TEST1|#]
  sessionBean.getText(): TEST2|#]

When 403 forbidden:
  RealmAdapter.findPrincipalWrapper
 principal:jakarta.security.enterprise.CallerPrincipal@41e82e26
 sessionPrincipal:jakarta.security.enterprise.CallerPrincipal@31614fef|#]
  RealmAdapter.findPrincipalWrapper
 principal:jakarta.security.enterprise.CallerPrincipal@31614fef
 sessionPrincipal:jakarta.security.enterprise.CallerPrincipal@31614fef|#]
  RealmAdapter.validate sessionSubject == null|#]
  sessionBean.getText(): TEST1|#]

Server return 403 for TEST2,
In fonction RealmAdapter.findPrincipalWrapper caller is TEST2 and sessionPrincipal is TEST1

## Steps to reproduce

[TestSession.zip](https://github.com/user-attachments/files/16970482/TestSession.zip)

Compile TestSession project
Deploy TestSession

Open chrome explorer and developper tools, call http://127.0.0.1:8080/test/
login with TEST1 / TEST1

Open edge explorer and developper tools, call http://127.0.0.1:8080/test/
login with TEST2 / TEST2

Sometimes server return 403 forbidden

## Impact of Issue
Client not refresh when server return 403

403-forbidden

OndroMih:
Fixes #25141 and https://github.com/eclipse-ee4j/glassfish/issues/25206.

The securityContext is stateful, cannot be shared among requests. [RealmAdapter](https://github.com/eclipse-ee4j/glassfish/blob/e20777f9194eb85d0ad39a35938f229f1ca5b794/appserver/security/webintegration/src/main/java/com/sun/web/security/RealmAdapter.java#L447) stores the principal from request to it. We need to create  a copy of the default context for each request instead of  using the shared default context.

This is a quick fix but I can imagine a better and nicer fix. For example, keep returning `defaultSecurityContext` and instead let `RealmAdapter` to create a new security context if a user is signed in, and then store the principal to it. This would keep the same `defaultSecurityContext` for anonymous user. 

Another option is to save the `sessionPrincipal` into a ThreadLocal. Although the `SecurityContext` is already stored in a thread local, the default context is shared in multiple thread locals. Storing `sessionPrincipal` in another thread local would keep it only for the current thread even if the security context is shared. However, I don't like adding another thread local.

Maybe @arjantijms , @dmatej , @avpinchuk , @pzygielo, @hs536 you have some idea to improve it? Or it's OK to go with my simple solution?

About tests, I'm not sure if adding tests makes sense. The issue is random. Sometimes doesn't happen even after 1000 requests, especially on a slower machine. So the test would not be reliable and might not fail each time even if the behavior is broken.

### DIFF
--- a/nucleus/security/core/src/main/java/com/sun/enterprise/security/SecurityContext.java
+++ b/nucleus/security/core/src/main/java/com/sun/enterprise/security/SecurityContext.java
@@ -75,7 +75,7 @@ public class SecurityContext extends AbstractSecurityContext {
     // Did the client log in as or did the server generate the context
     private boolean serverGeneratedSecurityContext;
 
-    private ThreadLocal<Principal> sessionPrincipal = new ThreadLocal<>();
+    private final ThreadLocal<Principal> sessionPrincipal = new ThreadLocal<>();
 
     /*
      * This creates a new SecurityContext object. Note: that the docs for Subject state that the internal sets (eg. the
--- a/nucleus/security/core/src/main/java/com/sun/enterprise/security/SecurityContext.java
+++ b/nucleus/security/core/src/main/java/com/sun/enterprise/security/SecurityContext.java
@@ -75,7 +75,7 @@ public class SecurityContext extends AbstractSecurityContext {
     // Did the client log in as or did the server generate the context
     private boolean serverGeneratedSecurityContext;
 
-    private Principal sessionPrincipal;
+    private ThreadLocal<Principal> sessionPrincipal = new ThreadLocal<>();
 
     /*
      * This creates a new SecurityContext object. Note: that the docs for Subject state that the internal sets (eg. the
@@ -289,8 +289,7 @@ public class SecurityContext extends AbstractSecurityContext {
     public static SecurityContext getCurrent() {
         SecurityContext securityContext = currentSecurityContext.get();
         if (securityContext == null) {
-            securityContext = generateDefaultSecurityContext();
-            currentSecurityContext.set(securityContext);
+            securityContext = defaultSecurityContext;
         }
 
         return securityContext;
@@ -371,11 +370,15 @@ public class SecurityContext extends AbstractSecurityContext {
     }
 
     public Principal getSessionPrincipal() {
-        return sessionPrincipal;
+        return sessionPrincipal.get();
     }
 
     public void setSessionPrincipal(Principal sessionPrincipal) {
-        this.sessionPrincipal = sessionPrincipal;
+        if (sessionPrincipal != null) {
+            this.sessionPrincipal.set(sessionPrincipal);
+        } else {
+            this.sessionPrincipal.remove();
+        }
     }
 
     public Set<Principal> getPrincipalSet() {
--- a/nucleus/security/core/src/main/java/com/sun/enterprise/security/SecurityContext.java
+++ b/nucleus/security/core/src/main/java/com/sun/enterprise/security/SecurityContext.java
@@ -289,7 +289,8 @@ public class SecurityContext extends AbstractSecurityContext {
     public static SecurityContext getCurrent() {
         SecurityContext securityContext = currentSecurityContext.get();
         if (securityContext == null) {
-            securityContext = defaultSecurityContext;
+            securityContext = generateDefaultSecurityContext();
+            currentSecurityContext.set(securityContext);
         }
 
         return securityContext;


dmatej:
12:25:15       [exec] Caused by: java.lang.NullPointerException
12:25:15       [exec] 	at com.sun.s1peqe.security.ssl.converter.ejb.ConverterBean.myCallerPrincipal(ConverterBean.java:38)

hs536:
My apologies, I misunderstood. If the problem is that a specific thread is deleting (or modifying) the shared `defaultSecurityContext`'s `sessionPrincipal` (and this modification is allowed), wouldn't removing `static` from `defaultSecurityContext` resolve the issue...?

OndroMih:
> wouldn't removing static from defaultSecurityContext resolve the issue...?

I effectively did this with the current solution in this PR - a new security context is created in a thread local if it doesn't exist, instead of using the one in the static variable. It's not possible to just remove the `static` keyword, because the variable is access from static methods. 

However, I think the null pointer now is related to this code:

https://github.com/eclipse-ee4j/glassfish/blob/6f9081620de827f41652d1a0191726f5736b7491/nucleus/security/core/src/main/java/com/sun/enterprise/security/SecurityContext.java#L359

It checks whether the context is the same as the `defaultSecurityContext` context, which it isn't anymore, it's just a copy of it. And then it returns `initiator`, which is most probably `null` in this case.

I'll rework the solution to store `sessionPrincipal` into a thread-local so that it's different for each thread, while the context itself is the same for all threads, if it's the default one. That should keep the logic and still fix the issue with the race condition and random 403 responses.

arjantijms:
Maybe the field `serverGeneratedSecurityContext` should be in thread-local storage too? 

As mentioned in the description; having all these thread locals is kinda confusing when reading the code. The security context itself is in a thread local, but it's individual fields are too.

Probably we need to take a good look at this entire setup later again.

OndroMih:
Yes, I would review it later. Now, it’s clear what’s happening and my fix is simple and reliable.

With my solution, we have a thread-local in another thread local, maily because we need to store a mutable value in a structure which was designed to be immutable and sharable between multiple threads. Although securityContext is in thread local, the default security context is stored in multiple thread locals and this shared by mutliple threads. It’s OK if the structure is immutable. But after we added mutable state, it resulted in race conditions. If we make each security context unique and change rhe way it’s compared to the default context, then we can remove the nested thread local for sessionPrincipal.

nucleus/security/core/src/main/java/com/sun/enterprise/security/SecurityContext.java
```diff
@@ -75,7 +75,7 @@ public class SecurityContext extends AbstractSecurityContext {
     // Did the client log in as or did the server generate the context
     private boolean serverGeneratedSecurityContext;
 
-    private Principal sessionPrincipal;
+    private ThreadLocal<Principal> sessionPrincipal = new ThreadLocal<>();
```
avpinchuk, line 78: The `sessionPrincipal` may be final

OndroMih, line 78:
```suggestion
    private final ThreadLocal<Principal> sessionPrincipal = new ThreadLocal<>();
```

avpinchuk (approved):

arjantijms:
Also passed EE 10 Security (and other) TCKs: https://ci.eclipse.org/jakartaee-tck/view/EFTL-Certification-Jobs-10/job/eftl-jakartaeetck-run-standalone/153/

PR #25301, opened 2024-12-28
Fixes classloading of the generated classes for the shared EAR libraries

### LINKED ISSUE #25054: Deployment of EAR with Remote EJBs doesn't work
Deployment of EAR with Remote EJBs doesn't work

Hello Community,

first of all - thank you for your work. Since the glassfish AS development was moved to Eclipse, the progress is on another level. Kudos to you!

## Environment Details
* GlassFish Version 7.0.15:
* JDK version 17:
* OS Windows:
* Database Oracle/Postgres:

----------

## Problem Description
We're trying to migrate our application from glassfish 5.1.0 to the current version (7.0.15)
Unfortunately the deployment of our EAR fails because several Remote EJBs cannot be found by the ejb-container.
We've tried to track the problem down and found the commit, which has changed the behaviour of the ejb-container in such way, that the deployment fails:
Commit HASH: 6fbf8956e02e61911771d9e198ee18cb8383ca15
Commit Message: Fixes generated EJB remote interfaces loading
Class com.sun.ejb.codegen.ClassGenerator

Just one code line was replaced by this commit in the method:

 private static boolean useMethodHandles(final ClassLoader loader, final Class<?> anchorClass,
        final String targetPackageName)

old  version:
return Objects.equals(targetPackageName, anchorClass.getPackageName());
new version:
return Objects.equals(anchorClass.getPackage(), loader.getDefinedPackage(targetPackageName));

To be sure, that was the only cause, we've reverted this commit, built the ejb-container locally and the application can be deployed without any further problems.

Our Remote EJB Interfaces and the corresponding implementations are parts of different jars within the EAR. The package names of the Remote EJB Interfaces and their implementation can also differ. The partition is done to provide Remote EJB Interface definitions to other remote service consumer applications in the minimal way. 

Kind Regards

avpinchuk:
Fixes #25054.

Uses MethodHandles to define generated remote EJB proxy only if the anchor class (remote EJB interface) in the same package and the anchor class is visible to the application classloader (this may not be true for the Embedded GF).

Uses application classloader in all other cases.

### DIFF
--- a/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/codegen/ClassGenerator.java
+++ b/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/codegen/ClassGenerator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Eclipse Foundation and/or its affiliates. All rights reserved.
+ * Copyright (c) 2022, 2024 Eclipse Foundation and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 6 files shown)


arjantijms (approved):

OndroMih (approved):

dmatej (approved):

chicko12345:
Thank you very much!

PR #25307, opened 2025-01-05
Mask secrets in CommandInvokedEvent and Command logger

OndroMih:
This fixes a security issue in the Command Logger feature, which allowed to expose passwords used in some admin commands. This fix replaces passwords in the logged messages by "******".

### DIFF
--- a/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/CommandRunnerImpl.java
+++ b/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/CommandRunnerImpl.java
@@ -17,7 +17,6 @@
 
 package com.sun.enterprise.v3.admin;
 
-import com.sun.enterprise.admin.event.AdminCommandEventBrokerImpl;
 import com.sun.enterprise.admin.util.CachedCommandModel;
 import com.sun.enterprise.admin.util.ClusterOperationUtil;
 import com.sun.enterprise.admin.util.CommandSecurityChecker;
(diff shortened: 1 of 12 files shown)


nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/CommandRunnerImpl.java
```diff
@@ -1642,12 +1643,22 @@ public void executeFromCheckpoint(JobManager.Checkpoint checkpoint, boolean reve
...
         final CommandInvokedEvent event = new CommandInvokedEvent(
                 invocation.name(),
-                invocation.parameters(),
+                maskSecretParameters(invocation.parameters()),
                 subject);
         eventService.getCommandInvokedTopic()
                 .publish(event);
     }
 
+    private ParameterMap maskSecretParameters(ParameterMap parameters) {
```
pzygielo, line 1652: Could it be `static` and have unit test?

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/CommandRunnerImpl.java
```diff
@@ -1642,12 +1643,22 @@ public void executeFromCheckpoint(JobManager.Checkpoint checkpoint, boolean reve
     private void publishCommandInvokedEvent(ExecutionContext invocation, Subject subject) {
         final CommandInvokedEvent event = new CommandInvokedEvent(
                 invocation.name(),
-                invocation.parameters(),
+                maskSecretParameters(invocation.parameters()),
```
pzygielo, line 1646: The current failures suggest (to me), that `invocation.parameters()` can be `null`.

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/CommandRunnerImpl.java
OndroMih: OK. I will add this specific case into the CommandLoggerTest tests.

hs536 (approved):
LGTM

PR #25314, opened 2025-01-07
Set totalEntitySizeLimit to larger value

arjantijms:
It was decreased in JDK 24 and causes issues

Changeset in JDK: https://github.com/openjdk/jdk/commit/28c8729019292820f17002cc007305418f2d2676#diff-d5cb9e1e659383432d1b1a3a7ee0864815fbb2fcdf778a0e5ed52f2835048b5dL109

### DIFF
--- a/appserver/admin/template/src/main/resources/config/domain.xml
+++ b/appserver/admin/template/src/main/resources/config/domain.xml
@@ -160,6 +160,7 @@
         <jvm-options>-Djava.awt.headless=true</jvm-options>
         <jvm-options>-Djdk.corba.allowOutputStreamSubclass=true</jvm-options>
         <jvm-options>-Djdk.tls.rejectClientInitiatedRenegotiation=true</jvm-options>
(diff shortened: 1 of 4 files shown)


pzygielo:
- https://bugs.openjdk.org/browse/JDK-8343004

OndroMih:
I would rather implement this in [JvmOptions](https://github.com/eclipse-ee4j/glassfish/blob/a2a19938cd6e07fe95fcaccb36d77b02670d22f9/nucleus/admin/launcher/src/main/java/com/sun/enterprise/admin/launcher/JvmOptions.java#L51) so that the default value is used also when upgrading from an older domain, which doesn't contain the JVM option. And we should document it as the default value somewhere in GlassFish docs.

We can also keep it in the default domain.xml, so that it's easier to see the default value in the stock GlassFish domain.

If this is added only to domain.xml, it will apply only for new installations and not for upgraded installations.

PR #25356, opened 2025-02-01
Fix principal in Jakarta SecurityContext when using SSO

OndroMih:
WebPrincipal was not serializable because of this fix: https://github.com/eclipse-ee4j/glassfish/pull/25290, which caused exceptions in SSO mechanism, like this:

```
java.lang.IllegalStateException: java.io.NotSerializableException: java.lang.ThreadLocal
	at org.glassfish.web.ha.authenticator.HASingleSignOnEntry.convertToByteArray(HASingleSignOnEntry.java:167)
...
Caused by: java.io.NotSerializableException: java.lang.ThreadLocal
...
	at java.base/java.io.ObjectOutputStream.writeObject(ObjectOutputStream.java:358)
	at org.glassfish.web.ha.authenticator.HASingleSignOnEntry.convertToByteArray(HASingleSignOnEntry.java:163)
	... 33 more
```

This PR moves the thread local variable to the static context so that it's not serialized together with securityContext instance. The value is only relevant within a scope of a specific request, and thus within a specific thread, so it doesn't have to be attached to a specific security context instance.

### DIFF
--- a/appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/GlassFishTestEnvironment.java
+++ b/appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/GlassFishTestEnvironment.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2024 Eclipse Foundation and/or its affiliates.
+ * Copyright (c) 2022, 2025 Eclipse Foundation and/or its affiliates.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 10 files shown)


appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/GlassFishTestEnvironment.java
```diff
@@ -415,4 +449,37 @@ private interface IOAction {
 
         void execute() throws IOException;
     }
+
+    private static HttpClient newInsecureHttpClient() throws KeyManagementException, NoSuchAlgorithmException {
+        if (client == null) {
+            // Java 17 doesn't allow to close http client, so we reuse a global one.
```
dmatej, line 455: See 
* https://stackoverflow.com/questions/53919721/close-java-http-client
* https://bugs.openjdk.org/browse/JDK-8304165

appserver/tests/application/src/main/java/org/glassfish/main/test/app/security/jakartasecurity/CustomPrincipal.java
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright (c) 2015, 2020 Oracle and/or its affiliates. All rights reserved.
```
pzygielo, line 2: Are some files in this PR resurrected or new?

OndroMih: You're right. I think I copied some older example and modified it. I fixed the copyright headers in all files now.

nucleus/security/core/src/main/java/com/sun/enterprise/security/SecurityContext.java
```diff
@@ -72,11 +72,12 @@ public class SecurityContext extends AbstractSecurityContext {
 
     private static AuthPermission doAsPrivilegedPerm = new AuthPermission("doAsPrivileged");
 
+    // this is static because it's a thread local, which isn't serializable
+    private static ThreadLocal<Principal> sessionPrincipal = new ThreadLocal<>();
```
hs536, line 76: Why did you remove `final`?

OndroMih: You're right, @hs536 , thank you for spotting this.

I must have removed final when I was trying a different approach, but then forgot to put it back. I added it back now.

hs536 (approved):

PR #25365, opened 2025-02-05
Ensure order of logged messages, fix a few flaky tests

### LINKED ISSUE #25349: Flaky `GlassFishLogManagerLifeCycleTest.reconfigureWithoutResolvingLevelsWithincompleteconfiguration`
https://github.com/eclipse-ee4j/glassfish/blob/a56d44c13810baf437bba73cbac8e648751c78f4/nucleus/glassfish-jul-extension/src/test/java/org/glassfish/main/jul/GlassFishLogManagerLifeCycleTest.java#L194-L221

Have seen recently:
```
2025-01-26T08:21:39.9703043Z Both records must pass via startup queue (4 failures)
2025-01-26T08:21:39.9704392Z >--java.lang.AssertionError: record 0
2025-01-26T08:21:39.9704815Z Expected: <FINEST>
2025-01-26T08:21:39.9705101Z      but: was <INFO>
2025-01-26T08:21:39.9705451Z >--java.lang.AssertionError: record 0
2025-01-26T08:21:39.9705854Z Expected: "message0"
2025-01-26T08:21:39.9706155Z      but: was "message1"
2025-01-26T08:21:39.9706532Z >--java.lang.AssertionError: record 1
2025-01-26T08:21:39.9706915Z Expected: <INFO>
2025-01-26T08:21:39.9707189Z      but: was <FINEST>
2025-01-26T08:21:39.9707521Z >--java.lang.AssertionError: record 1
2025-01-26T08:21:39.9707903Z Expected: "message1"
2025-01-26T08:21:39.9708195Z      but: was "message0"
```

Timing issue?

Looks like records are collected in order different than expected by assertions.
From the message _Both records must pass via startup queue_ I'd not suspect the order to be crucial. I believe that implementation is rather fine here and it's test that should follow.

_Originally posted by @pzygielo in https://github.com/eclipse-ee4j/glassfish/pull/24046#discussion_r1929708843_

OndroMih:
Fixes #25349 
Fixes an issue that caused the `GlassFishLogHandlerTest.roll` test to fail sometimes - the messages produced when rolling log files could be logged in reversed order.

Fixes flaky `GlassFishLogManagerLifeCycleTest.reconfigureWithoutResolvingLevelsWithincompleteconfiguration` test, which produced log messages in undeterministic order but expected them in order. The fix is similar, but this time in the test, not in production code.

### DIFF
--- a/nucleus/glassfish-jul-extension/src/test/java/org/glassfish/main/jul/GlassFishLogManagerLifeCycleTest.java
+++ b/nucleus/glassfish-jul-extension/src/test/java/org/glassfish/main/jul/GlassFishLogManagerLifeCycleTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2024 Eclipse Foundation and/or its affiliates. All rights reserved.
+ * Copyright (c) 2022, 2025 Eclipse Foundation and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -18,6 +18,8 @@ package org.glassfish.main.jul;
 
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
@@ -71,6 +73,7 @@ public class GlassFishLogManagerLifeCycleTest {
     private static GlassFishLogManagerConfiguration originalCfg;
     private static boolean originalLevelResolution;
     private static CompletableFuture<Void> process;
+    private ExecutorService sequentialExecutor = Executors.newSingleThreadExecutor();
 
     @BeforeAll
     public static void backupConfiguration() {
@@ -133,6 +136,7 @@ public class GlassFishLogManagerLifeCycleTest {
             () -> assertEquals(GlassFishLoggingStatus.CONFIGURING, MANAGER.getLoggingStatus()),
             () -> assertNull(COLLECTOR.pop(), "COLLECTOR should be empty after reconfiguration started"));
         doLog(Level.INFO, "Log after reconfiguration started", 0);
+        Thread.sleep(10L);
         assertNull(COLLECTOR.pop(), "COLLECTOR should be empty after reconfiguration started even if we log again");
     }
 
@@ -269,7 +273,7 @@ public class GlassFishLogManagerLifeCycleTest {
      * @throws Exception
      */
     private void doLog(final Level level, final String message, final int sleepMillis) throws Exception {
-        new Thread(() -> LOG.log(level, message)).start();
+        sequentialExecutor.submit(() -> LOG.log(level, message));
         if (sleepMillis > 0) {
             Thread.sleep(sleepMillis);
         }
--- a/nucleus/glassfish-jul-extension/src/test/java/org/glassfish/main/jul/handler/GlassFishLogHandlerTest.java
+++ b/nucleus/glassfish-jul-extension/src/test/java/org/glassfish/main/jul/handler/GlassFishLogHandlerTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2024 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022, 2025 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -18,6 +18,7 @@ package org.glassfish.main.jul.handler;
 
 import java.io.File;
 import java.nio.file.Files;
+import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.LogManager;
 import java.util.logging.Logger;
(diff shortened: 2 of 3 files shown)


nucleus/glassfish-jul-extension/src/test/java/org/glassfish/main/jul/GlassFishLogManagerLifeCycleTest.java
```diff
@@ -133,6 +136,7 @@ public void startReconfiguration() throws Exception {
             () -> assertEquals(GlassFishLoggingStatus.CONFIGURING, MANAGER.getLoggingStatus()),
             () -> assertNull(COLLECTOR.pop(), "COLLECTOR should be empty after reconfiguration started"));
         doLog(Level.INFO, "Log after reconfiguration started", 0);
+        Thread.sleep(10L);
```
dmatej, line 139: In all these tests should not be any sleeps, they make these tests ineffective. Also the line before has 0 sleep time, while the thread started before the sleep (but still could be executed before or after it).
I have to think what kind of error I intended to catch by this.

OndroMih: I added this sleep to ensure that the logging actually happens before the assertion on the next line. Otherwise the assertion doesn't have any value - it could be asserted before the log happens and then the test passes.

Should I remove the sleep here?

OndroMih:
> The format does not recognize colon between keyword and issue - hence the commit links to issue, however is not going to close the issue if merged.

Thanks, @pzygielo . I amended the commit message.

nucleus/glassfish-jul-extension/src/test/java/org/glassfish/main/jul/GlassFishLogManagerLifeCycleTest.java
pzygielo: Could `doLog` return something that the code here could check if it's safe to proceed?

OndroMih: I have no idea. Now it looks to me that @dmatej  intended to do the check immediately. It’s supposed to test locking, so maybe all should run as soon as possible. Although it would be good to clarify what the tests intends to verify.

OndroMih: Do you mean the executor I created here in rhe LogFileManager? It creates a thread only when it’s needed and disposes it after 100ms if it’s inactive. The only thing that is kept all the time is the executor itself, with no threads most of the time.

A better approach would be as you suggest - it would be enough to create a single threaded executor in the scope of roll method and use it for all async logging and then shut it down at the end of the roll method.

nucleus/glassfish-jul-extension/src/test/java/org/glassfish/main/jul/handler/GlassFishLogHandlerTest.java
```diff
@@ -164,10 +168,11 @@ public void roll() throws Exception {
         );
         handler.publish(new GlassFishLogRecord(Level.SEVERE, "File two, line two", false));
         Thread.sleep(MILLIS_FOR_PUMP);
+        final List<String> fileLines3 = Files.readAllLines(logFile.toPath());
```
dmatej, line 171: If the file doesn't exist, this will cause IOException, so the assertion will not make sense.

nucleus/glassfish-jul-extension/src/test/java/org/glassfish/main/jul/handler/GlassFishLogHandlerTest.java
```diff
@@ -152,10 +155,11 @@ public void roll() throws Exception {
         handler.roll();
         // There will be an asynchronous thread.
         Thread.sleep(10L);
+        final List<String> fileLines2 = Files.readAllLines(logFile.toPath());
         assertAll(
             () -> assertTrue(handler.isReady(), "handler.ready"),
-            () -> assertTrue(handler.getConfiguration().getLogFile().exists(), "file exists"),
-            () -> assertThat("file content", Files.readAllLines(handler.getConfiguration().getLogFile().toPath()),
+            () -> assertTrue(logFile.exists(), "file exists"),
+            () -> assertThat("file content, filename=" + logFile + ", content=" + fileLines2 , fileLines2,
```
dmatej, line 162: Putting whole content to "reason" argument is bit too much ... the assertion would provide quite good explanation, ie:

```
java.lang.AssertionError: file content
Expected: iterable containing [a string containing "INFO", "Rolling the file ", ".log", "output was originally enabled: true" in order, a string containing "INFO", "Araaachiving file ", ".log", " to ", ".log_" in order, a string containing "SEVERE", "main", "File two, line two" in order]
     but: item 1: was "21:56:32.278209    INFO er-Async-Info-Logger                                                             . Archiving file /tmp/org.glassfish.main.jul.handler.GlassFishLogHandlerTest11276075090100096775.log to /tmp/org.glassfish.main.jul.handler.GlassFishLogHandlerTest11276075090100096775.log_2025-02-05T21-56-32"
```
but maybe adding the file to the message is a good idea, perhaps directly to assertAll.

OndroMih:
I'm closing this, in favor of https://github.com/eclipse-ee4j/glassfish/pull/25368, which fixes the same issues but better.

PR #25376, opened 2025-02-17
Update pom versions and split up build pipeline some more

arjantijms:


dmatej:
You should avoid such changes in this branch, sooner or later they will start conflicting. Now can you cherrypick it for master too?

PR #25382, opened 2025-02-20
Option to prefer env vars when resolving variables in JVM options 

### LINKED ISSUE #25381: Improvement: Prefer env vars in Docker and cloud environments
In Docker and containerized cloud environments, environment variables are used to customize the deployment. This goes against the traditional approach, where variables are configured in GlassFish installation and they should be isolated from the environment.

This request is to allow configuring system properties in GlassFish using environment variables. This should be an optional feature, disabled by default via a system property or configuration attribute in domain.xml.

**Note:** GlassFish already supports using variable references. Currently, they are resolved only from system properties. However, JVM options are resolved by the launcher both from system properties defined in domain.xml and from environment variables, before the server process is started. So, it's already possible to create a JVM option, which refers an environment variable. During the resolution, system properties take precedence over environment variables, so it's not possible to define the default value as a system property and overwrite it using environment variable. It would be enough to switch the order of evaluation, .e.g. with a system property.

OndroMih:
Fixes https://github.com/eclipse-ee4j/glassfish/issues/25381.

Adds support for system property `org.glassfish.envPreferredToProperties`. When it's set to `true` (e.g. in server-config or in JVM options), then environment variables take precedence in resolution of variable references in JVM options.

Until now, it was possible to resolve a variable in JVM options using an environment variable, if the variable was not defined in GlassFish configuration. A default value cannot be provided, an environment variable would be then hidden. In Docker, a default value could be provided by a default value of an environment variable.

With this change, it's possible to redefine variables in JVM options via environment variables even if they are already in GlassFish configuration, e.g. in a system property. In Docker, this provides an alternative way to configure default properties and override them with environment variables.

It's also possible to directly override a value of a system property or option that is specified in JVM options. This was not possible before at all.

Examples:

**I want to:**

* allow to define a property "application.name" with environment variable APPNAME
* provide the default value as a system property

This can be done now with:

* create  JVM option:  `asadmin create-jvm-options '-Dapplication.name=${APPNAME}'`
* create a system property for the default value: `create-system-properties APPNAME=application`

The default value of property `application.name` is "application". If GlassFish started with environment variable `APPNAME=myapp`, then the value of `application.name` is `myapp`.

**I want to**

* specify an existing property "app.name" in JVM options with an environment variable

This can be done with:

* create JVM option "application.name": `asadmin create-jvm-options '-Dapplication.name=defaultapp`
* specify the environment variable APPLICATION_NAME=myapp
* start GlassFish - the system property `application.name` will be set to "myapp", the main JVM process will be started with `-Dapplication.name=myapp`

### DIFF
--- a/nucleus/admin/launcher/src/main/java/com/sun/enterprise/admin/launcher/GFLauncher.java
+++ b/nucleus/admin/launcher/src/main/java/com/sun/enterprise/admin/launcher/GFLauncher.java
@@ -14,7 +14,6 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  */
-
 package com.sun.enterprise.admin.launcher;
 
 import com.sun.enterprise.admin.launcher.CommandLine.CommandFormat;
@@ -45,7 +44,9 @@ import java.util.List;
 import java.util.ListIterator;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import static com.sun.enterprise.admin.launcher.GFLauncher.LaunchType.fake;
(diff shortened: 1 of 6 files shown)


nucleus/admin/launcher/src/main/java/com/sun/enterprise/admin/launcher/GFLauncher.java
```diff
@@ -660,6 +694,36 @@ private void resolveAllTokens() {
...
+                        entry.setValue(envValue);
+                    }
+                }
+            }
+
+        }
+    }
+
+    private static Boolean isPreferEnvOverProperties(Map<String, String> properties) {
+        return Optional.ofNullable(properties.get(PREFER_ENV_VARS_OVER_PROPERTIES))
```
avpinchuk, line 722: Why not use `Boolean.parseBoolean()` instead?

OndroMih: Of course, thanks! I forgot about this method.

avpinchuk:
Looks good. Please fix checkstyle violations.

avpinchuk (approved):

PR #25390, opened 2025-02-26
Increased timeout to pass tests on GitHub CI too

dmatej:
- Seems it is a bit slower than Eclipse Jenkins CI


OndroMih (approved):

chabala:
So, PR #25407 reverts this, correct?

PR #25398, opened 2025-03-03
Speed up deployments - optimization in finding files in deployment packages

### LINKED ISSUE #25397: Improve autodeploy performance
Improve autodeploy performance

Glassfish 7.0.22
OS: Windows 11
openjdk version "17.0.8.1" 2023-08-24
OpenJDK Runtime Environment Temurin-17.0.8.1+1 (build 17.0.8.1+1)
OpenJDK 64-Bit Server VM Temurin-17.0.8.1+1 (build 17.0.8.1+1, mixed mode, sharing)

Ear file size: 90MB, (includes Primefaces/Guava/Grpc/Hazelcast 40MB jars), 6 war files, 75 ejb jar files.
Deployment takes about 111 seconds.
Of these 111 seconds about 1/3rd of the time

Profiling autodeploy with Yourkit on a Windows instance shows:

- Deployment takes 150 seconds, due to Yourkit overhead.
- About 49 of the 150 seconds is spent in isFile and isDirectory

- Method list, ordered by Own Time
Image

Yourkit snapshot of CPU profiling shows:
Image

Most calls are coming via FileArchive:
Image

Debugging shows for 1 ear file full ear file is analysed about 30+ times by the 
**com.sun.enterprise.deploy.shared.FileArchive.getListOfFiles(File, Logger)**
method.

Tested a fix by:
- Adding a cache in FileArchive reduces 30+ calls to about 5 calls
- Rewriting recursive getListOfFiles to use Files.walkFileTree, and leave logic over to JVM
improves autodeploy time from 110 seconds to 92 seconds.

I once created a similar Payara issue, long time ago, which was integrated into 5.191
The commit was: https://github.com/payara/Payara/pull/3616
Issue with this fix is
- I see two separate threads starting to inspect the ear file once file is copied into auto deploy (AutoDeploy thread and some 'scan for changes' thread: TODO: add correct name)
- How to add cache? static variable in FileArchive?
- How to clean cache after deployment? org.glassfish.deployment.common.DeploymentContextImpl.postDeployClean(boolean) -> is called many times during single deployment, expecting only one call with isFinalClean, but no isFinalClean is called about 80 times for this ear deployment, possibly one time for each module in the ear?)

escay:
Issue #25397 use java.nio:
- Files.walkFileTree instead of recursively finding files.
- Use Path instead of URI. 

and use try with resources.

This change makes deployment times improve on my Windows 11 pc from 110 seconds to 100 seconds.
Not yet adding caching logic, which would improve deployment times for the same ear file around 92 seconds.

Yourkit result after this change:
- 10+ seconds less time spent in java.io.FileSystem.hasBooleanAttributes(File, int)
- (Was 49 seconds, now 30 seconds, but Yourkit is running, without Yourkit about 10 second improvement)

image
image

### DIFF
--- a/nucleus/deployment/admin/src/main/java/org/glassfish/deployment/admin/DeployCommand.java
+++ b/nucleus/deployment/admin/src/main/java/org/glassfish/deployment/admin/DeployCommand.java
@@ -583,6 +583,10 @@ public class DeployCommand extends DeployCommandParameters implements AdminComma
             if (deploymentContext != null) {
                 deploymentContext.postDeployClean(true);
             }
+
+            // Clear FileArchive cache that has been filled during deployment. Clear it exactly one time after deployment.
+            // Cannot use postDeployClean implementation, it is called multiple times (even with value true) during deployment.
+            com.sun.enterprise.deploy.shared.FileArchive.clearCache();
         }
     }
 
--- a/nucleus/deployment/admin/src/main/java/org/glassfish/deployment/admin/DeployCommand.java
+++ b/nucleus/deployment/admin/src/main/java/org/glassfish/deployment/admin/DeployCommand.java
@@ -583,10 +583,6 @@ public class DeployCommand extends DeployCommandParameters implements AdminComma
             if (deploymentContext != null) {
                 deploymentContext.postDeployClean(true);
             }
-
-            // Clear FileArchive cache that has been filled during deployment. Clear it exactly one time after deployment.
-            // Cannot use postDeployClean implementation, it is called multiple times (even with value true) during deployment.
-            com.sun.enterprise.deploy.shared.FileArchive.clearCache();
         }
     }
 
--- a/nucleus/deployment/common/src/main/java/com/sun/enterprise/deploy/shared/FileArchive.java
+++ b/nucleus/deployment/common/src/main/java/com/sun/enterprise/deploy/shared/FileArchive.java
@@ -44,11 +44,9 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Enumeration;
-import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.ListIterator;
-import java.util.Map;
 import java.util.jar.JarFile;
 import java.util.jar.Manifest;
 import java.util.logging.Level;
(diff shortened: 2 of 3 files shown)


escay:
Hmmm.. build failed due to:

```
[2025-03-05T20:18:41.646419+01:00] [GF 7.0.23-SNAPSHOT] [SEVERE] [] [jakarta.enterprise.system.tools.deployment.common] [tid: _ThreadID=63 _ThreadName=admin-listener(2)] [levelValue: 1000] [[
  Exception while invoking class org.glassfish.weld.WeldDeployer load method
java.lang.NullPointerException: Cannot invoke "org.glassfish.api.deployment.archive.ReadableArchive.close()" because "weblibJarArchive" is null
	at org.glassfish.weld.BeanDeploymentArchiveImpl.populate(BeanDeploymentArchiveImpl.java:481)
	at org.glassfish.weld.BeanDeploymentArchiveImpl.<init>(BeanDeploymentArchiveImpl.java:146)
	at org.glassfish.weld.RootBeanDeploymentArchive.createModuleBda(RootBeanDeploymentArchive.java:68)
	at org.glassfish.weld.RootBeanDeploymentArchive.<init>(RootBeanDeploymentArchive.java:64)
	at org.glassfish.weld.RootBeanDeploymentArchive.<init>(RootBeanDeploymentArchive.java:56)
```
For which Payara has added a NPE check in https://github.com/payara/Payara/commit/d2b335387d667680ffa7a5312141a1c544607ecc

Adding a NPE check in BeanDeploymentArchiveImpl makes the build succeed, like:
`if (weblibJarArchive != null && weblibJarArchive.exists(META_INF_BEANS_XML)) {`
but this feels like opening a can of worms. I lack the knowledge of understanding all the places where FileArchive is used and wether or not caching would be valid.

I will undo the cache commit part.

escay:
I do not know why the last test failed. Locally it runs ok. And the first build was also ok.
Ubuntu build: GlassFishLogHandlerTest - some timeout, right at the start?
Windows build: DetachAttachITest.uptimePeriodically NumberFormatException: For input string: "Up 1 mins 15 secs"

Let me know what to do. Regarding the fix: I do not plan any other changes.

dmatej (approved):

hs536 (approved):

PR #25403, opened 2025-03-10
Streamline and somewhat update MP TCKs

arjantijms:


appserver/tests/tck/microprofile/pom.xml
```diff
@@ -19,36 +19,44 @@
 
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
-
+     
     <parent>
-        <groupId>org.glassfish.main.tests.tck</groupId>
-        <artifactId>tck</artifactId>
-        <version>8.0.0-SNAPSHOT</version>
+        <groupId>org.eclipse.ee4j</groupId>
```
dmatej, line 24: Why?

appserver/tests/tck/microprofile/jwt/tck-suite.xml
```diff
@@ -0,0 +1,91 @@
+<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd" >
```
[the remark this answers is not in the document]
arjantijms: It's to be consistent with the other two mp tcks; they do the same thing

appserver/tests/tck/microprofile/pom.xml
arjantijms: I was trying to run the TCK against arbitirary GlassFish versions. With the project parent it insisted upon having basically all of GlassFisg build in the version that the project happened to be. This is unnecessary when testing with such specific version.

PR #25404, opened 2025-03-14
Merge 7.0.23 branch into master

arjantijms:


dmatej:
> @dmatej can you take a look at the failure?

I have found the cause, I will fix it on Monday or earlier if I find time. It is quite funny race condition - the check for the count of lines in the file is much faster than "producer" writing to the file, so the "consumer" allocates 8K+ buffer in the heap for every attempt to read the file. The `onSpinWait` then runs really quickly (might be a Java or CPU issue, it is out of our control), so then GC evacuation comes into play several times ... and then it is too late. In your case the garbage collecting took 9 seconds.

I am yet playing with that, changing thread priorities etc, because it is not happening always and I could not reproduce it locally not even once on AMD, but I understood from the code, what is wrong with it.

I am thinking how could file system implementation resolve such cases, when consumer tries to read the fail many times in the row, while another process tries to write into it ... caching? Locking to avoid switches and conflicts? Same will probably apply to reading the PID file.

OndroMih (approved):

OndroMih (changes requested):
Should change version to a SNAPSHOT version, right?

dmatej:
> Should change version to a SNAPSHOT version, right?

I have fixed that and replaced the last commit by mine.

hs536 (approved):

OndroMih (approved):

PR #25407, opened 2025-03-18
Fixed heavy need of garbage collector in GlassFishLogHandlerTest

dmatej:
- Every request to count lines causes creation of 8K buffer.
- Thread.onSpinWait() is usually fast
- Instead of sleeping we first call lighther method guarding the more expensive read of the file content.


dmatej:
There is yet some issue with starting+stopping instances in admin-tests, but it doesn't seem the same problem.

OndroMih (approved):

PR #25420, opened 2025-03-22
Fixes in runnable GlassFish Embedded for add-library command

OndroMih:
1. Dynamically get the common classloader before each task to handle config changes events

Fixes the following scenario:

* Initially, there are no additional JARs in the common classloader, which turns on lazy initialization and the parent classloader is returned instead
* The executor to handle config changes is initialized with the parent of the common classloader
* When add-library is executed, it adds a library to the common classloader, which triggers full initialization of the common classloader
* The executor keeps the reference to the common CL parent instead of refreshing the reference to the fully initialized common CL
* config change handler doesn't have access to the common CL, so setting Grizzly executor class fails to initialize a class in the added library

This is fixed by 2nd commit https://github.com/eclipse-ee4j/glassfish/pull/25420/commits/c2304627105a7385e090925e3e058f3d92606561 to refresh the common CL before each task is executed by the executor.

2. Don't give an ugly error in logs when calling add-library from runnable GlassFish Embedded. By default, modifications to domain.xml are disabled in GF Embedded and add-library command shouldn't touch it. The error didn't have any impact on execution, it was just logged and all continued without a problem

Fixed by 1st commit https://github.com/eclipse-ee4j/glassfish/pull/25420/commits/87a96734f62589c5ed50079090ae698dd411dbaf .

All other commits are only related to test code or copyrights.


appserver/tests/embedded/basic/config/src/test/java/org/glassfish/tests/embedded/basic/config/ConfigTest.java
```diff
@@ -55,9 +56,15 @@ static void bootStrap() throws GlassFishException {
 
     @ParameterizedTest
     @ArgumentsSource(ConfigFileArgumentsProvider.class)
-    void testConfigFile(String configFile) throws Exception {
+    void testConfigFile(Object configFile) throws Exception {
         GlassFishProperties gfp = new GlassFishProperties();
-        gfp.setConfigFileURI(configFile);
+        if (configFile instanceof String) {
+            gfp.setConfigFile(new File((String)configFile));
```
pzygielo, line 62: Is this branch used?

OndroMih: Yes, this is used with the argument "domain.xml" created by ConfigFileArgumentsProvider.

pzygielo: :grinning: Now I can see. 
:+1:

appserver/tests/embedded/runnable/src/test/java/org/glassfish/tests/embedded/runnable/GfEmbeddedUtils.java
```diff
@@ -0,0 +1,55 @@
...
+        }
+        return new ProcessBuilder()
+                .redirectOutput(ProcessBuilder.Redirect.INHERIT)
+                .redirectError(ProcessBuilder.Redirect.PIPE)
+                .command(arguments)
+                .start();
+    }
+
+    public static Stream<String> outputToStreamOfLines(Process gfEmbeddedProcess) {
+        InputStream gfEmbeddedOutput = gfEmbeddedProcess.getErrorStream();
```
pzygielo, line 49: Why not `.errorReader`?

OndroMih: You're right, thank you. I missed this method.

OndroMih: Actually, I have to revert back to `getErrorStream()`, because `errorReader` is not in Java 11. But thanks for the hint, I learnt something new anyway.

pzygielo: I've seen `@since 17` for `errorReader`, but have forgotten GF7 is still at 11 only. :disappointed:

dmatej (approved):

PR #25436, opened 2025-04-01
Revisited scripts around asenv

dmatej:
### Changes
- Backported "in advance" from future 7.1 branch and rewritten to reduce duplication and to fix all related issues.
- All linux scripts use bash now
- All scripts use the same way of AS_INSTALL resolution
- Removed mentions about Java 1.6
- Fixed %t resolving on Windows (ie JDK log formatting option) which has to be  escaped.
- Added config.sh script sourced by all bash scripts
  - Loads asenv.conf
  - Sets AS_INSTALL
- AS_INSTALL removed from asenv.conf
   - it doesn't make sense here
   - it overrides values set in scripts
   - it was just a relative path
 - Added test to cover GF starts on Linux and Windows (depending on CI operating system)

### Original state
- GlassFish could not start when on path with spaces, both Linux and Windows
- Inconsistent scripts, some were patched in the past, some respected JAVA_HOME, some did not.
- Missing test for paths with spaces

### DIFF
--- a/appserver/appclient/client/appclient-scripts/src/main/resources/glassfish/bin/appclient.bat
+++ b/appserver/appclient/client/appclient-scripts/src/main/resources/glassfish/bin/appclient.bat
@@ -1,5 +1,6 @@
 @echo off
 REM
+REM  Copyright (c) 2024, 2025 Contributors to the Eclipse Foundation
 REM  Copyright (c) 2018 Oracle and/or its affiliates. All rights reserved.
 REM
 REM  This program and the accompanying materials are made available under the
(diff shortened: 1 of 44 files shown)


appserver/appclient/client/appclient-scripts/src/main/resources/glassfish/bin/appclient.bat
```diff
@@ -65,18 +66,11 @@ rem issue the error so we don't have to do i18n of our own message from the scri
...
-rem Convert the java.exe path and the classpath path to
-rem Windows "short" versions - with no spaces - so the
-rem for /F statement below will work correctly.  Spaces cause
-rem it great troubles.
-rem
-for %%a in ("%JAVA%") do set ACCJava=%%~sa%
-for %%a in ("%_AS_INSTALL%/lib/gf-client.jar") do set XCLASSPATH=%%~sa
-for /F "usebackq tokens=*" %%a in (`%ACCJava% -classpath %XCLASSPATH% org.glassfish.appclient.client.CLIBootstrap`) do set javaCmd=%%a
-%javaCmd%
+FOR /F "delims=" %i IN ("%JAVA%" -classpath "%AS_INSTALL%\lib\gf-client.jar" org.glassfish.appclient.client.CLIBootstrap %*) DO set CMD=%i
```
OndroMih, line 74: Can you keep the comment or add a new one to clarify what this does now?

appserver/distributions/glassfish-common/src/main/resources/bin/asadmin
```diff
@@ -15,17 +16,15 @@
...
+export AS_INSTALL="$(dirname "$(realpath -s "$0")")/../glassfish"
 case "`uname`" in
   CYGWIN*) AS_INSTALL=`cygpath --windows $AS_INSTALL`
 esac
-AS_INSTALL_LIB="$AS_INSTALL/lib"
 . "${AS_INSTALL}/config/asenv.conf"
+
 JAVA=java
-#Depends upon Java from ../config/asenv.conf
 if [ ${AS_JAVA} ]; then
```
OndroMih, line 26: Can we keep the comment "Depends upon Java from ../config/asenv.conf" ?

appserver/distributions/glassfish-common/src/main/resources/bin/asadmin.bat
```diff
@@ -15,20 +16,21 @@ REM
...
 echo "Unable to enable extensions"
 exit /B 1
+
 :ok
-call "%~dp0..\glassfish\config\asenv.bat"
+set AS_INSTALL=%~dp0..\glassfish
+call "%AS_INSTALL%\config\asenv.bat"
 if "%AS_JAVA%x" == "x" goto UsePath
-set JAVA="%AS_JAVA%\bin\java"
+set JAVA=%AS_JAVA%\bin\java
```
OndroMih, line 29: Quotes really not needed here?

appserver/distributions/glassfish-common/src/main/resources/bin/stopserv
```diff
@@ -15,7 +16,16 @@
...
 
-exec java -jar "$AS_INSTALL_LIB/admin-cli.jar" stop-domain "$@"
+case "`uname`" in
+  CYGWIN*) AS_INSTALL=`cygpath --windows $AS_INSTALL`
+esac
+. "${AS_INSTALL}/config/asenv.conf"
+
+JAVA=java
+if [ ${AS_JAVA} ]; then
+    JAVA=${AS_JAVA}/bin/java
```
OndroMih, line 28: Should be in quotes in case contains spaces:

```suggestion
    JAVA="${AS_JAVA}"/bin/java
```

appserver/distributions/glassfish-common/src/main/resources/bin/stopserv
```diff
@@ -15,7 +16,16 @@
 # SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
 #
 
-AS_INSTALL=`dirname "$0"`/../glassfish
-AS_INSTALL_LIB="$AS_INSTALL/modules"
+export AS_INSTALL="$(dirname "$(realpath -s "$0")")/../glassfish"
 
-exec java -jar "$AS_INSTALL_LIB/admin-cli.jar" stop-domain "$@"
+case "`uname`" in
+  CYGWIN*) AS_INSTALL=`cygpath --windows $AS_INSTALL`
```
OndroMih, line 22: AS_INSTALL should be in quotes as it can contain spaces:

```suggestion
  CYGWIN*) AS_INSTALL=`cygpath --windows "$AS_INSTALL"`
```

appserver/distributions/glassfish-common/src/main/resources/bin/stopserv.bat
```diff
@@ -16,4 +17,21 @@ REM  SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
...
+setlocal ENABLEEXTENSIONS
+if ERRORLEVEL 0 goto ok
+echo "Unable to enable extensions"
+exit /B 1
+
+:ok
+set AS_INSTALL=%~dp0..
+call "%AS_INSTALL%\config\asenv.bat"
+if "%AS_JAVA%x" == "x" goto UsePath
+set JAVA=%AS_JAVA%\bin\java
```
OndroMih, line 30: I guess this should be in quotes in case it contains spaces:
```suggestion
set JAVA="%AS_JAVA%"\bin\java

```

OndroMih (commented):
A few comments, I'll continue reviewing later

appserver/distributions/glassfish-common/src/main/resources/bin/asadmin
dmatej: The comment was false.

appserver/distributions/glassfish-common/src/main/resources/bin/stopserv
dmatej: Not needed here, but it is needed below when using exec, if I remember well.

appserver/appclient/client/appclient-scripts/src/main/resources/glassfish/bin/appclient.bat
dmatej: I asked Copilot for explanation, I think it is better than the original one and better than mine :+1:

appserver/distributions/glassfish-common/src/main/resources/bin/asadmin.bat
dmatej: Ok, I learned how it is done in BAT file - you wrap whole argument of the SET command. So the original was wrong, this was better, but safest option is to wrap it all.

appserver/distributions/glassfish-common/src/main/resources/bin/stopserv.bat
dmatej: As I wrote in another comment, the correct solution is `set "JAVA=%AS_JAVA%\bin\java"`

dmatej:
Seems it passed on CI, however I domain start on my Windows now failed.

appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/asadmin/Asadmin.java
```diff
@@ -224,8 +223,10 @@ private AsadminResult exec(final int timeout, final boolean detachedAndTerse, fi
         command.add(asadmin.getAbsolutePath());
         command.add("--user");
         command.add(adminUser);
-        command.add("--passwordfile");
-        command.add(getPasswordFile().getAbsolutePath());
+        if (getPasswordFile() != null) {
```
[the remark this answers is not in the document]
dmatej: It is for testing, and as we by default distribute GlassFish without any password, it would not make much sense - it doesn't lead to any user/developer action. The test I added is minimalistic, just creates an empty domain without password and starts and stops it on path with spaces to check that is does work.

nucleus/distributions/nucleus-common/src/main/resources/config/config.sh
```diff
@@ -0,0 +1,63 @@
...
+    else
+        JAVA=`which java`
+        javaSearchType=PATH
+        javaSearchTarget=$PATH
+    fi
+
+    if [ ! -x "${JAVA}" ]; then
+        echo
+        echo "The java command \"${JAVA}\" is not executable."
+        echo "It was configured as ${javaSearchType}=${}javaSearchTarget}"
```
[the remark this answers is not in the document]
dmatej: Good catch, thanks!

OndroMih (approved):

PR #25448, opened 2025-04-10
Fixed url to file conversion breaking paths with spaces

dmatej:
This PR fixes part of issues seen when manually testing another branch on Windows using path with spaces.
Fixes also a validation issue when URL with empty path could pass into the classloader.
I have cherrypicked also another commit refactoring test which I needed here to confirm issues we have noticed in sources.

### DIFF
--- a/nucleus/common/common-util/src/main/java/com/sun/enterprise/util/io/FileUtils.java
+++ b/nucleus/common/common-util/src/main/java/com/sun/enterprise/util/io/FileUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2024 Contributors to the Eclipse Foundation.
+ * Copyright (c) 2022, 2025 Contributors to the Eclipse Foundation.
  * Copyright (c) 2006, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -33,6 +33,7 @@ import java.io.PrintWriter;
 import java.io.Writer;
 import java.lang.System.Logger;
 import java.lang.System.Logger.Level;
+import java.net.URL;
 import java.nio.ByteBuffer;
 import java.nio.channels.Channels;
 import java.nio.channels.FileChannel;
@@ -1031,6 +1032,24 @@ public final class FileUtils {
         }
         return matches;
     }
+
+
+    /**
+     * Convert a URL to an absolute file if it is possible..
+     *
+     * @param resourceUrl
+     * @return absolute path to the file or directory pointed by the URL
+     * @throws IllegalArgumentException if the URL cannot be converted to a File
+     *
+     */
+    public static File toFile(URL resourceUrl) throws IllegalArgumentException {
+        try {
+            return new File(resourceUrl.toURI()).getAbsoluteFile();
+        } catch (Exception e) {
+            throw new IllegalArgumentException("Cannot convert URL to File: " + resourceUrl, e);
+        }
+    }
+
 
     /**
      * Represents a unit of work that should be retried, if needed, until it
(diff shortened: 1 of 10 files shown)


nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/CommonClassLoaderServiceImpl.java
```diff
@@ -172,12 +172,7 @@ public String getCommonClassPath() {
     }
 
     private static String urlsToClassPath(Stream<URL> urls) {
-        return urls
-                .map(URL::getFile)
-                .filter(Predicate.not(String::isBlank))
-                .map(File::new)
-                .map(File::getAbsolutePath)
-                .collect(Collectors.joining(File.pathSeparator));
+        return urls.map(FileUtils::toFile).map(File::getAbsolutePath).collect(Collectors.joining(File.pathSeparator));
```
pzygielo, line 175: I'd keep original line breaks to show that only part of the method changed, not the whole one.

dmatej: It is under 120 characters border, so it is not so hard to read. I would agree otherwise.

nucleus/common/common-util/src/main/java/com/sun/enterprise/util/io/FileUtils.java
```diff
@@ -1032,6 +1033,37 @@ public static File[] findFilesInDir(File dir, final String regexp) {
...
+
+    /**
+     * Convert a URL to an absolute path if it is possible..
+     *
+     * @param resourceUrl
+     * @return absolute path to the file or directory pointed by the URL
+     * @throws IllegalArgumentException if the URL cannot be converted to a File
+     *
+     */
+    public static String toAbsolutePath(URL resourceUrl) throws IllegalArgumentException {
```
OndroMih, line 1045: This method is not used. And that's correct - I would always prefer to explicitly call `toFile(URL).getPath()` for clarify.
Can we delete this method?

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/CommonClassLoaderServiceImpl.java
OndroMih, line 175: Where did `.filter(Predicate.not(String::isBlank))` disappear? Why was it even there? Is it OK to remove it or it's possible that some urls are blank?
Maybe it's worth adding `.filter(File::exists)` ?

OndroMih: Also note that `new File("").getAbsolutePath()` gives the value of the home directory. So I suggest adding also `.filter( f -> !f.getPath().isBlank())`

dmatej: Hmm, I see it is used by the addLibraryCommand, so you are right and that resolves even Piotrek's suggestion, because it will be longer line.  I still have doubts it is possible to get an empty string here, maybe would be useful to cover that by a test.

nucleus/common/common-util/src/main/java/com/sun/enterprise/util/io/FileUtils.java
dmatej: It is used in 7.1 branch, this is backport. But ok, I will rebase that later and remove it there too.

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/CommonClassLoaderServiceImpl.java
OndroMih: `addLibraryCommand` verifies that the file exists and only then it adds the URL. So it seems that urls cannot be empty in any case and my comment is void.

dmatej: Funny, I extended tests and found much more old and dangerous bugs ... will be fixed now.

OndroMih (approved):

PR #25468, opened 2025-04-19
Improved Derby and CommonClassLoaderService initialization

dmatej:
- The Install root is resolved from the `StartupContext`, however it is not mandatory.
- `AppServerStartupTest` intentionally does not depend on `KernelJUnitExtension`, I did not notice that when I made previous changes
- `KernelJUnitExtension` should not set any System properties
- `HK2JUnit5Extension` uses own `ServerEnvironmentImpl` extension which provides the preconfigured `StartupContext` and not null.
- Improved `DriverLoader` and its landscape
- Introduced `org.glassfish.connectors.dbVendorMappingRoot` system property
- `DriverLoader` is a singleton now
- Improved resource management
- Reduced copy and paste

### DIFF
--- a/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/connectors/util/DriverLoader.java
+++ b/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/connectors/util/DriverLoader.java
@@ -19,17 +19,19 @@ package com.sun.enterprise.connectors.util;
 
 import com.sun.appserv.connectors.internal.api.ConnectorConstants;
 import com.sun.enterprise.connectors.ConnectorRuntime;
-import com.sun.enterprise.util.SystemPropertyConstants;
 import com.sun.logging.LogDomains;
+
+import jakarta.annotation.PostConstruct;
+import jakarta.inject.Singleton;
 
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileInputStream;
-import java.io.FilenameFilter;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.lang.reflect.Modifier;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.HashSet;
@@ -40,15 +42,17 @@ import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
 import java.util.jar.Attributes;
+import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
 import java.util.jar.Manifest;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import java.util.zip.ZipEntry;
 
 import org.jvnet.hk2.annotations.Service;
 
 import static com.sun.enterprise.glassfish.bootstrap.cfg.BootstrapKeys.DERBY_ROOT_PROP_NAME;
+import static com.sun.enterprise.glassfish.bootstrap.cfg.BootstrapKeys.INSTALL_ROOT_PROP_NAME;
+import static com.sun.enterprise.glassfish.bootstrap.cfg.BootstrapKeys.INSTANCE_ROOT_PROP_NAME;
 
 /**
  * Driver Loader to load the jdbc drivers and get driver/datasource classnames
@@ -57,12 +61,12 @@ import static com.sun.enterprise.glassfish.bootstrap.cfg.BootstrapKeys.DERBY_ROO
  * @author Shalini M
  */
 @Service
+@Singleton
 public class DriverLoader implements ConnectorConstants {
 
-    private static Logger logger =
-    LogDomains.getLogger(DriverLoader.class, LogDomains.RSR_LOGGER);
+    private static Logger logger = LogDomains.getLogger(DriverLoader.class, LogDomains.RSR_LOGGER);
 
-    private static final String DRIVER_INTERFACE_NAME="java.sql.Driver";
+    private static final String DRIVER_INTERFACE_NAME = "java.sql.Driver";
     private static final String SERVICES_DRIVER_IMPL_NAME = "META-INF/services/java.sql.Driver";
     private static final String DATABASE_VENDOR_DERBY = "DERBY";
     private static final String DATABASE_VENDOR_JAVADB = "JAVADB";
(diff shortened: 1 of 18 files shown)


nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/CommonClassLoaderServiceImpl.java
```diff
@@ -189,87 +178,57 @@ private static String urlsToClassPath(Stream<URL> urls) {
...
-            arguments = startupContext.getArguments();
+        final File derbyLib = derbyHome.resolve("lib").toFile();
+        if (!derbyLib.exists()) {
+            LOG.info(KernelLoggerInfo.cantFindDerby);
+            return Collections.emptyList();
         }
 
-        String derbyHome = null;
-
-        if (arguments != null) {
```
OndroMih, line 203: Is it OK now to ignore any `DERBY_HOME_PROP` property in `StartupContext`? Is it not used anywhere?

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/CommonClassLoaderServiceImpl.java
```diff
@@ -189,87 +178,57 @@ private static String urlsToClassPath(Stream<URL> urls) {
...
-
-        return Arrays.asList(derbyLib.listFiles(new FilenameFilter(){
-            @Override
-            public boolean accept(File dir, String name) {
-                // Include only files having .jar extn and exclude all localisation jars, because they are
-                // already mentioned in the Class-Path header of the main jars
-                return (name.endsWith(".jar") && !name.startsWith("derbyLocale_"));
-            }
-        }));
+        return new File(installDir, "config").toPath().resolve(derbyHome);
```
OndroMih, line 204: Why does this use "installDir/config" as the root path? Is it an arbitrary path introduced here if the property specifies a relative path? Should we also use it here: https://github.com/eclipse-ee4j/glassfish/blob/7bb61e595e5f43b02406a2264a7d351f2389db0e/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/connectors/util/DriverLoader.java#L495 ?

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/CommonClassLoaderServiceImpl.java
```diff
@@ -189,87 +178,57 @@ private static String urlsToClassPath(Stream<URL> urls) {
...
         }
 
-        String derbyHome = null;
-
-        if (arguments != null) {
-            derbyHome = arguments.getProperty(DERBY_HOME_PROP,
-                System.getProperty(DERBY_HOME_PROP));
-        }
+        return Arrays
+            .asList(derbyLib.listFiles((dir, name) -> name.endsWith(".jar") && !name.startsWith("derbyLocale_")));
```
OndroMih, line 192: Is it worth to introduce a constant for `.jar`? It's also used in `JarFileFilter` class

dmatej: Not sure, because we are messing with classloaders. Any new/removed dependency can change order of initialization which can break something.

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/CommonClassLoaderServiceImpl.java
dmatej: Seems it is set just in `ACCStartupContext`. So you are probably right, I should use better strategy for these changes and always check following changes in my branch which probably fixed that. On the other hand no test failed, GF is quite overengineered here. Or underengineered with all those dependencies crossing paths ...

appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/connectors/util/DriverLoader.java
```diff
@@ -492,7 +494,7 @@ private boolean isVendorSpecific(File f, String dbVendor, String className,
 
     private List<File> getJdbcDriverLocations() {
     List<File> jarFileLocations = new ArrayList<>();
-        jarFileLocations.add(getLocation(SystemPropertyConstants.DERBY_ROOT_PROPERTY));
+        jarFileLocations.add(getLocation(DERBY_ROOT_PROP_NAME));
         jarFileLocations.add(getLocation(SystemPropertyConstants.INSTALL_ROOT_PROPERTY));
```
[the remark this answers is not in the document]
dmatej: Outside of my intentions in this PR. These changes cause more issues then one would expect. But I have some changes already in my stash, the problem is that this is a basic commit, so every change I do here will cause conflicts in rebase.

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/CommonClassLoaderServiceImpl.java
```diff
@@ -79,78 +84,62 @@ public class CommonClassLoaderServiceImpl implements PostConstruct {
...
-            if (installLibPath.isDirectory()) {
-                Collections.addAll(cpElements,
-                    installLibPath.listFiles(new CompiletimeJarFileFilter()));
-            }
-        } else {
-            logger.logp(Level.WARNING, "CommonClassLoaderServiceImpl",
-                "createCommonClassLoader",
-                KernelLoggerInfo.systemPropertyNull,
-                SystemPropertyConstants.INSTALL_ROOT_PROPERTY);
+        final String installRoot = System.getProperty(INSTALL_ROOT_PROP_NAME);
```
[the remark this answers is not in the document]
dmatej: Actually it is a good idea, I will try that.

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/CommonClassLoaderServiceImpl.java
dmatej: Aha, yes, I added that because when it is a relative path, it is always resolved to config dir. However all these relative paths we use now are bit insecure, so this is rather temporary. And maybe not needed any more, after recent changes, I have to check.

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/CommonClassLoaderServiceImpl.java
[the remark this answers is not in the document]
dmatej: StartupContextUtil is deleted in 710, all these things around are much less complicated.

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/CommonClassLoaderServiceImpl.java
dmatej: I started understanding why this ... all those values have single source, however BootstrapKeys are not available on classpath everywhere. 
But still, convention of the StartupContext is to use "JVM options like" keys, not ENV like keys.

dmatej:
The SSHClusterITest failure - I need to rebase to the last commit of #25470 fixing that.

arjantijms (approved):

PR #25469, opened 2025-04-20
Fix more tests for JDK 24

arjantijms:


hs536 (approved):

pzygielo:
If there only was a clue what is the nature of the fix and why so many weblogic files are needed in `src/main`s...

arjantijms:
Good point; JDK 24 defaults to not loading dtd and XSD over the network, so gf resorts to loading them from /lib

All tests in deployments that tested wls compatibility now failed because of that. After computing the wls schemas they passed again.

arjantijms:
Some of the other changes are to make the deployment tests easier to debug; specifically printing out which test is being run, and allowing to set a single test to be run.

PR #25470, opened 2025-04-22
Fixed race condition in LoggingOutputStream

### LINKED ISSUE #25438: Flaky LoggingPrintStreamTest.testStreamMethods test
Fails for example here: https://github.com/eclipse-ee4j/glassfish/actions/runs/14180082343/job/39723962973#step:5:3838

dmatej:
* Closes #25438 
* Test is now reproducible (repeated 100 times, failed 3-40% locally, depending on CPU profile)
* Few more small improvements

### DIFF
--- a/nucleus/glassfish-jul-extension/src/main/java/org/glassfish/main/jul/handler/LogCollectorHandler.java
+++ b/nucleus/glassfish-jul-extension/src/main/java/org/glassfish/main/jul/handler/LogCollectorHandler.java
@@ -105,12 +105,8 @@ public class LogCollectorHandler extends Handler {
      */
     public List<GlassFishLogRecord> getAll() {
         final List<GlassFishLogRecord> list = new ArrayList<>(this.buffer.getSize());
-        while (!this.buffer.isEmpty()) {
-            final GlassFishLogRecord record = this.buffer.poll();
-            if (record == null) {
-                // Unreachable code, but just in case.
-                throw new IllegalStateException("Buffer is empty but should contain records.");
-            }
+        GlassFishLogRecord record;
+        while ((record = this.buffer.poll()) != null) {
             list.add(record);
         }
         return list;
@@ -133,8 +129,9 @@ public class LogCollectorHandler extends Handler {
      * Drops all collected records.
      */
     public void reset() {
-        while (!this.buffer.isEmpty()) {
-            this.buffer.poll();
-        }
+        GlassFishLogRecord record;
+        do {
+            record = this.buffer.poll();
+        } while (record != null);
     }
 }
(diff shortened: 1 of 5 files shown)


hs536 (approved):

nucleus/glassfish-jul-extension/src/main/java/org/glassfish/main/jul/handler/LogCollectorHandler.java
```diff
@@ -106,7 +106,12 @@ public GlassFishLogRecord pop() {
     public List<GlassFishLogRecord> getAll() {
         final List<GlassFishLogRecord> list = new ArrayList<>(this.buffer.getSize());
         while (!this.buffer.isEmpty()) {
-            list.add(this.buffer.poll());
+            final GlassFishLogRecord record = this.buffer.poll();
+            if (record == null) {
+                // Unreachable code, but just in case.
```
pzygielo, line 111: To help me understand. This can be reached if after checking `!this.buffer.isEmpty()` here, but before `this.buffer.poll()` some other party empties `buffer`.
`this.buffer` is a field in this class, that could be MT-access protected, but it's decided to not do that.
And if this pretty normal situation happens, suddenly `ISEx` is thrown.
:eyes:

nucleus/glassfish-jul-extension/src/main/java/org/glassfish/main/jul/handler/LoggingOutputStream.java
```diff
@@ -165,14 +165,25 @@ public void run() {
...
             pumpClosed = true;
-            this.interrupt();
+            while (getState() == State.RUNNABLE) {
+                Thread.onSpinWait();
+            }
+            if (getState() == State.WAITING) {
+                this.interrupt();
+            }
+            try {
+                this.join();
```
chabala, line 183: What's going on with all the unnecessary `this` in this class?

dmatej: Sometimes it helps tired devs to think easily ;-)

And in one scenario it is very useful - when refactoring fields it prevents mistake collisions (usually reported by checkstyle as warnings, also by Eclipse IDE, we have lot of such cases in GF sources).

But probably just like you, I would not usually use `this` for methods. Seems, when I was writing this code I was thinking about which thread is waiting for the other. The calling thread is waiting for the end of the called, owning the method.

The more important on this code is previous wrong assumption that when I interrupt a thread, it is immediately interrupted and I don't need to wait for its end. I do wonder how many times I did the same silly mistake.

Currently we don't have any checkstyle rule forbidding unnecessary `this`.

nucleus/glassfish-jul-extension/src/main/java/org/glassfish/main/jul/handler/LogCollectorHandler.java
dmatej: This class is used usually just in tests to catch logs and process them at some safe point, so null means that we do the check too early because logs are still coming.
Original code would just pass take what was already there, this code should "panic" and report at least some flaky test.

PR #24898, opened 2024-04-06
Command recorder Admin Console plugin to configure admin command logging (with plain facelets, no JSF templating)

OndroMih:
This PR does 2 things:

* Adds support for using plain facelets with `*.xhtml` files
* Adds a button in the Admin Console header to enable/disable admin command logging (fixes #25039)

This allows creating new pages in Admin Console using the standard Jakarta Faces facelets and bundle them in a JAR in the Admin GUI WAR. Those pages can be then injected into old JSF templating pages via iframe or ajax call. it's also possible to provide an injection point to the existing Admin Console pages via standard means, all in the same JAR file. 

New Facelets should be placed into the root of the web app. Or, even better, in a separate JAR file inside the web app, into the `META-INF/resources/ directory. CDI beans can be places in the WAR or nested JAR as usual.

image

Video: 

https://github.com/user-attachments/assets/07e99389-c3f4-4ad1-ad79-6d3f6562f0ee

PR #25485, opened 2025-05-06
Expose services with local parent locator scope (e.g. @RunLevel) to the service locator in system apps

### LINKED ISSUE #25465: Broken View JMS Physical Destination Statistics

7.0.0-7.0.23+

### JDK version

11

### OS

GNU/Linux

### Database

_No response_

### Problem Description

Admin GUI -> server ->  JMS Physical Destination -> Statistics -> View does not work.

With 7.0.23 because:
Image
```
[2025-04-17T10:52:36.337596+02:00] [GF 7.0.23] [INFO] [] [org.glassfish.admingui] [tid: _ThreadID=62 _ThreadName=admin-listener(5)] [levelValue: 800] [[
  Exception Occurred :A MultiException has 2 exceptions.  They are:
1. java.lang.IllegalStateException: Could not find an active context for org.glassfish.hk2.runlevel.RunLevel
2. java.lang.IllegalStateException: While attempting to create a service for SystemDescriptor(
  implementation=com.sun.enterprise.connectors.jms.system.JmsProviderLifecycle
  contracts={com.sun.enterprise.connectors.jms.system.JmsProviderLifecycle}
  scope=org.glassfish.hk2.runlevel.RunLevel
  qualifiers={}
  descriptorType=CLASS
  descriptorVisibility=NORMAL
  metadata=runLevelValue={20},runLevelMode={0},Bundle-SymbolicName={org.glassfish.main.jms.core},Bundle-Version={7.0.23}
  rank=0
  loader=OsgiPopulatorPostProcessor.HK2Loader(OSGiModuleImpl:: Bundle = [org.glassfish.main.jms.core [108]], State = [READY],1496022235)
  proxiable=null
  proxyForSameScope=null
  analysisName=null
  id=1204
  locatorId=0
  identityHashCode=1443194318
  reified=true) in scope org.glassfish.hk2.runlevel.RunLevel an error occured while locating the context
]]
```

With 7.0.0 (up to 7.0.15) because:
Image

### Steps to reproduce

Open Admin GUI and navigate to: server -> JMS Physical Destination. Select **View** for any destination from **Statistics** column.

### Impact of Issue

I'd expect something like in 6.2.5

Image

### LINKED ISSUE #25467: Missing active context for `org.glassfish.hk2.runlevel.RunLevel` in Admin GUI App
### Description

https://github.com/eclipse-ee4j/glassfish/blob/c19cd4cedfb787415dc86e3121ed7edac6dddaa0/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebModuleListener.java#L279-L284

`defaultServices` seems to contain

```
org.glassfish.hk2.runlevel.internal.AsyncRunLevelContext
org.glassfish.hk2.runlevel.internal.RunLevelControllerImpl
org.glassfish.hk2.runlevel.RunLevelContext
```

which later are not present in `webAppServices` in l. 283.

- This happened after 94ae8aa32adefde8b69d4eac43d7e03b88403e28 (#24898).

OndroMih:
Fixes #25467.
Fixes #25465.

This PR changes the locator of system apps so that it falls back to the parent locator and retrieves services from it directly and not only indirectly via the child locator. The new combining locator searches for services first in the child locator as before, but then delegates to the parent locator directly. This allows retrieving services local to the parent locator, as it was happening before when the parent locator was used as the locator for web apps.

With the change in https://github.com/eclipse-ee4j/glassfish/pull/24898 we introduced a child HK2 locator in webapps, delegating to the previous locator as the parent locator. This was later changed to do this only for system apps (needed for Admin Console) because there were some issues with combining child and parent locator with Faces in Embedded GlassFish.

After that change, the locator doesn't have access to services with scope local to the parent locator, e.g. with the `@RunLevel` scope. Some of these services are accessed directly by Admin Console, e.g. `JmsProviderLifecycle` as described in #25465. Which now results in an exception because the service doesn't have a context in the child locator.

This fix avoids that by asking the parent locator directly if a services is not found or doesn't have a proper context in the child locator.

### DIFF
--- a/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebModuleListener.java
+++ b/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebModuleListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2024 Contributors to the Eclipse Foundation.
+ * Copyright (c) 2022, 2025 Contributors to the Eclipse Foundation.
  * Copyright (c) 1997, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -62,6 +62,7 @@ import org.glassfish.web.LogFacade;
 import org.glassfish.web.deployment.runtime.SunWebAppImpl;
 import org.glassfish.web.deployment.runtime.WebProperty;
 import org.glassfish.web.deployment.util.WebValidatorWithCL;
+import org.glassfish.web.hk2.CombiningServiceLocator;
 
 import static java.util.logging.Level.FINE;
 import static java.util.logging.Level.FINER;
@@ -276,12 +277,15 @@ final class WebModuleListener implements LifecycleListener {
 
     }
 
+    ServiceLocator webAppServices;
+
     private ServiceLocator createWebAppServices(WebModule webModule, ServiceLocator defaultServices) {
         ServiceLocator webAppServices = ServiceLocatorFactory.getInstance().create(webModule.getComponentId(), defaultServices);
         initializeServicesFromClassLoader(webAppServices, Thread.currentThread().getContextClassLoader());
 
-        return webAppServices;
+        return new CombiningServiceLocator(webAppServices, defaultServices);
     }
+
 
     private static void initializeServicesFromClassLoader(ServiceLocator serviceLocator, ClassLoader classLoader) {
         DynamicConfigurationService dcs =
@@ -446,4 +450,5 @@ final class WebModuleListener implements LifecycleListener {
             wrapper.addInitParameter("fileEncoding", fileEncoding);
         }
     }
+
 }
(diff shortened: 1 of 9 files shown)


appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebModuleListener.java
```diff
@@ -276,13 +277,16 @@ private void configureJsp(WebModule webModule) {
 
     }
 
+    ServiceLocator webAppServices;
```
pzygielo, line 280: What is this field used for?

OndroMih: A left-over :)

appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebModuleListener.java
```diff
@@ -276,13 +277,16 @@ private void configureJsp(WebModule webModule) {
...
+
```
OndroMih, line 281:
```suggestion
```

dmatej (approved):

PR #25486, opened 2025-05-07
Remove unused parameter

pzygielo:
- testing https://github.com/eclipse-ee4j/glassfish/pull/25484#discussion_r2078062237


dmatej (approved):
Eh, I forgot to add the parameter to the set. But perhaps you are right, if we don't need it now, it was just my overengineering and we can simply remove it.

PR #25490, opened 2025-05-08
Classloaders parallelCapable and named

dmatej:
- all GlassFish classloaders are parallelCapable and have names
- OSGi bootstrap classloading - classloader is a parameter

Testing
- all available tests (GF+TCK) passed locally.

### DIFF
--- a/appserver/appclient/client/acc-standalone/src/main/java/org/glassfish/appclient/client/acc/agent/ACCAgentClassLoader.java
+++ b/appserver/appclient/client/acc-standalone/src/main/java/org/glassfish/appclient/client/acc/agent/ACCAgentClassLoader.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022, 2025 Contributors to the Eclipse Foundation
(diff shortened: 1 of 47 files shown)


pzygielo:
image

- I believe - with https://github.com/eclipse-ee4j/glassfish/pull/25474 it will not be 7.0.25 from master

dmatej:
> * I believe - with [Next version is 7.1.0 #25474](https://github.com/eclipse-ee4j/glassfish/pull/25474) it will not be 7.0.25 from master

It still can be, what I am doing now is that I am slowly separating parts of the gjule-vs-gc branch which is the original reason for 7.1.0, which are still safe for 7.0.x. 

The version 7.0.25 then can still be released, but maybe we will not do that.

It is a safe way of the evolution - I am also running TCK tests (all tests we have) locally, so when we really merge more radical changes, we still have some overlap in testing.

And also - reviewing the final state of the "radical" part should be easier.

So, simply - I am marking this to separate PRs which could go to 7.0.25, and those which must not (they would not pass current TCK10 as TCK10 doesn't support JPMS; because of the JPMS changes I have found also some issues in the TCK like hard coded dependency on some GF classes; TCK11 changes lot of things, and when I later switch to help with GF8 and TCK11, I will be ready for that at least - if guys already did not resolve that anyway).

arjantijms (approved):

PR #25498, opened 2025-05-15
Improved stopping local servers

dmatej:
- admin-tests were sometimes failing, because the oldPid was null
- It is better to rely on the pid file than on server responding, however on the future 710 branch I noticed few "grey zone effects" as it is yet bit faster than 7.0.x and stop command was executed too early:
  - pid file can be created, however the content is not flushed yet, so it is empty.
  - pid file may exist and have the pid number, however server is not listening yet.
  - pid file may exist and have the pid number, server is listening, however it is not ready
  - to serve requests (I am not sure if I ever noticed this, however it might be something to consider; we check the pid to know what to do next - use/kill/stop?).
- when stopping the sever, it can be similar:
  - kill -9 will not remove the pid file. However we can use Process to check if such pid is valid.
  - if the process is listening on port, but all files are gone, we cannot know if it is "our" process and it is not our problem. User did something wrong, user have to resolve it, usually kill the process manually. Other scenarios should not happen and if they would happen, they would be probably considered as bugs (if I did not forget something now).
- Then: the pid file should be created as soon as possible and deleted right before the process dies.

### DIFF
--- a/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/LocalServerCommand.java
+++ b/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/LocalServerCommand.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2024 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022, 2025 Contributors to the Eclipse Foundation
  * Copyright (c) 1997, 2018 Oracle and/or its affiliates. All rights reserved.
  *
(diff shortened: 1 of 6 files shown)


dmatej:
> > however on the future 710 branch
> 
> What do you mean by that? I was under impression master is that branch, and I can't recognize anything else to be candidate for such in https://github.com/eclipse-ee4j/glassfish/branches/all.

I meant this PR, sorry for the confusion, I should use more consistent identification. https://github.com/eclipse-ee4j/glassfish/pull/25183

arjantijms (approved):

PR #25510, opened 2025-05-21
Handle covariant return types in TypeUtil.sameMethodSignature

### LINKED ISSUE #25500: TypeUtil.sameMethodSignature does not handle covariant return types

8.0.11

### JDK version

any

### OS

any

### Database

_No response_

### Problem Description

The platform TCK has a test that is testing whether an ejb with a method with a mandatory transaction annotation is correctly processed even if the bean method which overrides the remote method interface is covariant in the return type. A simpler example is this: 

```java
@Stateless
public class FuzzyBean extends FuzzyBeanBase implements FuzzyRemoteIF, FuzzyLocalIF {

    public FuzzyBean() {
    }

    @Override
    @TransactionAttribute(MANDATORY)
    public Double[] getNumbers() {
        return (Double[]) super.getNumbers();
    }

    @Override
    @TransactionAttribute(MANDATORY)
    public String[] getMessages() {
        return super.getMessages();
    }

    @Override
    @TransactionAttribute(MANDATORY)
    public String getMessage() {
        return super.getMessage();
    }

}
```

```java
import jakarta.ejb.Remote;

@Remote
public interface FuzzyRemoteIF {
    Object getMessage();

    Object[] getMessages();

    Object[] getNumbers();
}
```

The web profile `com.sun.ts.tests.ejb30.tx.session.stateless.cm.covariant.ClientTest` test fails because this type of override is not matched by the `com.sun.enterprise.deployment.util.TypeUtil#sameMethodSignature(Method,Method)` method used by the `org.glassfish.ejb.deployment.annotation.handlers.TransactionAttributeHandler` in this code block:

```java
            Method annMethod = (Method) ainfo.getAnnotatedElement();
            Set<MethodDescriptor> txBusMethods = ejbDesc.getTxBusinessMethodDescriptors();
            for (MethodDescriptor md : txBusMethods) {
                Method method = md.getMethod(ejbDesc);
                if (TypeUtil.sameMethodSignature(method, annMethod) && ejbDesc.getContainerTransactionFor(md) == null) { // <---- Fails here
                    // override by xml
                    ejbDesc.setContainerTransactionFor(md, containerTransaction);
                }
            }
```

There is a simple fix to get this working:

```diff
Index: appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/util/TypeUtil.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/util/TypeUtil.java b/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/util/TypeUtil.java
--- a/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/util/TypeUtil.java	(revision 43fb2bf706cc036467ee6663590d47673f2e2ce6)
+++ b/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/util/TypeUtil.java	(date 1747463062403)
@@ -428,6 +428,10 @@
             return true;
         }
 
+        if(m1.getReturnType().isAssignableFrom(m2.getReturnType())) {
+            return true;
+        }
+
         Type grt1 = m1.getGenericReturnType();
         Type grt2 = m2.getGenericReturnType();
 
```

With that change I see the tests passing.

### Steps to reproduce

Run the following test in the platform-tck:

```
starksm@Scotts-Mac-Studio enterprise-beans-tck-run % mvn -Pstaging,ejb30 -Dtest=com.sun.ts.tests.ejb30.tx.session.stateless.cm.covariant.ClientTest -DdependenciesToScan=jakarta.tck:ejb30 test
```

### Impact of Issue

This will block the platform tck release.

arjantijms:
Fixes #25500

### DIFF
--- a/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/util/TypeUtil.java
+++ b/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/util/TypeUtil.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022, 2025 Contributors to the Eclipse Foundation
  * Copyright (c) 1997, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -425,6 +425,10 @@ public class TypeUtil {
      */
     private static boolean sameReturnTypes(Method m1, Method m2) {
         if (m1.getReturnType().equals(m2.getReturnType())) {
+            return true;
+        }
+
+        if (m1.getReturnType().isAssignableFrom(m2.getReturnType())) {
             return true;
         }
 


arjantijms:
I wonder. Presumably GlassFish 7 has always passed the EE 10 TCK for every release, and the reason for this change is a failing test. That test didn't change (for what it tests) between EE 10 and EE 11.

arjantijms:
> Then why it did not fail before? 

That's essentially the number one questions for hundreds of failing tests over the last year. To really find the answer we should investigate that exact same test (in its EE 10 version) on GlassFish 7, and see what is essentially different.

OndroMih:
Maybe Java 21 is the difference? Or some bug in it? GF 7 is tested on Java 7, GF 8 is tested on Java 21.
If that's the case, it makes sense to backport this. But it would be better to create a test for GF 7 based on the test that was failing in EE 11 TCK and reproduce it, and only fix it if it's reproduced.

dmatej:
Btw the method name is wrong now, because return types now can be different. Was wrong even before because of the TypeVariable usage ... we should rename it.

PR #25517, opened 2025-05-23
Migrating keystores from JKS to PKCS12, fixed TCK tests

### LINKED ISSUE #25529: Migrate from JKS to PKCS12
### Description

* See https://openjdk.org/jeps/229
* Problems: 
  * Keystore usages and access are spread all over project
  * Some places don't support keystore types and explicitly use JKS
  * Provider selection usually respects Java defaults
  * PKCS12 format has different behavior than JKS, tested on JDK11 (Temurin)
    * PKCS12: when changing keystore password, keytool automatically changes passwords also to all keys using the same password
    * JKS: You have to do that later

dmatej:
### PKCS12
* Fixes #25529 , we are using PKCS12 now.
* Some places like AMX had own management of keystores not reflecting changes over years.

### Default security files removed
* Removes "default" files in security.jar
  * Were sometimes used when files were missing like in embedded. cacerts was just an empty keystore. Other files seem unused as tests are passing.
  * I did not like that runtime is parasiting on a set of files without any maintenance, which were sometimes replaced, sometimes not.

### Default timeouts removed
* asadmin start-domain/stop-domain doesn't have default timeout any more. When using autodeploy directory to deploy hundreds war files, we are closing to one minute. 
* It is not easy to guess the good timeout, we always had problems with it, and 7.1 now has the --timeout attribute for that, added in #25580 .
* By default we don't expect that the startup should time out. It should just do what is needed or crash if it cannot.

### Tests
* Updated and synchronized TCK
* Adapted to usage of p12 files as many have the JKS hardcoded.
* All TCK tests now use the same version of Arquillian Containers, 2.0.0 from OmniFish (originally from OmniFaces).
* Changes to Platform TCK were already merged to master branch, however not released yet.
* Changes to Standalone TCKs will be distributed later to their project repositories and jenkins builds, this PR served as a prototype of these changes. After releases of TCKs we can just delete the code from poms (mostly filtering xml and property files).


dmatej:
Ok, I followed the idea, it broke some pieces, after their resolving embedded should have bit faster starts, however it still broke few tests as now embedded doesn't contain certificates by default. Then tests using HTTPS cannot pass, however they lead to the need to provide an API to enable secured endpoints.

dmatej:
> > ```
> > * Changes DN of certificates (Ottawa -> Brussels, Canada -> Belgium)
> > ```
> 
> Q1: Why?

Eclipse Foundation moved to EU few years ago, all active GF developers are in Europe or Japan. However the second question is more interesting.

> Q2: Would the certificate work as good without geo attributes in DN at all?

It seems it would, those attributes are optional. Basically we need just CN which should be a hostname, which is much more complicated question ... so for the "example named domain1" certificate we use "localhost".

I am still working on it, it will not go to 7.0.x, maybe some commits could, but no radical changes.

hs536 (dismissed):

hs536 (commented):
My apologies, the above approval was a misoperation.

dmatej:
> My apologies, the above approval was a misoperation.

No problem, I know it is not ready to be reviewed+merged, I will probably split it later to smaller PRs.

dmatej:
Status: Time to stop now.
* domain1 contains keystores generated in build. User should create new domain for production usage.
* embedded doesn't contain anything, BUT it is possible to set own keystore+truststore
* default keystore type is PKCS12, not JKS
* The keystore type can be detected from the file name (tested P12, PFX, JKS, JCEKS - last two are legacy formats, so we avoid them, but it is possible).
* new class KeyTool wraps keytool program; most tasks can be done in Java already; however P12 and JKS formats have bit different behavior when changing passowrds. See #25529 for more. KeyTool was ideal to simplify that, so we don't have to care now.
* Reduced copy and paste - using constants instead of hardcoded strings and copy pasted constants. However SystemPropertyConstants class should die, now it resolves time consuming "search and replace" problem.
* master-password, domains-passwords - use .p12 file extension now (used JCEKS, in Java5 was the only impl supporting keys and their CRUD).

TODO: 
* Not sure about the TCK, I have to try. 
* Now it is 47 commits - I will split at least part to own PRs.

dmatej:
Status: seems all tests including TCKs on Eclipse and my infrastructure are passing.

dmatej:
Status: passed TCKs using integration in the project using my hardware; passed also updated platform TCK on [Eclipse Jenkins](https://ci.eclipse.org/jakartaee-tck/job/jakartaee-tck/job/PR-2374/), however standalone TCKs were not touched yet (our integration does it).

dmatej:
And yet once more to draft again - when I was rebasing this I forgot two commits with the TCK changes, meanwhile I was on the long journey with race conditions and arquillian - which I have to release before merging this.

dmatej:
Status: Done, I am running TCKs on private Jenkins for the last time. Two days ago few blocks failed: CDI (forgot to rename class), EJB (Jenkins node network issue) and Faces (removed one  of Arquillian GlassFish jars which was required so I had to revert it back), now all of them passed alone, so I am verifying this PR again as a whole. 
The ARQ GF project 2.0.0 was released today to Maven Central.

nucleus/admin/cli/src/main/java/com/sun/enterprise/admin/cli/CLIConstants.java
```diff
@@ -35,7 +35,7 @@ public final class CLIConstants {
     public static final String DEFAULT_HOSTNAME = "localhost";
     public static final String EOL = System.lineSeparator();
 
-    public static final Duration WAIT_FOR_DAS_TIME_MS = getEnv(TIMEOUT_START_SERVER, 60L);
```
hs536, line 38: isn't there a risk of the start commands waiting indefinitely?

dmatej: If it would wait indefinitely, what that means?
* Server crashed - then the command should react somehow, probably escalate the error, but it should not wait forever.
* Server is still starting - then it should just wait. It doesn't have any control over the process, it could kill it, but that could damage the domain. 
* I have seen many times asadmin timing out and then the server started in the background; when we used scripts ie. in TCK, it created zombie instances living forever in the background (until the virtual machine was killed).
* Now you still can use timeout as before, it is just not the default one minute, but you have to set it explicitly using the `--timeout` argument or by environment option (which was possible before too - and we used that in the TCK if I remember well).
* You can always sent SIGTERM signal (CTRL+C on Windows does something similar).

dmatej: I have created yet another issue (#25615) yesterday, because I noticed something to check yet once again, however the more I was thinking about it, it seems as a correct behavior (I noticed it waited nearly a minute when I assumed that it waits just for the PID - it seems it waited for ports open; I have to check that in the new issue). On GF7 we spent quite a lot of time fixing all possible scenarios of start/stop/restart failures.

avpinchuk (approved):
LGTM

nucleus/admin/cli/src/main/java/com/sun/enterprise/admin/cli/CLIConstants.java
hs536: GlassFish occasionally hangs during startup due to unresolved bugs. Previously, when this occurred, GlassFish would cease its operations and inform the user of an error (Yes, there was indeed a problem with lingering zombie processes). Although there's an option to set a timeout as before, this change could present an incompatibility, particularly for users who restart their servers via nightly batch processes.

dmatej: GF must not hang, that is what I will target in following days and weeks, it must not also fail without a message, what it does now ie when some OSGi dependency increases major version and it must be loaded before logging starts.

I slowly collected these issues to resolve in 7.1; also this PR made it easier to reproduce, race conditions, unnecessary slowdowns, so I hope we will get all those problems resolved before we release production ready 7.1.0.
https://github.com/eclipse-ee4j/glassfish/milestone/62

PR #25518, opened 2025-05-23
Porting #25517 to 7.0

dmatej:
- #25517

TODO: TCK


dmatej:
It will not go to 7.0.25 as it could break some features if not perfect. To be perfect I need to do some refactoring and make clear all use cases, add few features (generate self signed certificate, print help to generate certificate request) in embedded.

PR #25520, opened 2025-05-27
Faster detection of implicit CDI archives during deployment

OndroMih:
Speed up detection of implicit CDI archives - creates an index of archives with an implicit CDI bean class for all classes in the deployment and caches it for later checks of implicit archives during deployment.

Also fixes an issue that sniffers were added multiple times to a collection of application sniffers.

I did a simple measurement - deploying a 19 MB WAR with 22 JAR files in the lib directory, which contain around 12,000 class files.

The best deployment time I got before the optimizations, was 1.98s:

Screenshot from 2025-05-27 02-45-25

With these optimizations, I repeated the same steps, with the best deployment time of 1.31s - around half a second boost:

image

### DIFF
--- a/appserver/web/gf-weld-connector/src/main/java/org/glassfish/weld/connector/WeldUtils.java
+++ b/appserver/web/gf-weld-connector/src/main/java/org/glassfish/weld/connector/WeldUtils.java
@@ -199,7 +199,7 @@ public class WeldUtils {
         Set<URI> implicitBeanArchivePaths = (Set<URI>)context.getTransientAppMetadata().get(METADATA_KEY);
 
         if (implicitBeanArchivePaths == null) {
-            implicitBeanArchivePaths = findImplicitBeanArchivePaths(implicitBeanArchivePaths, context);
+            implicitBeanArchivePaths = findImplicitBeanArchivePaths(context);
             context.addTransientAppMetaData(METADATA_KEY, implicitBeanArchivePaths);
         }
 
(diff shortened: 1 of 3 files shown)


appserver/web/gf-weld-connector/src/main/java/org/glassfish/weld/connector/WeldUtils.java
```diff
@@ -195,25 +195,49 @@ public static boolean hasCDIEnablingAnnotations(DeploymentContext context, URI p
...
+        for (URI path : paths) {
+            if (implicitBeanArchivePaths.contains(path)) {
+                return true;
+            }
+        }
+        return false;
+    }
 
+    private static Set<URI> findImplicitBeanArchivePaths(Set<URI> pathsWithImplicitCDIBeans, DeploymentContext context) {
+        pathsWithImplicitCDIBeans = new HashSet<>();
```
pzygielo, line 215: New, private method - maybe drop `pathsWithImplicitCDIBeans` parameter if the first thing is to overwrite it with empty set?

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/SnifferManagerImpl.java
```diff
@@ -150,6 +152,7 @@ private <T extends Sniffer> List<T> getApplicableSniffers(DeploymentContext cont
                 continue;
             }
 
+            addSniffer:
```
dmatej, line 155: Couldn't it be refactored to avoid labels and "goto"?
https://www.baeldung.com/java-labeled-break

appserver/web/gf-weld-connector/src/main/java/org/glassfish/weld/connector/WeldUtils.java
OndroMih: Sure, I’ll change it. It’s a leftover after automatic refactoring.

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/SnifferManagerImpl.java
OndroMih: Sure, but would require more time. I’ll see if I can find a quick way to refactor it.

arjantijms (approved):

PR #25522, opened 2025-05-28
Merge 8.0 into jdk17 branch

arjantijms:


pzygielo:
As my browser refuses to open _Files changed_ tab: 74b15a8963185cf1872c3e29d4efe12f1bb33091 - why not to stay with 6.5.1 as it was before a081876763f?

arjantijms:
Good point! But; the JDK17 is a bit of a throw-away branch anyway, and only to pass the TCK. Details don't matter much, only that it passes the TCK.

pzygielo:
> But the JDK17 is a bit of a throw-away branch anyway, and only to pass the TCK. Details don't matter much, only that it passes the TCK.

My bad, I keep forgetting that. :face_with_diagonal_mouth:

hs536 (approved):

Issue #25553, opened 2025-06-13
Runnable jars appserver-cli.jar, admin-cli.jar are unrunnable


pzygielo:
The manifests define `Main-Class`, but no `Class-Path`.

```
Version = Eclipse GlassFish 7.1.0-SNAPSHOT (commit: a6defafa687f716fcbb36198a01adb81613784c3)
```

_Originally posted by @pzygielo in https://github.com/eclipse-ee4j/glassfish/issues/25183#issuecomment-2970835351_

OndroMih:
It would be good to cover these runnable JARs by tests, to at least verify they are executable and do something. 

For example, execute `java -jar glassfish7/glassfish/modules/admin-cli.jar list-domains` and expect OK result.

dmatej:
It is partially intentional, the problem is that mixing classpaths in manifests doesn't play well with JPMS, especially when the classpaths overlap with other jar files on the classpath using different paths - how Java reacts changes between versions. Originally it took them as relative URIs to the jar file which contained them.

This old GlassFish design is a disaster, simply said. If we want to provide a client, it should be standalone application NOT parasiting on directories of GlassFish. The deeper you go, the worse it is - if you start analyzing how many dependencies these jar files need. 

Now they are enumerated in asenv.conf.

Where I would go is to remove most of them and find some much lighter solution; if we need some, they could shadowed into the runnable jar file and even distributed alone, usable in user applications.

dmatej:
They are runnable, you just need to provide the modulepath and classpath from the command line. However I see the point, basically nobody should run them at this state directly, so then I would say you are right.
I am still curious when at some point we would be able to be full-JPMS, if this can be made simple; or even better, if we would be able to provide a standalone client, really independent runnable jar. I am crawling in this direction, however there's still too many things in the path.

OndroMih:
When running runnable JARs with -jar, Java doesn’t allow to set classpath or modulepath. They must be defined on the manifest. So it’s correct to change the manifest to make them not runnable, if we don’t intend to specify classpath or modulepath in the manifest.

pzygielo:
I find this as not nice regression.

Although I understand no document promised the `admin-cli.jar` location to be stable. Or for `admin-cli.jar` to be runnable.

But:

1. [GF 3.1.1](https://repo.maven.apache.org/maven2/org/glassfish/distributions/glassfish/3.1.1/)
```
$ java -jar modules/admin-cli.jar version
[...]

PR #25554, opened 2025-06-14
Do not declare `Main-Class`es

pzygielo:
Replace
```
Error: Unable to initialize main class com.sun.enterprise.admin.cli.AdminMain
Caused by: java.lang.NoClassDefFoundError: org/glassfish/api/admin/CommandValidationException
```

with
```
no main manifest attribute, in admin-cli.jar
```


pzygielo:
- #25553

dmatej (approved):

PR #25558, opened 2025-06-16
DomainBuilder must not swallow failures

dmatej:
- If we are not able to generate certificates, report an exception, don't continue with domain creation.
- Failure generating certificates caused creating a new domain using the fallback and obsoleted certificate and especially publicly available private key.


OndroMih:
This is not a very nice way to fix the issue and block from creating a new domain config in case of a failure, without giving a hint how to fix it.

However, as this error shouldn't normally happen with any decent JDK, I think it's OK.

OndroMih (approved):

PR #25559, opened 2025-06-16
Excluded module-info.class files poisoning embedded packages

dmatej:
- Could result in random results
- The module-info was inherited from some dependency

### DIFF
--- a/appserver/extras/embedded/all/pom.xml
+++ b/appserver/extras/embedded/all/pom.xml
@@ -1723,6 +1723,7 @@
                             <includeScope>runtime</includeScope>
                             <includeTypes>jar,glassfish-jar</includeTypes>
                             <includes>**/*.class</includes>
+                            <excludes>module-info.class,META-INF/versions/*/module-info.class</excludes>
                         </configuration>
                     </execution>
                     <execution>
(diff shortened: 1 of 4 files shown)


OndroMih (approved):

appserver/extras/embedded/all/pom.xml
```diff
@@ -1723,6 +1723,7 @@
                             <includeScope>runtime</includeScope>
                             <includeTypes>jar,glassfish-jar</includeTypes>
                             <includes>**/*.class</includes>
+                            <excludes>module-info.class</excludes>
```
pzygielo: Would it exclude `META-INF/versions/9/module-info.class` as well?

----
In 7.0.25: 
- `org.glassfish.ha.api@3.1.13` in top-level
- `com.fasterxml.jackson.core@2.18.3` (for `--release 9+`)

pzygielo: Checked it locally. It would keep that one.

dmatej: Good catch!
I have to check what I have in the second branch and do it properly. Thanks!

dmatej: Same in the second branch; I have fixed that and force pushed the commit.

pzygielo (approved):

PR #25569, opened 2025-06-18
New SystemProperties logging global changes

dmatej:
- Also deleted unused GFSystem class
- In several cases I have noticed that some values are set repeatedly or even to different values which can cause some fragility (ordering, race conditions, surprises, etc).
- This solution is not global, affects just classes in this project, but even that is useful.
- Java doesn't offer anything like this, it is just possible to print system properties using jcmd or inside JVM.


appserver/admin/cli-optional/src/main/java/com/sun/enterprise/admin/cli/optional/DerbyControl.java
```diff
@@ -125,10 +126,10 @@ public DerbyControl(final String derbyCommand, final String derbyHost, final Str
 
         // Do not set derby.system.home if derbyHome is empty
         if (!isEmpty(derbyHome)) {
-            System.setProperty("derby.system.home", derbyHome);
+            setProperty("derby.system.home", derbyHome, true);
```
pzygielo, line 129: With all usages updated, could we have `setProperty(String, String)`, that would default the 3rd parameter to `true`?

dmatej: I wanted to force decision if it should or should not overwrite existing value - which is quite important. Correctly no such events should ever happen, but it is not always true.

arjantijms (approved):

PR #25574, opened 2025-06-20
Fixed race conditions in detached job management

### LINKED ISSUE #25555: Flaky DetachAttachITest.detachOnesAttachMulti
### Description

- https://github.com/eclipse-ee4j/glassfish/pull/25554 - https://github.com/eclipse-ee4j/glassfish/actions/runs/15654311966/job/44103308705?pr=25554#step:4:45652

```
Error:  Tests run: 3, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 31.76 s <<< FAILURE! -- in org.glassfish.main.admin.test.progress.DetachAttachITest
Error:  org.glassfish.main.admin.test.progress.DetachAttachITest.detachOnesAttachMulti -- Time elapsed: 10.99 s <<< FAILURE!
java.lang.AssertionError: 

Expected: asadmin succeeded
     but: was <Command attach failed.
remote failure: Job with id 10 does not exist.

>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
	at org.glassfish.main.admin.test.progress.DetachAttachITest.detachOnesAttachMulti(DetachAttachITest.java:116)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

- Microsoft Windows Server 2025
- Path: C:\hostedtoolcache\windows\Java_Temurin-Hotspot_jdk\21.0.7-6.0\x64
- 185f543762ff43411419884546a2bf9df103e796

dmatej:
* Fixes #25555 - there really was a race condition
  * Fixed partially by refactoring, see the commit
  * The problem happened when three attached clients waited for job to complete, then first processed the result and SOMETIMES before two other were able to do the same, it removed the data BEFORE they could read them. The Windows on GH CI probably gives threads larger blocks of time to process, so the first thread was able to finish its work in a single time block, while other two sometimes did not move. Later I have found there is yet another one, caused partially by a feature - when attached client read the result, it "purged" the result data, so nobody else was able to find them, except threads which also were already waiting on the same lock.
  * Finally I decided to remove the purge from the attach command. It is simply not responsible for removal of such data, the `JobCleanupTask` is, based on user defined rules.
  * To be able to have complete control, I changed regexp rules for the configuration values and allowed zeroes.
    * 0 `job-retention-period` means that when the cleanup task makes decision, deletes all completed jobs. Otherwise removes jobs older than configured time.
    * 0 `cleanup-initial-delay` means that when server notices change in the configuration, runs the job directly or uses scheduler and tells him to start the cleanup thread immediately.
    * 0 `cleanup-poll-interval` means to use the scheduler and start the cleanup repeatedly after its first execution.

* Plus two less important commits
  * Synchronization can take more time on GH CI Windows, 30 seconds is not always enough
  * Missing resource bundles reported in logs - stop trying to find them if we know they do not exist

Wrong intermediate idea I had, but which did not work, so was removed:
* to count clients attached to a single job, once there is a last one client, he made the cleanup (purrrrge). Problem: some clients already started the attach command, but did not have a CPU slot yet, so they were very sorry, came too late.

Note: See comment about CPUs below; I have changed the test, so now the count of attaching clients is same as the count of cpu threads plus two, so we hare sure it will be more than can be scheduled in parallel; the test is now repeatable on all CPUs I have available.


dmatej:
Passed once, however I am still not convinced it is everything:
```
23:55:40.997578    INFO           Thread-153 . exec(timeout=30,000, detached=false, args=[attach, 10])
23:55:40.997578    INFO           Thread-154 . exec(timeout=30,000, detached=false, args=[attach, 10])
23:55:40.997578    INFO           Thread-152 . exec(timeout=30,000, detached=false, args=[attach, 10])

 25%: progress-custom: Starting

 38%: progress-custom: Block [1] 8x1, step: 3

 50%: progress-custom: Block [1] 8x1, step: 4

 63%: progress-custom: Block [1] 8x1, step: 5
 75%: progress-custom: Block [1] 8x1, step: 6
 88%: progress-custom: Block [1] 8x1, step: 7
100%: progress-custom: Finished block [1] 8x1
Command progress-custom executed with status SUCCESS.
Command attach executed successfully.

 50%: progress-custom: Starting
 63%: progress-custom: Block [1] 8x1, step: 5
 75%: progress-custom: Block [1] 8x1, step: 6
 88%: progress-custom: Block [1] 8x1, step: 7
100%: progress-custom: Finished block [1] 8x1
Command progress-custom executed with status SUCCESS.
Command attach executed successfully.


 50%: progress-custom: Starting

 63%: progress-custom: Block [1] 8x1, step: 5
 75%: progress-custom: Block [1] 8x1, step: 6
 88%: progress-custom: Block [1] 8x1, step: 7
100%: progress-custom: Finished block [1] 8x1
Command progress-custom executed with status SUCCESS.
Command attach executed successfully.
```

dmatej:
Well, now I shortened the job, but increased the number of clients from 3 to 20. Locally no problem, but ci has problems!
Even logs are out of sync with long gaps of few seconds... so it might be low mem problem, or some cpu scheduling issue... caused probably by something on our side, because it always points to this test.... however some other tests are sometimes slower too, especially when they need cpu power.

dmatej:
Well, I added more logs to the test and now I can see another elephant in the room. The executor pool blocks until all jobs finish while authors including me expected that the get method will block.
In Java21 comes new `Executors.newThreadPerTaskExecutor` which is ready for this.

dmatej:
My machine:
```
CPUs: 32 total, 32 available
Memory: 95827M
AMD Ryzen 9 7945HX with Radeon Graphics, 32 cores, 93G, TUXEDO OS 24.04.1 LTS
```
GH CI, Linux:
```
CPUs: 4 total, 4 available
Memory: 15995M
AMD EPYC 7763 64-Core Processor, 4 cores, 15G, Ubuntu 24.04.2 LTS
```
GH CI, windows:
```
CPUs: 4 total, 4 available
Memory: 16383M
AMD EPYC 7763 64-Core Processor, 4 cores, 15G,  Windows Server 2025 , 64 bit Build 26100 (10.0.26100.4202)
```
Jenkins
```
CPUs: 40 total, 6 available
Memory: 8192M
Intel(R) Xeon(R) Silver 4210R CPU @ 2.40GHz, 40 cores, 8G, Ubuntu 24.04.1 LTS
```

And as seen in logs, what is happening is this:
* Server runs the progress-custom command
* Test creates a number of threads running the attach command, started, not executed
* Scheduler (java+os) takes part of threads and schedules them - the count is defined by the count of available CPU threads: on my all, on Jenkins 5, on GH 3.
* These threads then wait for server response.
* More threads can be scheduled, but once the progress-custom wakes up, finishes its job, system prefers to continue with already started threads. Linux sometimes schedules more, Windows not, but maybe is just slower.
* Windows usually don't schedule new attach clients until previous don't finish.
* But when the last of them finishes, removes results both from the cache (purge) and from the jobs.xml file. Then there are no results to provide to remaining attach jobs and they fail.

Possible solutions:
* Accept that - we can attach just to running jobs.
* However there already was code to get results for completed jobs. Then they should not be cleaned up by the attach job. There is yet another JobCleaner which removes expired jobs ...
* Schedule new thread, doing the cleanup ie. minute later, which would remove the data if there is no new attached client.

Probably the second option is the best one ...

dmatej:
Yet I have to update the description, but review is possible.

arjantijms (approved):

dmatej:
I am not sure yet how often we can see #25575, but lets create another PR for that; actually I have no idea why that happens sometimes.

PR #25580, opened 2025-06-24
Improved timeout management for admin commands

dmatej:
- The problem was that we had
  - timeout of the asadmin command
  - timeout of the asadmin start-domain, restart-domain, ...
- The test client timeout was just 5 seconds more than the internal asadmin, however just for the first start, not for all other in tests. Those used hard coded value which defined waiting just of the asadmin command.

The change
* added --timeout [seconds] parameter to some of commands
  * start-local-instance, start-instance, start-domain
  * stop...
  * restart-local-instance, restart-domain
* Using consistently defaults
  * env options, 60 s if not set
  * additional 5 seconds for restarts
* restarts
  * timeout parameter without changes
  * OR defaults = sum default for start, for stop, and 5 seconds.
  * if the server was already stopped, uses just start default as it uses also local command.

Possible failures:
Too short timeout - enough for stop, not enough to start, command fails. It doesn't mean server did not start, we don't know as we gave up. So I would use this with caution, however it is very flexible and easier than using shell variables. It is also easier on SSH nodes as sending env variables via SSH must be explicitly allowed.
```
/app/appservers/glassfish7/bin/asadmin restart-domain --timeout 1

Waiting finished after 392 ms.
Server instance is stopped, now we wait for the start on localhost:4848

Waiting finished after 1 000 ms.
Timed out waiting for the server to restart
Command restart-domain failed.
```

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -22,6 +22,7 @@ jobs:
       run: mvn -B -e -ntp install -Pstaging -Pqa '-Dcheckstyle.skip=true'
     - name: Upload server logs
       uses: actions/upload-artifact@v4
+      if: failure()
       with:
         name: server-logs
         path: '**/server.log*'
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -22,6 +22,7 @@ jobs:
       run: mvn -B -e -ntp install -Pstaging '-Dcheckstyle.skip=true' -pl :glassfish-itest-tools
     - name: Upload server logs
       uses: actions/upload-artifact@v4
+      if: failure()
       with:
         name: server-logs
         path: '**/server.log*'
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -20,6 +20,7 @@ jobs:
       run: mvn -B -e -ntp install -Pstaging -Pqa '-Dcheckstyle.skip=true'
     - name: Upload server logs
       uses: actions/upload-artifact@v4
+      if: failure()
       with:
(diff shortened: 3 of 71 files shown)


avpinchuk (approved):

dmatej:
Well, ok, Houston, we have a problem: the timeout set for asadmin is not propagated to remote instances.

The timeout environment options are not propagated and cannot be set via SFTP exec channel; 

Locally we could do that, I have no idea if it would work on windows without other changes, however I am quite sure it would not work on Linux. For local instances it could be done, however it would have to be different for Linux and Windows, or rather for different shells.

* https://stackoverflow.com/questions/9366914/setting-environment-variables-on-a-jsch-channelexec

Also the configuration of System Properties in Admin UI is very confusing:
* All names are upper cased, which is a convention for environment options
* But they are system options, used to override system options from configuration (default, server, you know)

So this is too not usable way.

So as a result, we have just one option: make timeouts a parameter of nadmin+asadmin commands:
* start*
* stop*
* attach

And default would be unset; that can be based on our previous work, we made all those commands quite reliable, the only breaking thing causing chaotic results were these layers of timeouts causing several bad scenarios:

* A starts B -> B is starting -> A times out -> A stops, throws error, domain did not start -> B finished the start, domain started successfully.
  * Here we have additional trick, if A timed out, kills starting process. However that could for another which still started. Reproduces locally when I used restart-domain with very short asadmin timeout around 5 seconds.
  * If B failed (crashed), A notices the error, it doesn't end with timeout.
* A starts B -> A times out -> B fails to start with exception, but we will never know that.

arjantijms (approved):

PR #25585, opened 2025-06-30
Fix #25584 The message which is "resource not found" is outputed when sending a javamail 

### LINKED ISSUE #25584: The message which is "resource not found" is outputed when sending a javamail

GlassFish 7.0.24

### JDK version

openjdk 21.0.2 2024-01-16

### OS

Red Hat Enterprise Linux release 9.2

### Database

_No response_

### Problem Description

There would have an unnecessary warning printed in server log due to missing default configuration file javamail.default.address.map.

```
[WARNING] [] [jakarta.mail] [tid: _ThreadID=47 _ThreadName=http-listener-1(3)] [levelValue: 900] [[
  expected resource not found: /META-INF/javamail.default.address.map]]
```

This behavior has been happening on glassfish version for a long time. It is still there for latest glassfish release v7.0.24.

### Steps to reproduce

This is reproduced when an email notification send to sender using any JavaMail session resources.

- create new JavaMail Session resources.
	- asadmin start-domain
        - asadmin create-mail-resource --mailhost "mailhost" --mailuser "test" --fromaddress "a@demo.com" --transprotocol "smtp" --enabled "true" --debug "true" mail/MySession
- Deploy any application that is responsible to send email notifications(app: [jakartamail.zip](https://github.com/user-attachments/files/20971209/jakartamail.zip))
	- asadmin deploy jakartamail.war
- Access application.(http://localhost:8080/jakartamail/test/SendMail?from-address=test@test.com&to-address=test1@test1.com)
- Review the warning messages that appeared in server.log.

### Impact of Issue

There are some unintentional messages to be appeared while sending email notification, it might confuse user to check send status.

kaido207:
* Fixes #25584 

This fix will add a default configuration file javamail.default.address.map in glassfish context classpath location which gets loaded to avoid warning. 
In glassfish for jakarta email service loading a default file resource involves accessing a file from the classpath. 

FYI, see also jakarta-mail api for reference.
https://github.com/jakartaee/mail-api/blob/892fae4ac7601976ee3270040a5b16d6349ce955/api/src/main/java/jakarta/mail/Session.java#L1141


pzygielo:
Shouldn't the resource from `glassfishX/glassfish/modules/angus-mail.jar` be used instead?

kaido207:
> Shouldn't the resource from `glassfishX/glassfish/modules/angus-mail.jar` be used instead?

Indeed. I thought about doing that too, but it seems [the mail-api](https://github.com/jakartaee/mail-api/blob/892fae4ac7601976ee3270040a5b16d6349ce955/api/src/main/java/jakarta/mail/Session.java#L1141) cannot refer to the resource within angus-mail.

arjantijms (approved):

PR #25589, opened 2025-07-02
Fix #25588 RemoveCount is not incremented when @Remove-annotated method is called

tnagao7:
Fixes #25588.

For EJB 2.x remove methods, `RemoveCount` is correctly incremented because the bean-destoyed event notification `ejbProbeNotifier.ejbBeanDestroyedEvent(...)` is executed in the `removeBean(EjbInvocation)` method in the `StatefulSessionContainer` class.
For EJB 3.x `@Remove` -annotated methods, adding a similar event notification in the `releaseContext(EjbInvocation)` method resolves the problem.

### DIFF
--- a/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatefulSessionContainer.java
+++ b/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatefulSessionContainer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, 2023 Contributors to the Eclipse Foundation
+ * Copyright (c) 2021, 2025 Contributors to the Eclipse Foundation
  * Copyright (c) 1997, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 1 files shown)


appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatefulSessionContainer.java
```diff
@@ -1718,6 +1718,8 @@ public void releaseContext(EjbInvocation inv) {
                         + invInfo.method + " due to (retainIfException" + " == true) and exception " + inv.exception);
                 } else {
                     try {
+                        ejbProbeNotifier.ejbBeanDestroyedEvent(getContainerId(),
```
pzygielo, line 1721: `ejbBeanDestroyedEvent` - for me it sounds like it shall be called once bean is destroyed. Not when bean is about to be, perhaps, destroyed (i.e. - after `destroyBean` returns).

tnagao7: Indeed your thoughts make sense because such behaviour can be implied from the event name.
On the other hand, as seen in the existing code below, the `ejbBeanDestroyedEvent` event is fired before executing `destroyBean` in EJB 2.x.
If we put `ejbBeanDestroyedEvent` after `destroyBean` for EJB 3.x, some may think the behaviour is inconsistent.

https://github.com/eclipse-ee4j/glassfish/blob/8e633f018c3a1fbeefca32155589f78e4f605cff/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatefulSessionContainer.java#L1113-L1132

dmatej (approved):

PR #25596, opened 2025-07-07
Fix cleanup on deployment error

OndroMih:
Before this, JNDI bindings were not cleaned up for all EJBs, leaking information to future deployments of the same name. If an app didn't deploy successfully and was deployed later, JNDI bindings remained from the previous failed deployment instead of being initialized from scratch. I saw an issue that EntityManagerFactory was closed because of a failed deployment, and then it was reused for the next deployment, leading to a closed EMF assigned to a newly deployed app. 

The cause for this was that the JNDI bindings were removed only after they were unregistered the same number of times as they were registered. During failed deployment, they were not unregistered in some cases, and an old JNDI reference was kept in the map. The fix removes all the bindings once again during application unregistration.

### DIFF
--- a/appserver/common/container-common/src/main/java/com/sun/enterprise/container/common/impl/ComponentEnvManagerImpl.java
+++ b/appserver/common/container-common/src/main/java/com/sun/enterprise/container/common/impl/ComponentEnvManagerImpl.java
@@ -426,9 +426,7 @@ public class ComponentEnvManagerImpl implements ComponentEnvManager {
     private void unregisterAllEjbs(Application application) {
         application.getEjbDescriptors().stream()
                 .map(this::getComponentEnvId)
-                .forEach(componentEnvId -> {
-                    compId2Env.remove(componentEnvId);
-        });
+                .forEach(compId2Env::remove);
     }
 
     private void undeployAllDescriptors(JndiNameEnvironment env) {
--- a/appserver/common/container-common/src/main/java/com/sun/enterprise/container/common/impl/ComponentEnvManagerImpl.java
+++ b/appserver/common/container-common/src/main/java/com/sun/enterprise/container/common/impl/ComponentEnvManagerImpl.java
@@ -379,19 +379,19 @@ public class ComponentEnvManagerImpl implements ComponentEnvManager {
     }
 
     @Override
-    public void unbindFromComponentNamespace(JndiNameEnvironment JndiEnvironment) throws NamingException {
+    public void unbindFromComponentNamespace(JndiNameEnvironment jndiEnvironment) throws NamingException {
         // undeploy all descriptors
-        undeployAllDescriptors(JndiEnvironment);
+        undeployAllDescriptors(jndiEnvironment);
 
         // Unpublish any global entries exported by this environment
         Collection<JNDIBinding> globalBindings = new ArrayList<>();
-        addJNDIBindings(JndiEnvironment, ScopeType.GLOBAL, globalBindings);
+        addJNDIBindings(jndiEnvironment, ScopeType.GLOBAL, globalBindings);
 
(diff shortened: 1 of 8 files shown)


appserver/common/container-common/src/main/java/com/sun/enterprise/container/common/impl/ComponentEnvManagerImpl.java
```diff
@@ -379,47 +379,58 @@ private ResourceDeployer getResourceDeployer(Object resource) {
...
             namingManager.unbindComponentObjects(componentEnvId);
             this.unregister(componentEnvId);
         }
 
     }
 
+    private void unregisterAllEjbs(Application application) {
+        application.getEjbDescriptors().stream()
+                .map(this::getComponentEnvId)
+                .forEach(componentEnvId -> {
```
pzygielo: `(compId2Env::remove)`?

appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/asadmin/AsadminResultMatcher.java
```diff
@@ -13,14 +13,13 @@
...
  */
-
 package org.glassfish.main.itest.tools.asadmin;
 
 import org.hamcrest.CustomTypeSafeMatcher;
 
-
 /**
- * Matcher checking that {@link Asadmin} command succeeded. Prints it's output otherwise.
+ * Matcher checking that {@link Asadmin} command succeeded. Prints it's output
```
pzygielo: If this is changed, perhaps `it's` should become `its`.

appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/asadmin/AsadminResultMatcher.java
```diff
@@ -30,17 +29,41 @@ private AsadminResultMatcher() {
...
     @Override
     protected boolean matchesSafely(AsadminResult item) {
         return !item.isError();
     }
 
-
     /**
-     * @return matcher checking that {@link Asadmin} command succeeded. Prints it's output otherwise.
+     * @return matcher checking that {@link Asadmin} command succeeded. Prints
+     * it's output otherwise.
```
pzygielo: as [above](https://github.com/eclipse-ee4j/glassfish/pull/25596/files#r2191958450)

appserver/tests/application/src/test/java/org/glassfish/main/test/app/deploy/afterfailed/DeployAfterFailedDeploymentTest.java
```diff
@@ -0,0 +1,96 @@
...
+                .addAsResource(TEST_PACKAGE, "persistence.xml", "META-INF/persistence.xml");
+        LOG.log(INFO, webArchive.toString(true));
+        File warFile = new File(tempDir, OK_APP_NAME + ".war");
+        webArchive.as(ZipExporter.class).exportTo(warFile, true);
+        return warFile;
+    }
+
+    /**
+     * This test verifies that the JNDI environment is cleaned up for the application context in case of failed deployment.
+     * It used to happen that after a failed deployment, JNDI environment for the app name and context root stayed in memory,
```
pzygielo, line 94: So it's fine to have this long-line here, but https://github.com/eclipse-ee4j/glassfish/pull/25596/files#diff-b4341c34f0362bdfd050694ba705e63c0db115daaab33fbb54a80e7e00910c53R38 needs to be changed?

```
     * It used to happen that after a failed deployment, JNDI environment for the app name and context root stayed in memory,

AND

     * @return matcher checking that {@link Asadmin} command succeeded. Prints it's output otherwise.

```

arjantijms (approved):

PR #25594, opened 2025-07-05
A working integration of Data into GlassFish

OndroMih:
A working integration of Data into GlassFish. Still rough and not fully optimized, doesn’t support NoSQL yet.

Requires building the Jakarta Persistence driver with `mvn install -Pshade` from this branch: https://github.com/OmniFish-EE/jnosql-extensions-jakarta-data/pull/1

An app that uses Data repository works and uses the default EntityManager.
Metadata about repositories is provided by GlssFish instead of ClassGraph

Still lots of things to do. From the top of my head:
- EntitiesMetadata is temporarily implemented by reflection, with a copy of code from JNoSQL reflection module. Should be also provided by GlassFish and then all the code from the `org.glassfish.main.jnosql.jakartapersistence.mapping.reflection` and subpackages could be removed
- Support repositories for NoSQL entities. Currently disabled, would conflict with JPA
- Select EM for repository, e.g. by an annotation with PU name - OK - done in #25597

PR #25597, opened 2025-07-10
A working integration of Data into GlassFish (multiple EM support)

OndroMih:
Supercedes https://github.com/eclipse-ee4j/glassfish/pull/25594

Based on https://github.com/OmniFish-EE/jnosql-extensions-jakarta-data/pull/2.

Please ignore all the code in the `reflection` and `nosql` packages - it's going to be removed later, by an implementation based on class metadata already collected by GlassFish. 

New functionality:

- Supports specifying entityManager via persistence unit name or qualifiers.
- Adds @Repository as bean-defining annotation - means that CDI scanning is enabled also if there's only Data repositories in an app, with no other CDI beans, in case of annotated discovery

An example of specifying entityManager:

By qualifier:

```
@Repository
@MyPersistenceUnit
public interface MyRepository...
```

```
@Inject @MyPersistenceUnit
MyRepository myrepository;
```

By persistence unit name:

```
@Repository(dataStore = "myPersistenceUnit") // myPersistenceUnit is the unit name
public interface MyRepository...
```


dmatej (approved):

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/ApplicationLifecycle.java
```diff
@@ -196,6 +197,8 @@ public class ApplicationLifecycle implements Deployment, PostConstruct {
     private ExecutorService executorService;
     private Collection<ApplicationLifecycleInterceptor> alcInterceptors = emptyList();
 
+    private ThreadLocal<Deque<ExtendedDeploymentContext>> currentDeploymentContext;
```
dmatej, line 200: How are you ensuring cleanup on failed deployments? I think we had some similar leak at some point in the past ...

OndroMih: After each `pushCurrentDeploymentContext()` call, there's `popCurrentDeploymentContext()` call in a finally block inside the same method. The last call to `popCurrentDeploymentContext` will remove the thread local variable.

PR #25608, opened 2025-07-18
Execute SQL against JDBC datasources

arjantijms:
[Execute SQL against JDBC datasources](https://github.com/eclipse-ee4j/glassfish/commit/186ba848700770e7f27dd43902a2bc3c837b4d9b)


OndroMih:
What’s this about, @arjantijms ? Does it add an admin command to execute SWL on a datasource? If yes, would you document it in the GlassFish Reference docs?

PR #25612, opened 2025-07-19
A transaction interceptor for Data repositories

OndroMih:
Enables transaction interceptor from Java SE impl - it will create a transaction for repository methods if no transaction exists.

Doesn't support JTA entity managers so it needs to be improved or replaced.


pzygielo:
Edit PR and change base?

OndroMih:
I found a better approach that already creates a JTA transaction for JTA EMs: https://github.com/eclipse-ee4j/glassfish/pull/25613.

I'm closing this PR.

PR #25613, opened 2025-07-20
Jakarta Data transactional interceptor plus fixes in class scanner

OndroMih:
This depends on a new snapshot release of the Jakarta Data JPA driver with this commit: https://github.com/OmniFish-EE/jnosql-extensions-jakarta-data/commit/f5b957b7fb5f10c59b15f8074a2c0fccfe6af91c.

Runs repository method in a new transaction if no transaction is active. Basically adds @Transactional interceptor on all methods of repository beans. For non-JTA EMs, it wraps the method calls into a transaction created using the EM.

Required to pass some any TCK test for which we don't/can't create a global transaction (a few TCK tests expect there's no global Tx and that each repository method runs in its own transaction, others can run with a global transaction but also pass when a transaction is created for each repository method call).

### DIFF
--- a/appserver/persistence/jnosql-integration/pom.xml
+++ b/appserver/persistence/jnosql-integration/pom.xml
@@ -29,6 +29,10 @@
     <packaging>glassfish-jar</packaging>
 
     <name>Eclipse JNoSQL Jakarta Persistence Integration</name>
+    
+    <properties>
+        <persistence.osgi.bundle.classpath>${project.build.directory}/dependencies/jnosql-jakarta-persistence-${jnosql.version}.jar${path.separator}${project.build.directory}/dependencies/jnosql-mapping-core-${jnosql.version}.jar${path.separator}${project.build.directory}/dependencies/jnosql-mapping-api-core-${jnosql.version}.jar${path.separator}${project.build.directory}/dependencies/jnosql-communication-core-${jnosql.version}.jar</persistence.osgi.bundle.classpath>
+    </properties>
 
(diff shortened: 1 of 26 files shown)


OndroMih:
I also tried a simpler solution using interceptors, something like this in the JNoSQL Driver Extension:

```
       T instance = (T) Proxy.newProxyInstance(type.getClassLoader(),
                new Class[]{type},
                handler);

        InterceptionFactory<T> factory =
                CDI.current()
                   .getBeanManager()
                   .createInterceptionFactory(context, type);

        factory.configure()
               .add(new AnnotationLiteral<Transactional>() {});

        return factory.createInterceptedInstance( instance );
    }
```

But it doesn't work reliably so I would stick to the solution in this PR.

It would work if Weld applied the interceptor on all methods of the instance object. However, it applies it only on methods declared in the `type` interface but not to methods declared in interfaces it extends. If a repository interface, e.g. MyRepository, extends from some Data interfaces, e.g. DataRepository or BasicRepository, methods defined on those repositories would not be intercepted with this approach. Maybe we could create factories for all superinterfaces and combine them, but then the existing approach in this PR is better and more reliable.

OndroMih:
The JNoSQL OSGi bundle snapshot has been updated, so we can merge as long as all looks good.

OndroMih:
Added more improvements, this time to the TCK runner. It creates a JTA transaction for each test method and adds some classes missing in the deployed apps. Passes now 86 / 92 tests.

PR #25631, opened 2025-07-25
Raise the generated bytecode level to JDK 17

avpinchuk:


dmatej (approved):
Oh, that is what I was looking for, but I was searching RELEASE_n constant! Thanks!

PR #25643, opened 2025-07-30
Replaced references to release version with a variable

### LINKED ISSUE #25389: In docs, replace references to release version with a variable
GlassFish docs contain many references to the current latest version 7, e.g. [here](https://github.com/eclipse-ee4j/glassfish/blob/e38706f39035e3acf1b58ae395032e2eb799e0ee/docs/administration-guide/src/main/asciidoc/title.adoc?plain=1#L1). These would need to be updated once we release every new major version.

We can replace the literal version name with a variable, e.g. {releaseVersion}, similarly to {productName}. It can be even derived from the maven artifact version, so the [pom.xml](https://github.com/eclipse-ee4j/glassfish/blob/e38706f39035e3acf1b58ae395032e2eb799e0ee/docs/parent/pom.xml#L109) can define the variable for asciidoctor like this: <productName>${project.version}</productName>.

Ariho-Seth:
This PR:

- Introduces new variable names, i.e,  `<productMajorVersion>7</productMajorVersion>`
`<productCurrentVersion>${project.version}</productCurrentVersion>`

- Updates all the adoc documentation files with a variable  `{productMajorVersion}` which currently points to `version 7`, which is expected to change every after the release of a new GlassFish release.

- Fixes <https://github.com/eclipse-ee4j/glassfish/issues/25389>

### DIFF
--- a/docs/parent/pom.xml
+++ b/docs/parent/pom.xml
@@ -174,24 +174,6 @@
             </plugin>
 
             <plugin>
-                    <groupId>org.codehaus.mojo</groupId>
-                    <artifactId>build-helper-maven-plugin</artifactId>
-                    <version>3.6.1</version>
-                    <executions>
-                        <execution>
-                            <id>parse-version</id>
-                            <goals>
-                                <goal>parse-version</goal>
-                            </goals>
-                            <phase>validate</phase>
-                            <configuration>
-                                <propertyPrefix>parsedVersion</propertyPrefix>
-                            </configuration>
-                        </execution>
-                    </executions>
-                </plugin>
-
-            <plugin>
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>build-helper-maven-plugin</artifactId>
                 <executions>
@@ -210,6 +192,16 @@
                             </artifacts>
                         </configuration>
                     </execution>
+                        <execution>
+                            <id>parse-version</id>
+                            <goals>
+                                <goal>parse-version</goal>
+                            </goals>
+                            <phase>validate</phase>
+                            <configuration>
+                                <propertyPrefix>parsedVersion</propertyPrefix>
+                            </configuration>
+                        </execution>
                 </executions>
             </plugin>
             <plugin>
--- a/docs/parent/pom.xml
+++ b/docs/parent/pom.xml
@@ -41,7 +41,7 @@
         <html.rel.file.prefix></html.rel.file.prefix>
         <mq.docs.url>https://eclipse-ee4j.github.io/openmq/guides/</mq.docs.url>
         <bookDirectory>${project.build.directory}/book</bookDirectory>
-        <productMajorVersion>7</productMajorVersion>
+        <productMajorVersion>${parsedVersion.majorVersion}</productMajorVersion>
         <productCurrentVersion>${project.version}</productCurrentVersion>
     </properties>
 
@@ -172,6 +172,24 @@
                     </execution>
                 </executions>
             </plugin>
+
+            <plugin>
+                    <groupId>org.codehaus.mojo</groupId>
+                    <artifactId>build-helper-maven-plugin</artifactId>
+                    <version>3.6.1</version>
+                    <executions>
+                        <execution>
+                            <id>parse-version</id>
+                            <goals>
+                                <goal>parse-version</goal>
+                            </goals>
+                            <phase>validate</phase>
+                            <configuration>
+                                <propertyPrefix>parsedVersion</propertyPrefix>
+                            </configuration>
+                        </execution>
+                    </executions>
+                </plugin>
 
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
--- a/docs/parent/pom.xml
+++ b/docs/parent/pom.xml
@@ -41,6 +41,8 @@
         <html.rel.file.prefix></html.rel.file.prefix>
         <mq.docs.url>https://eclipse-ee4j.github.io/openmq/guides/</mq.docs.url>
         <bookDirectory>${project.build.directory}/book</bookDirectory>
+        <productMajorVersion>7</productMajorVersion>
+        <productCurrentVersion>${project.version}</productCurrentVersion>
     </properties>
 
     <build>
--- a/docs/parent/pom.xml
+++ b/docs/parent/pom.xml
@@ -41,8 +41,8 @@
         <html.rel.file.prefix></html.rel.file.prefix>
         <mq.docs.url>https://eclipse-ee4j.github.io/openmq/guides/</mq.docs.url>
         <bookDirectory>${project.build.directory}/book</bookDirectory>
-        <productMajorVersion>${parsedVersion.majorVersion}</productMajorVersion>
-        <productCurrentVersion>${project.version}</productCurrentVersion>
+        <productMajorVersion>${product.majorVersion}</productMajorVersion>
+        <product.currentVersion>${project.version}</product.currentVersion>
     </properties>
 
     <build>
@@ -174,6 +174,25 @@
             </plugin>
 
             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-antrun-plugin</artifactId>
+                <version>3.0.0</version>
+                <executions>
+                    <execution>
+                        <phase>initialize</phase>
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
+                        <configuration>
+                            <target>
+                                <property name="productMajorVersion" value="${product.majorVersion}"/>
+                            </target>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+
+            <plugin>
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>build-helper-maven-plugin</artifactId>
                 <executions>
@@ -199,7 +218,7 @@
                             </goals>
                             <phase>validate</phase>
                             <configuration>
-                                <propertyPrefix>parsedVersion</propertyPrefix>
+                                <propertyPrefix>product</propertyPrefix>
                             </configuration>
                         </execution>
                 </executions>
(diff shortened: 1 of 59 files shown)


docs/parent/pom.xml
```diff
@@ -41,6 +41,8 @@
         <html.rel.file.prefix></html.rel.file.prefix>
         <mq.docs.url>https://eclipse-ee4j.github.io/openmq/guides/</mq.docs.url>
         <bookDirectory>${project.build.directory}/book</bookDirectory>
+        <productMajorVersion>7</productMajorVersion>
```
pzygielo: We won't remember to update it.

Could https://www.mojohaus.org/build-helper-maven-plugin/parse-version-mojo.html be used instead?

Ariho-Seth: Hello @pzygielo, Is there a way I can run this locally so I can always verify before I push the code?

OndroMih: You can build the Installation guide PDF or HTML with maven. 

Go to the /docs directory, and then run:

```
mvn -T4C clean install -pl :installation-guide -am
```

- `-pl :installation-guide` will instruct to build the installation-guide module (`installation-guide` is the artifact name)
- `-am` will also build all the parent modules it needs

Then you'll see the PDF and HTML in `/docs/installation-guide/target/html/` or `target/pdf`

Ariho-Seth: Hello, I think something ain't right with my local repo!
Can you help me please, in case you know what is wrong?

```
arihoseth@Macbook ~/ariho/Programming/Opensource-Projects/glassfish/docs % mvn -T4C clean install -pl :installation-guide -am
[INFO] Scanning for projects...
[ERROR] [ERROR] Could not find the selected project in the reactor: org.glassfish.main.distributions:glassfish @ 
[ERROR] Could not find the selected project in the reactor: org.glassfish.main.distributions:glassfish -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MavenExecutionException
arihoseth@Macbook ~/ariho/Programming/Opensource-Projects/glassfish/docs % 

```

docs/parent/pom.xml
```diff
@@ -171,6 +173,24 @@
                 </executions>
             </plugin>
 
+            <plugin>
+                    <groupId>org.codehaus.mojo</groupId>
+                    <artifactId>build-helper-maven-plugin</artifactId>
```
pzygielo: :-1: 
This plugin is already declared JUST BELOW.

docs/parent/pom.xml
OndroMih: Do you have .mvn/maven.config file in the root directory of your glassfish repository? If yes, then rename it before running this command, or comment the commands in it. For some reason, Maven reads this file also when you execute builds in subfolders.

docs/parent/pom.xml
```diff
@@ -41,6 +41,8 @@
         <html.rel.file.prefix></html.rel.file.prefix>
         <mq.docs.url>https://eclipse-ee4j.github.io/openmq/guides/</mq.docs.url>
         <bookDirectory>${project.build.directory}/book</bookDirectory>
+        <productMajorVersion>${parsedVersion.majorVersion}</productMajorVersion>
```
OndroMih: This won't work - the `parsedVersion.majorVersion` is resolved later and is not available at this time. You need to define this property after the `parse-version` goal is executed, e.g. with the antrun plugin:

```
<plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-antrun-plugin</artifactId>
  <version>3.0.0</version>
  <executions>
    <execution>
      <phase>initialize</phase>
      <goals>
        <goal>run</goal>
      </goals>
      <configuration>
        <target>
          <property name="productMajorVersion" value="${parsedVersion.majorVersion}"/>
        </target>
      </configuration>
    </execution>
  </executions>
</plugin>
```

docs/parent/pom.xml
Ariho-Seth: I've surely done that, though the problem is persisting.

docs/parent/pom.xml
```diff
@@ -190,6 +192,16 @@
...
                         </configuration>
                     </execution>
+                        <execution>
+                            <id>parse-version</id>
+                            <goals>
+                                <goal>parse-version</goal>
+                            </goals>
+                            <phase>validate</phase>
+                            <configuration>
+                                <propertyPrefix>parsedVersion</propertyPrefix>
```
OndroMih: I suggest using "product" as the prefix. Then the property would be "product.majorVersion". It could even be used in adoc files directly, as {product-majorVersion}, without setting the `productMajorVersion` property. Then we wouldn't even need to set the `productMajorVersion` property with the antrun plugin.

```suggestion
                                <propertyPrefix>product</propertyPrefix>
```

docs/parent/pom.xml
pzygielo: Or use `parsedVersion.majorVersion` directly without introducing `productMajorVersion`.

Oh - I see the comment below says that as well.

OndroMih: We replied at the same time :)

docs/parent/pom.xml
```diff
@@ -41,6 +41,8 @@
         <html.rel.file.prefix></html.rel.file.prefix>
         <mq.docs.url>https://eclipse-ee4j.github.io/openmq/guides/</mq.docs.url>
         <bookDirectory>${project.build.directory}/book</bookDirectory>
+        <productMajorVersion>${parsedVersion.majorVersion}</productMajorVersion>
+        <productCurrentVersion>${project.version}</productCurrentVersion>
```
OndroMih: I realized that this should rather be set to ${glassfish.version.latest}, which we now manually set before each release in https://github.com/eclipse-ee4j/glassfish/blob/master/docs/pom.xml

Ariho-Seth: So that means that the plugin is not necessary anymore, if we are going with that suggestion.

OndroMih: > I realized that this should rather be set to ${glassfish.version.latest}, which we now manually set before each release in https://github.com/eclipse-ee4j/glassfish/blob/master/docs/pom.xml

Please forget about this, it should be ${project.version}. It will be set to a non-SNAPSHOT version during a release utomatically. `${glassfish.version.latest}` is used only in the website to refer to previous versions.

pzygielo: :confused: Strange, as for example 7.0.25 - along with its docs, is modified in different branch. Does it mean we do code for 7.1 in master, but the docs here are for 7.0.x (7.0.25 at the moment)?

I think I'm lost.

OndroMih: > So that means that the plugin is not necessary anymore, if we are going with that suggestion.

We can do this:

* set `<product.currentVersion>${project.version}</product.currentVersion>` - note the change from `productCurrentVersion` to `product.currentVersion`. The refer to this in adoc with `{product-currentVersion}`
* not set `productMajorVersion` here. Instead, use only the build-helper plugin, with the `product` prefix. It would create a property `product.majorVersion`, which we can refer to in adoc as `{product-majorVersion}`

OndroMih: > I think I'm lost.

@pzygielo , I got confused too :)

This is what's happening:

* In master, docs are for the upcoming release, so for 7.1.0-SNAPSHOT. They are rebuilt and published on the website at https://glassfish.org/documentation.html (see Documentation for other GlassFish versions) with each PR
* During a release, the version is set to the release version, e.g. 7.1.0 in the future, or 7.0.25 for the last release. The docs are published as maven artifacts in Maven central. 
* The website job downloads docs for all the previous releases and publishes them under the URL with the version, e.g. https://glassfish.org/docs/7.0.25/administration-guide.html
* The website also creates a symlink to the latest version but it has to know which version is the latest one. Currently, the `glassfish.version.latest` property is set to 7.0.25, so the symlink https://glassfish.org/docs/latest/administration-guide.html is to the version 7.0.25. The documentation page only links to the "latest" version. In the future, we can implement a version selector and people would be able to select an older version of the docs

So, in short
* master should refer to the current version in the master
* during a release, docs are published in Maven Central under a non-snapshot version
* website needs to know the latest version (because the master is already on the next snapshot version) - that's why we manually set the `glassfish.version.latest` property during a release

OndroMih: This also means that docs for a specific version need to be up to date at the time of a release, we can't easily modify/fix docs for an already released version. Everything we add to the master branch will be connected with docs for the next version.

Ariho-Seth: Do the new changes help achieve what we want?

docs/parent/pom.xml
```diff
@@ -41,6 +41,8 @@
         <html.rel.file.prefix></html.rel.file.prefix>
         <mq.docs.url>https://eclipse-ee4j.github.io/openmq/guides/</mq.docs.url>
         <bookDirectory>${project.build.directory}/book</bookDirectory>
+        <productMajorVersion>${product.majorVersion}</productMajorVersion>
```
OndroMih: Remove this property, and in adoc refer to {product-majorVersion}. This property is not set properly, because the build-helper plugin has not run yet.

docs/parent/pom.xml
OndroMih: @Ariho-Seth , this needs to be changed: https://github.com/eclipse-ee4j/glassfish/pull/25643/files#r2245743196

docs/parent/pom.xml
Ariho-Seth: Surely I have done that and replaced it in all the adoc files.

OndroMih (approved):
Brillian, @Ariho-Seth !

All seems to work now when I generate the guides locally.

Ariho-Seth:
Thanks @OndroMih and @pzygielo  for working with me throughout the whole process.

OndroMih:
Thank you, @Ariho-Seth, this was an important and very iseful contribution from you!

PR #25653, opened 2025-08-03
Upload PDFs built in workflow

pzygielo:


avpinchuk (approved):

pzygielo:
> If only the PDFs could be updloaded individually, not just as a ZIP file with all of them. It's a known issue that the `actions/upload-artifact` action doesn't support that, a pity.

Yeah, even if we'd enumerate all of them for the separate artifacts - the action would zip them, and we'd end up with several archives to work with. :disappointed: 

Let's wait for:
- https://github.com/actions/upload-artifact/issues/426

OndroMih:
@pzygielo , we would need to copy the action for each PDF. Is it worth the effort?

PR #25654, opened 2025-08-04
Display GitHub workflow names in pull requests

OndroMih:
It works, see the Checks panel below and compare with, for example, https://github.com/eclipse-ee4j/glassfish/pull/25653#event-18959218138

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -1,11 +1,12 @@
 
-name: Build on Ubuntu
+name: Build GlassFish on Ubuntu
 on:
   pull_request:
     branches: [ "master", "7.0", "8.0" ]
 jobs:
   build:
     runs-on: ubuntu-24.04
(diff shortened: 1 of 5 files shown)


.github/workflows/build-linux.yml
```diff
@@ -1,11 +1,12 @@
 
-name: Build on Ubuntu
+name: Build GlassFish on Ubuntu
 on:
   pull_request:
     branches: [ "master", "7.0", "8.0" ]
 jobs:
   build:
     runs-on: ubuntu-24.04
+    name: Build on Ubuntu
```
avpinchuk, line 9: In all other places you use `Build` **GlassFish** `on {system_name}`. Maybe here too?

OndroMih: Sure, I missed it here. Thanks for pointing this out.

hs536 (approved):

PR #25667, opened 2025-08-14
Fixed appclient classpath

dmatej:
* Fix for #25666 - when using no arguments, the appclient crashed. When we fixed it, it still crashed on Windows when using arguments.
* Fixed appclient.bat having issues to parse output from CLIBootstrap
* New test for at least simple execution of the appclient script

### DIFF
--- a/appserver/tests/application/src/test/java/org/glassfish/main/test/app/appclient/AppClientTest.java
+++ b/appserver/tests/application/src/test/java/org/glassfish/main/test/app/appclient/AppClientTest.java
@@ -21,9 +21,9 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.Charset;
 
+import org.glassfish.main.itest.tools.GlassFishTestEnvironment;
 import org.junit.jupiter.api.Test;
 
-import static org.glassfish.main.itest.tools.GlassFishTestEnvironment.getGlassFishDirectory;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.stringContainsInOrder;
 import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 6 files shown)


pzygielo (dismissed):
With https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/job/PR-25667/1/artifact/bundles/glassfish.zip
```
$ glassfish7/glassfish/bin/appclient
appclient [ <classfile> | -client <appjar> ] [-mainclass <appClass-name>|-name <display-name>] [-xml <xml>] [-textauth] [-user <username>] [-password <password>|-passwordfile <password-file>] [-targetserver host[:port][,host[:port]...] [app-args]
  or  :
	appclient [ <valid JVM options and valid ACC options> ] [ <appClass-name> | -jar <appjar> ] [app args]
```

dmatej:
This should be better - the gf-client is still duplicated, because the second process inherits classpath from the previous. That is bit problematic, because user can provide his own where the file could be missing. One option would be to search the classpath and filter it, but it is not worth of the work - another option is to make gf-client.jar JPMS compliant, which I have i stash now, then we can remove it from all "automated" classpaths here; however JPMS hates then declaring Main class in manifest from another module ... I am not sure if it even has to be there, so more coverage is required ...

I made yet one commit - it was printing help to STDERR, but STDERR is meant for developers while this is user help so it should be printed to STDOUT.

appserver/tests/application/src/test/java/org/glassfish/main/test/app/appclient/AppClientTest.java
```diff
@@ -0,0 +1,61 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ *
+ */
+public class AppClientTest {
+
+    @Test
+    void noArgsAndHelp() throws Exception {
+        File appclient = new File(getGlassFishDirectory(), "bin/appclient");
```
avpinchuk: Script name should be `bin\appclient.bat` in Windows.

avpinchuk: Also, for Windows this does not work. Whole command in parentheses should be enclosed with `'^"` and `^"'`.

https://github.com/eclipse-ee4j/glassfish/blob/1d413bbf7e723e0410ca001d8c0c9178622c0fc6/appserver/appclient/client/appclient-scripts/src/main/resources/glassfish/bin/appclient.bat#L39

dmatej: Oh, yeah, there is the suffix ...

avpinchuk (approved):

PR #25603, opened 2025-07-15
Do not swallow InterruptedException

pzygielo:
This is since forever (https://github.com/javaee/glassfish/blob/10f9ad762f9fd07f4ccbaaa712a48d6715ae80cf/v2/appserv-core/src/java/com/sun/enterprise/resource/AbstractResourcePool.java#L487, https://github.com/javaee/glassfish/blob/10f9ad762f9fd07f4ccbaaa712a48d6715ae80cf/v2/appserv-core/src/java/com/sun/enterprise/resource/AbstractResourcePool.java#L903) so maybe I'm wrong?

PR #25678, opened 2025-08-25
Integrate Jersey 3.1.11

senivam:


pzygielo:
- https://github.com/eclipse-ee4j/glassfish/pull/25603#issuecomment-3081041067:

> I would prefer PRs to master first; if we would want to release 7.0.26, I would go over changes in master to find candidates for 7.0.26 ... but not in the opposite direction.

pzygielo:
But I guess this *would be* picked for 7.0.x :smile: 

Please open the same for master.

PR #25680, opened 2025-08-26
Simplify and refactor the GlassFish upgrade guide

OndroMih:
Move the practical stuff to the top of the guide, simplify it, remove redundant or old info.
Move the reference stuff into appendix.
Delete list of examples which was not used anywhere and is not relevant.

Built PDF guides: https://github.com/eclipse-ee4j/glassfish/actions/runs/17240898111/artifacts/3854186266


docs/upgrade-guide/src/main/asciidoc/upgrading-legacy-installation.adoc
```diff
@@ -1014,67 +169,48 @@ host2$ asadmin --host dashost create-local-instance --node na2 --cluster cluster
...
-
-[[cluster-profile-security-setting]]
+[[upgrading-installations-that-use-nss-cryptographic-tokens]]
+==== Upgrading Installations Using NSS Cryptographic Tokens
+{productName} {product-majorVersion} does not support NSS. If you're upgrading from a version that supports them, follow these steps:
+
+1. **Prepare for Upgrade**:
+   - Install {productName} {product-majorVersion} in a new directory.
+   - Copy the source domain to the new installation.
+   - Update `domain.xml` to remove NSS references and add JKS keystore paths.
```
avpinchuk, line 189: Now we use PKCS#12 keystore format

docs/upgrade-guide/src/main/asciidoc/upgrading-legacy-installation.adoc
```diff
@@ -1014,67 +169,48 @@ host2$ asadmin --host dashost create-local-instance --node na2 --cluster cluster
...
+2. **Perform Post-Upgrade Configuration**:
+   - Migrate NSS keys to JKS using `keytool` and `certutil`.
```
avpinchuk: PKCS#12

docs/upgrade-guide/src/main/asciidoc/upgrading-legacy-installation.adoc
OndroMih: I think GlassFish 7.1 still supports JKS but uses P12 by default. Is that right, @dmatej ?

If yes, then I'll change this to "JKS or PKCS#12 keystore paths". If only `PKCS#12` is now supported, I'd add instructions to migrate to `PKCS#12`.

docs/upgrade-guide/src/main/asciidoc/upgrading-legacy-installation.adoc
OndroMih: I'll change to:

```suggestion
   - Migrate NSS keys to PKCS#12 using `keytool` and `certutil`.
```

docs/upgrade-guide/src/main/asciidoc/upgrading-legacy-installation.adoc
dmatej: It can be overridden, but only insane person would do that, I would say :D

dmatej: However the certutil was deleted, it did not work properly, if I remember well, and the NSS is not supported any more. At some point in the past it was considered as a commercial variant from Netscape, then JCEKS, but evolution went different way.
Over time changed order of arguments, their names, etc.

dmatej: It would be good to download ie. GlassFish2 or 3 and try to follow instructions. Then update them. I would not rewrite them without testing.

dmatej: Upgrade from 7.0 to 7.1 could be useful too.

OndroMih: What about removing this section about NSS altogether? If it's applicable only to ancient GlassFish versions, we don't need to keep it. I would certainly not bother with testing an upgrade from GF 2 or 3.

docs/upgrade-guide/src/main/asciidoc/upgrading-legacy-installation.adoc
```diff
@@ -1,932 +1,87 @@
...
-
-* xref:#upgrade-paths[Upgrade Paths]
-* xref:#upgrade-terminology[Upgrade Terminology]
-* xref:#summary-of-upgrade-tools-and-procedures[Summary of Upgrade Tools and Procedures]
-* xref:#supported-releases-for-upgrade-to-glassfish-server-7[Supported Releases for Upgrade to {productName} {product-majorVersion}]
-* xref:#GSUPG00063[Upgrading From Version 8.x or Older Product Releases]
-* xref:#upgrading-glassfish-server-inside-a-closed-network[Upgrading {productName} Inside a Closed Network]
+* <<upgrade-paths,Upgrade Paths>>
+* <<upgrade-procedure,Upgrade Procedure>>
+* <<special-cases,Special Cases (Clusters, Node Agents, NSS)>>
```
OndroMih, line 19:
```suggestion
* <<special-cases,Special Cases (Clusters, Node Agents)>>
```

docs/upgrade-guide/src/main/asciidoc/upgrading-legacy-installation.adoc
```diff
@@ -1014,67 +169,48 @@ host2$ asadmin --host dashost create-local-instance --node na2 --cluster cluster
...
+   - Install {productName} {product-majorVersion} in a new directory.
+   - Copy the source domain to the new installation.
+   - Update `domain.xml` to remove NSS references and add JKS keystore paths.
+2. **Perform Post-Upgrade Configuration**:
+   - Migrate NSS keys to PKCS#12 using `keytool` and `certutil`.
+   - Update the master password if needed.
+3. **Upgrade PKCS#11 Hardware Tokens**:
+   - Configure the token using JDK-JSSE mechanisms.
+   - Update `domain.xml` to reference the hardware token.
+
```
OndroMih, line 196:
```suggestion
```

OndroMih (commented):
What about removing the NSS parts altogether? They are applicable only to ancient GlassFish versions, we don't need to support them in the guide.

arjantijms (approved):

PR #25688, opened 2025-09-02
Decrease log level of logs in command execution

OndroMih:
Some commands are regularly being executed, either from Admin Console, or IDEs (e.g. __locations command). INFO level would log them by default, which only trashes the logs.

Example:

```
Executing command __locations with parameters {}, outboundPayload: org.glassfish.admin.payload.ZipPayloadImpl$Outbound@44d4e19f|#]
```


dmatej:
Can you grep how many logs it created? Just for sure, it was quite useful to see what exactly was executed finally when I used just few arguments.

OndroMih:
@dmatej , did you need to have this enabled by default? Isn't it enough to keep it at DEBUG level? It's also possible to enable the Command Logger, which also logs commands, and allows more fine-grained config (only write commands, read commands, all commands: https://github.com/eclipse-ee4j/glassfish/pull/24993). Command logger also masks secrets, I'm not sure if this logger here would log secrets or not.

dmatej (approved):

PR #25698, opened 2025-09-10
Unmake gf-client runnable JAR

pzygielo:


OndroMih (approved):
I agree. As commented by @dmatej on the related issue, running the JAR is not going to be supported.

Instead, the `asadmin`, `asadmin.bat`, or `asadmin.java` scripts should be used to execute admin commands.

PR #25699, opened 2025-09-11
Fixed asadmin get -m for monitoring

### LINKED ISSUE #25695: Fix asadmin monitoring commands

All versions, tested 7.0.25 and master branch

### JDK version

17

### OS

Tuxedo

### Database

Derby, Postgress

### Problem Description

The following sequence should result in printing a list of monitored values and their descriptions (in a format like usual get command). But it doesn't do anything, HOWEVER ... under some circumstances it starts working, but I could not find out what enables that.
The Admin GUI shows those data, but asadmin doesn't.

### Steps to reproduce

```
asadmin set configs.config.server-config.monitoring-service.module-monitoring-levels.jdbc-connection-pool=HIGH
asadmin set configs.config.server-config.monitoring-service.module-monitoring-levels.jvm=HIGH
asadmin start-database
asadmin ping-connection-pool DerbyPool
asadmin ping-connection-pool __TimerPool
asadmin --terse get -m 'server.resources.*'
asadmin --terse get -m '*'
```

### Impact of Issue

_No response_

dmatej:
- Fixes #25695 
- The plain reporter's message was lost finally; removal resolved the problem.
- added generics
- removed unused fields
- reduced visibility of fields
- I am using monitoring in another branch, so this became a blocker for me
- 7.0.25 works (mostly)

### DIFF
--- a/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/MonitoringReporter.java
+++ b/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/MonitoringReporter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024 Contributors to the Eclipse Foundation.
+ * Copyright (c) 2024, 2025 Contributors to the Eclipse Foundation.
  * Copyright (c) 1997, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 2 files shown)


nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/MonitoringReporter.java
```diff
@@ -183,49 +187,37 @@ public void execute() {
...
     private void runAggregate() {
         List<String> list = getOutputLines();
-        ActionReport aggregateReporter = null;
+        final ActionReport aggregateReporter;
         if (aggregateDataOnly) {
-            plainReporter = new PlainTextActionReporter();
-            aggregateReporter = plainReporter.addSubActionsReport();
+            aggregateReporter = reporter.addSubActionsReport();
         } else {
             aggregateReporter = reporter.addSubActionsReport();
```
avpinchuk: Why here the `if else` statement? Both conditions do the same assignment.

dmatej: Because originally I was trying to work on something else, thanks for noticing that!

avpinchuk (approved):

PR #25703, opened 2025-09-16
Expand support for MicroProfile APIs to GlassFish Web Profile and Embedded

### LINKED ISSUE #25677: Support MicroProfile APIs in WebProfile and Embedded GlassFish

8.0

### JDK version

N/A

### OS

N/A

### Database

_No response_

### Problem Description

Currently, only GlassFish Full Platform distribution supports MicroProfile. On GlassFish 7, it was supported only when running on Java 17+. 

Since all the MicroProfile implementations are pretty slim and GlassFish 8 requires Java 21+, nothing prevents GlassFish Web Profile and both Embedded GlassFish distributions to support MicroProfile too.

### Steps to reproduce

Deploy an app with MicroProfile config (e.g. it calls `ConfigProvider.getConfig()` on startup) on GlassFish Web Profile or on Embedded GlassFish Full or Web Profile - deployment fails with ConfigProvider class not found.

### Impact of Issue

MicroProfile not available in most of the GlassFish distributions.

OndroMih:
Fixes https://github.com/eclipse-ee4j/glassfish/issues/25677.

Until now, MicroProfile APIs are only supported in GlassFish Server Full Platform. MicroProfile support requires Java 17. Since GlassFish version 7.1.0 requires Java 17, MicroProfile can become supported by all distributions.

In this PR:

* MicroProfile Config, JWT, Rest Client support is added to GlassFish Web Profile and Embedded GlassFish All and Web
* Conditions related to Java 11/17 are removed, all runs all the time with MicroProfile, on Java 17+
* Restructured the build of Embedded distributions, so that they are based on feature sets, similarly to server distributions - this works a bit differently. In Embedded distributions, feature sets are just Maven dependencies, and all transitive dependencies are included. That's why I added excludes into the feature sets. In Server, feature sets are specially handled by the build-helper plugin, which pulls in only direct dependencies of feature sets and ignores transitive dependencies
* Embedded All is a superset of Embedded Web
* Because Embedded Web contained some artifacts that were missing in Embedded All, these artifacts are now removed (related to clustering)
* I also added Concurrency API to Embedded Web, which was missing there

I compared the contents of the final artifacts before and after this change:

Embedded All contains all files from before, plus adds MicroProfile-related ones:

* MicroProfile APIs
* MicroProfile implementations (Jersey Rest Client + Jersey Proxy Client, Helidon, OmniFaces JWT, GlassFish Health)
* Reactive streams API - `org.reactivestreams` (required by Jersey Rest Client)
* META-INF/placeholder.txt - I don't know where from, but it's harmless
* META-INF/services/jakarta.ws.rs.container.DynamicFeature - this is for JWT integration

Embedded Web contains the same new files as Embedded All, plus excludes all that are not in Embedded All:

* jsch (SSH client library)
* cluster-cli
* cluster-common
* cluster-admin
* cluster-ssh
* gms-adapter

I ran the Embedded tests against Embedded All, and all of them passed. I ran them also against Embedded Web, but they failed because of an issue that wasn't introduced by these changes - Web tries to use a class from All, related to IIOP, which isn't present in Web, resulting in NoClassDefFoundError. It should be easy to refactor but I'm not going to do that in this PR.


dmatej:
* distribution zips are same.
* embedded all - changed staged directory content, however the artifact content doesn't seem missing anything.

So I am giving a green light despite I am not a huge fan of feature sets.

dmatej (approved):

avpinchuk:
We should either inline dependencies (bad if user will use different versions of the inlined libraries in his application) or inline repackaged versions of the dependencies.

dmatej:
It would be good to add at least one simple test for that ...

OndroMih:
I added a simple test in https://github.com/eclipse-ee4j/glassfish/pull/25708

PR #25708, opened 2025-09-17
Fix MicroProfile Config in Embedded - add transitive dependencies

OndroMih:
Embedded distributions need all transitive dependencies, while server only takes direct dependencies, so no need to exclude anything for the server.

Some Helidon dependencies are packaged as embedded JARs inside an OSGi bundle. In Embedded they are needed as transitive dependencies.

I tested with a simple app that uses MP Config and it worked with Embedded GlassFish All.

### DIFF
--- a/appserver/tests/embedded/microprofile/config/src/test/java/org/glassfish/tests/embedded/microprofile/config/MicroProfileConfigTest.java
+++ b/appserver/tests/embedded/microprofile/config/src/test/java/org/glassfish/tests/embedded/microprofile/config/MicroProfileConfigTest.java
@@ -16,8 +16,6 @@
 
 package org.glassfish.tests.embedded.microprofile.config;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.InputStreamReader;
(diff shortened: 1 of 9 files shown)


avpinchuk:
Looks good but why so many ant tests failed? Also found 1 checkstyle violation.

appserver/tests/embedded/microprofile/config/src/test/java/org/glassfish/tests/embedded/microprofile/config/MicroProfileConfigTest.java
```diff
@@ -0,0 +1,86 @@
...
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+
+package org.glassfish.tests.embedded.microprofile.config;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
avpinchuk: Must be after all non-static imports

avpinchuk (approved):

hs536 (approved):

PR #25714, opened 2025-09-23
Created asadmin.java and nadmin.java

dmatej:
### The Story Behind This

@arjantijms recently noticed an execution of `Runtime.getRuntime().exec(args, new String[0])` which unfortunately is quite problematic, because then it creates new process without any environment options. See javadoc:

>      * @param   envp      array of strings, each element of which
>      *                    has environment variable settings in the format
>      *                    <i>name</i>=<i>value</i>, or
>      *                    {@code null} if the subprocess should inherit
>      *                    the environment of the current process.

That was not a big problem when it executed directly the java executable with absolute path, but when you need to execute script resolving environment options to get the java executable, ie. from JAVA_HOME or PATH ... 

I remembered that since some Java version it is possible to execute *.java files even without compilation just like usual scripts ... then if we would have asadmin.java ... we don't need to resolve any JAVA_HOME, because then we would already have it.

Even better, in the past we had issues with the ignorance of JAVA_HOME, so scripts then used default java command found in `/usr/bin` or `/usr/local/bin`, sometimes even incompatible old version ... this resolves that!

It makes easier also usage on Windows ... not tested yet, but if I did not make any mistake, GitHub Actions should confirm at least that it works ...

### Usage

Practically same as with original asadmin/nadmin commands, bat or bash based. With few traps.

* Java is same you used to run the asadmin.java; no env options JAVA_HOME or AS_JAVA will change it.
* env option `AS_USE_NATIVE_SCRIPT=true` will cause execution of the original bat/bash script instead of using Java directly. It was my original idea, however why we would add another complicated layer when we can do that directly? However it might be useful sometimes. 
* env option `AS_INHERIT_ENVIRONMENT=false` disables providing env options to the process, except JAVA_HOME and AS_JAVA
* env option `AS_TRACE=true` enables some limited logging to STDERR.
* The `asadmin.java get` command doesn't need `"*"` to avoid bash/bat resolution; however on command line you will be again in bash/bat, so you have to do that. Executed from JUnit tests directly you don't. It is confusing sometimes, but it makes sense.

```
/usr/lib/jvm/jdk21/bin/java ./glassfish7/bin/asadmin.java start-domain --debug
```

Slight warning about Windows - just like Linux it interpretes asterisks, however it does it even when executed from Java. It is a good practice to use `""` even for `-Dx=y` system options, otherwise it complains and screams. 

For example with Maven:
```
PS C:\Users\vboxuser\Documents\glassfish-gama> C:\Users\vboxuser\maven\bin\mvn.cmd clean install -pl :admin-tests -Dit.test=JobManagerITest
...
[ERROR] Unknown lifecycle phase ".test=JobManagerITest". You must specify a valid lifecycle phase or a goal in the format
```
That is probably the cause why some GlassFish scripts call jar to generate command to execute another jar. I mean things around appclient.


appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/asadmin/Asadmin.java
```diff
@@ -168,7 +174,7 @@ public <T> KeyAndValue<T> getValue(final String key, final Function<String, T> t
     public <T> List<KeyAndValue<T>> get(final String key, final Function<String, T> transformer) {
         AsadminResult result = exec("get", key);
         assertThat(result, asadminOK());
-        return Arrays.stream(result.getStdOut().split(System.lineSeparator())).map(KEYVAL_SPLITTER)
+        return Arrays.stream(result.getStdOut().split("[\n\r]+")).map(KEYVAL_SPLITTER)
```
dmatej, line 177: Surprising bug found, when I used asadmin.java, it failed on windows server 2025 in PowerShell and same on GitHub Actions, because System.lineSeparator() was not enough to split these lines.

appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/asadmin/Asadmin.java
```diff
@@ -238,6 +244,16 @@ private AsadminResult exec(final Integer timeout, final boolean detachedAndTerse
         final List<String> parameters = Arrays.asList(args);
         LOG.log(TRACE, "exec(timeout={0}, detached={1}, args={2})", timeout, detachedAndTerse, parameters);
         final List<String> command = new ArrayList<>();
+        if (asadmin.getName().endsWith(".java")) {
```
dmatej, line 247: If this would cause some problems in the future, we can remove it, however it limits memory used by asadmin.java as much as possible. It also disables GC, because this command has really short life in tests. It might not on command line as it supports --verbose and --multimode too, etc.

appserver/tests/admin/tests/src/test/java/org/glassfish/main/admin/test/progress/ProgressStatusFailITest.java
```diff
@@ -61,7 +61,7 @@ public void failDuringExecution() {
      */
     @Test
     public void timeout() {
-        AsadminResult result = ASADMIN.exec(2_000, "progress-custom", "1x1", "1x2", "1x1");
+        AsadminResult result = ASADMIN.exec(3000, "progress-custom", "1x1", "1x2", "1x1");
```
dmatej, line 64: Lower timeout sometimes caused that the task did not make it to write the "33%" line.

appserver/tests/admin/tests/src/test/java/org/glassfish/main/admin/test/PathWithSpacesITest.java
```diff
@@ -19,44 +19,67 @@
...
+import org.junit.jupiter.api.condition.OS;
 
 import static org.glassfish.main.itest.tools.asadmin.AsadminResultMatcher.asadminOK;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class PathWithSpacesITest {
 
-    private static Asadmin asadmin;
+    private static File binDirectory;
```
dmatej, line 38: Now we cover all three asadmins. startserv script is tested on another place.

appserver/tests/admin/tests/src/test/resources/META-INF/services/org.junit.jupiter.api.extension.Extension
```diff
@@ -0,0 +1 @@
+org.glassfish.tests.utils.junit.TestLoggingExtension
```
dmatej, line 1: Prints timing of tests, makes my life easier when I am searching in logs.
It also avoids situation that developer forgets to add the `@ExtendWith` annotation on new test.

nucleus/admin/cli/src/main/java/com/sun/enterprise/admin/cli/remote/LocalStrings.properties
```diff
@@ -33,11 +33,7 @@ ServerIsNotSecure=It appears that server [{0}:{1}] does not accept secure connec
 fileTransferTargetNotDir=File transfer target {0} is expected to be a directory but is not
 errSavingContent=Error saving returned content {0}: {1}
 unknownOption=Command {0} does not have an option {1}
-notEnoughOperands=Command {0} requires an operand
```
dmatej, line 36: Some of these properties were copy pasted without usage on multiple places. And when it broke, I could not find out why - what exactly it really received.

nucleus/common/common-util/src/main/java/com/sun/enterprise/universal/process/ProcessManager.java
```diff
@@ -334,7 +334,10 @@ public void run() {
                     }
                 }
             } catch (IOException e) {
-                LOG.log(WARNING, () -> "ReaderThread " + getName() + " cannot read the process output.", e);
+                if (!isInterrupted()) {
```
dmatej, line 337: There was and is a race condition - the process can end and the input can be closed then before we managed to read everything. Or we even managed to read everything, but we are still trying to fill some internal buffer, we receive IOException "stream closed" then. There is no way to synchronize with another process, we just have to manage it. And as we know if we were interrupted (that is why the stream was closed) ... so we can just skip the worthless warning and stacktrace. Maybe this description is not exact, my brain is still bit confused of all relevant javadocs 🤯

nucleus/distributions/nucleus-common/src/main/resources/bin/nadmin.java
```diff
@@ -0,0 +1,162 @@
+/*
```
dmatej, line 1: These three "java       scripts" are nearly copy pasted, but they have no dependencies and are compiled on execution. It has advantages and disadvantages.

```
- Needs more memory needed than with bash/bat
- Slower than bash/bat 
  - controversial as it doesn't parse asenv and config.sh
  + jar files executed by that are much heavier problem, seconds vs milliseconds
+ OS independent
```

appserver/distributions/glassfish-common/src/main/resources/bin/asadmin.java
```diff
@@ -0,0 +1,164 @@
...
+
+    private static final String SYS_LAUNCHERFILE = "jdk.launcher.sourcefile";
+
+    private static final boolean AS_TRACE = "true".equals(System.getenv("AS_TRACE"));
+    private static final boolean INHERIT_ENV = "false".equals(System.getenv("AS_INHERIT_ENVIRONMENT"));
+    private static final boolean USE_SCRIPT = "true".equals(System.getenv("AS_USE_NATIVE_SCRIPT"));
+    private static final boolean IS_WINDOWS = System.getProperty("os.name").toLowerCase().contains("windows");
+
+    private static final String JAVA_HOME = System.getProperty("java.home");
+    private static final Path MY_JAVA_FILE = getMyJavaFile();
```
avpinchuk: The `NullPointerException` is thrown when `SYS_LAUNCHERFILE` property is not set (see L:95).

appserver/distributions/glassfish-common/src/main/resources/glassfish/bin/asadmin.java
```diff
@@ -0,0 +1,164 @@
...
+
+    private static final String SYS_LAUNCHERFILE = "jdk.launcher.sourcefile";
+
+    private static final boolean AS_TRACE = "true".equals(System.getenv("AS_TRACE"));
+    private static final boolean INHERIT_ENV = "false".equals(System.getenv("AS_INHERIT_ENVIRONMENT"));
+    private static final boolean USE_SCRIPT = "true".equals(System.getenv("AS_USE_NATIVE_SCRIPT"));
+    private static final boolean IS_WINDOWS = System.getProperty("os.name").toLowerCase().contains("windows");
+
+    private static final String JAVA_HOME = System.getProperty("java.home");
+    private static final Path MY_JAVA_FILE = getMyJavaFile();
```
avpinchuk: The `NullPointerException` is thrown when `SYS_LAUNCHERFILE` property is not set.

nucleus/distributions/nucleus-common/src/main/resources/bin/nadmin.java
```diff
@@ -0,0 +1,162 @@
...
+
+    private static final String SYS_LAUNCHERFILE = "jdk.launcher.sourcefile";
+
+    private static final boolean AS_TRACE = "true".equals(System.getenv("AS_TRACE"));
+    private static final boolean INHERIT_ENV = "false".equals(System.getenv("AS_INHERIT_ENVIRONMENT"));
+    private static final boolean USE_SCRIPT = "true".equals(System.getenv("AS_USE_NATIVE_SCRIPT"));
+    private static final boolean IS_WINDOWS = System.getProperty("os.name").toLowerCase().contains("windows");
+
+    private static final String JAVA_HOME = System.getProperty("java.home");
+    private static final Path MY_JAVA_FILE = getMyJavaFile();
```
avpinchuk: The `NullPointerException` is thrown when `SYS_LAUNCHERFILE` property is not set.

appserver/distributions/glassfish-common/src/main/resources/bin/asadmin.java
dmatej: How? There is System.exit call if it is null

avpinchuk:
```java
System.out.println("The '" + SYS_LAUNCHERFILE
                + "' property is not set, you probably did not execute this program running java "
                + MY_JAVA_FILE.getFileName() + ", right?");
 System.exit(1);
```

In the `System.out.println()` `MY_JAVA_FILE` is not initialized yet.

appserver/itest-tools/src/test/java/org/glassfish/main/admin/test/tool/StartStopITest.java
```diff
@@ -34,16 +35,19 @@
...
+        assertThat(asadmin.exec("get", "*"), asadminOK());
     }
 
 
     @Test
     @EnabledOnOs(OS.WINDOWS)
     void asadminGet_Windows() {
-        assertThat(getAsadmin().exec("get", "\"*\""), asadminOK());
+        Asadmin asadmin = getAsadmin().withEnv("AS_TRACE", "true");
+        assertThat(asadmin.exec("get", "*"), asadminOK());
```
avpinchuk, line 39: Methods `asadminGet_Linux` and `asadminGet_Windows` now do the same thing, One of them is redundant.

appserver/distributions/glassfish-common/src/main/resources/bin/asadmin.java
dmatej: Fixed, I made some refactoring to make it more visible next time.

avpinchuk (approved):

PR #25716, opened 2025-09-29
Fix #25715 Enabling monitoring for JDBC connection pools unexpectedly enables monitoring for non-JDBC connection pools

tnagao7:
Fixes #25715.

For any connection pools (including JDBC connection pools and connector connection pools), the `registerPool` methods of the `ConnectionPoolMonitoringExtension` interface implementations are executed. 
The issue occurs because the `registerPool` method of the `JdbcPoolMonitoringExtension` class (implementing `ConnectionPoolMonitoringExtension`) is executed and handles not only JDBC connection pools but also connector connection pools.

To resolve the issue, we can modify the `registerPool` method of the `JdbcPoolMonitoringExtension` class to handle only JDBC connection pools, as in the [`registerConnectionPool`](https://github.com/eclipse-ee4j/glassfish/blob/4870d6def4c11bdc27451b40b71f32083caded1d/appserver/jdbc/jdbc-runtime/src/main/java/org/glassfish/jdbcruntime/JdbcPoolMonitoringExtension.java#L138) method.

### DIFF
--- a/appserver/jdbc/jdbc-runtime/src/main/java/org/glassfish/jdbcruntime/JdbcPoolMonitoringExtension.java
+++ b/appserver/jdbc/jdbc-runtime/src/main/java/org/glassfish/jdbcruntime/JdbcPoolMonitoringExtension.java
@@ -85,7 +85,8 @@ public class JdbcPoolMonitoringExtension implements ConnectionPoolMonitoringExte
      */
     @Override
     public void registerPool(PoolInfo poolInfo) {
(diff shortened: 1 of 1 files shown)


dmatej (approved):
Hmm, you are right, it implements the `ConnectionPoolMonitoringExtension` and the only caller of the method I see is `ConnectionPoolStatsProviderBootstrap` which does that for all pools.

PR #25724, opened 2025-10-06
Embedded GlassFish - fixes related to system properties

OndroMih:
System properties can be set by the following means:

1. on command line, with -D Java argument - overrides properties set in the configuration
2. via create-system-properties admin command (spaces in property names or values must be escaped by '\')
3. via setting the property directly in the `--propeties` file (e.g. property "my.name" should have line "my.name=GlassFish Server" in the properties file, spaces shouldn't be escaped)
4. via system-property element in domain.xml

Fixes number 3, which worked in 7.0.25 but stopped working in master.
Allows empty spaces in property values if setting via the create-system-properties admin command - space escaped by `\` is treated as value not as a separator. This follows the same pattern as in the asadmin CLI tool.

### DIFF
--- a/nucleus/common/simple-glassfish-api/src/main/java/org/glassfish/embeddable/GlassFishVariable.java
+++ b/nucleus/common/simple-glassfish-api/src/main/java/org/glassfish/embeddable/GlassFishVariable.java
@@ -18,7 +18,6 @@ package org.glassfish.embeddable;
 
 import java.util.Arrays;
 import java.util.Map;
-import java.util.Optional;
 import java.util.stream.Collectors;
 
 /**
@@ -127,10 +126,6 @@ public enum GlassFishVariable {
      */
     public String getSystemPropertyName() {
         return this.sysPropName;
-    }
-
-    public Optional<String> getSystemPropertyValue() {
-        return Optional.ofNullable(System.getProperty(getSystemPropertyName()));
     }
 
     /**
--- a/nucleus/common/simple-glassfish-api/src/main/java/org/glassfish/embeddable/GlassFishVariable.java
+++ b/nucleus/common/simple-glassfish-api/src/main/java/org/glassfish/embeddable/GlassFishVariable.java
@@ -18,6 +18,7 @@ package org.glassfish.embeddable;
 
 import java.util.Arrays;
 import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 /**
@@ -128,6 +129,9 @@ public enum GlassFishVariable {
         return this.sysPropName;
     }
 
+    public Optional<String> getSystemPropertyValue() {
+        return Optional.ofNullable(System.getProperty(getSystemPropertyName()));
+    }
 
     /**
      * The map contains pairs of {@link #getEnvName()} and {@link #getSystemPropertyName()}.
--- a/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/SystemTasksImpl.java
+++ b/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/SystemTasksImpl.java
@@ -135,7 +135,7 @@ public class SystemTasksImpl implements SystemTasks, PostConstruct {
     private void setSystemPropertiesFromEnv(boolean forceOverwrite) {
         // adding our version of some system properties.
         setProperty(JAVA_ROOT.getSystemPropertyName(), System.getProperty(JAVA_HOME.getSystemPropertyName()), forceOverwrite);
-        if (HOST_NAME.getSystemPropertyValue().isEmpty()) {
+        if (forceOverwrite || null == System.getProperty(HOST_NAME.getSystemPropertyName())) {
             String hostname = "localhost";
             try {
                 // canonical name checks to make sure host is proper
@@ -144,7 +144,7 @@ public class SystemTasksImpl implements SystemTasks, PostConstruct {
                 LOG.log(Level.SEVERE, KernelLoggerInfo.exceptionHostname, ex);
             }
             if (hostname != null) {
-                setProperty(HOST_NAME.getSystemPropertyName(), hostname, false);
+                setProperty(HOST_NAME.getSystemPropertyName(), hostname, true);
             }
         }
     }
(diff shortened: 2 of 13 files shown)


nucleus/common/simple-glassfish-api/src/main/java/org/glassfish/embeddable/GlassFishVariable.java
```diff
@@ -128,6 +129,9 @@ public String getSystemPropertyName() {
         return this.sysPropName;
     }
 
+    public Optional<String> getSystemPropertyValue() {
```
dmatej: 1. This lib should be without tech debt and with good design. (EDIT: originally I thought it is in jdke, but it applies to the simple api too)
2. Missing javadoc
3. When I created `GlassFishVariable`, I had doubts if I should create also `getValue` methods here, however I have found that it is a not good idea, because they don't save me of any pain. In your case, `-Dcom.sun.aas.hostName=""` (empty string) is not an empty optional. Also, sometimes I need a supplier, sometimes it would be better to use just optional, sometimes, I need some processing ... and soon it would be messy as usual.
4. Exactly opposite, evaluation of these variables are always somehow specific to some domain. There it belongs to. This class serves just to replace old set of hashmaps and constants used on older version, and give some order to the mapping of env option and system properties.

So I would move it to the SystemTasksImpl and maybe we could even improve the validation. Actually I am now looking on other usages, one other usages outside this PR contains a bug, another has completely different evaluation of the hostname, maybe depending on this code, not sure ... we should clean it up.

However I did not put these methods to GlassFishVariable intentionally.

nucleus/core/kernel/src/main/java/org/glassfish/runnablejar/UberMain.java
```diff
@@ -177,14 +177,15 @@ private void runCommandPromptLoop() throws GlassFishException {
...
+        // Split according to empty space but not if empty space is escaped by \
+        String[] split = stringCommand.split("(?<!\\\\)\\s+");
         String command = split[0].trim();
         String[] commandParams = null;
         if (split.length > 1) {
             commandParams = new String[split.length - 1];
             for (int i = 1; i < split.length; i++) {
                 commandParams[i - 1] = split[i].trim();
-            }
+        }
```
dmatej: Typo

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/SystemTasksImpl.java
```diff
@@ -128,22 +132,24 @@ private void setVersion() {
...
+        if (HOST_NAME.getSystemPropertyValue().isEmpty()) {
+            String hostname = "localhost";
+            try {
+                // canonical name checks to make sure host is proper
+                hostname = NetUtils.getCanonicalHostName();
+            } catch (Exception ex) {
+                LOG.log(Level.SEVERE, KernelLoggerInfo.exceptionHostname, ex);
+            }
+            if (hostname != null) {
+                setProperty(HOST_NAME.getSystemPropertyName(), hostname, false);
```
dmatej: I think we both have a very good question ... why to ignore the `forceOverwrite` here.

nucleus/common/simple-glassfish-api/src/main/java/org/glassfish/embeddable/GlassFishVariable.java
dmatej: Another idea: just remove this and use just nullcheck in SystemTasksImpl and we will target the hostname evaluation in another PR.

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/SystemTasksImpl.java
OndroMih: OK, I'll change it to follow forceOverwrite also here.

nucleus/common/simple-glassfish-api/src/main/java/org/glassfish/embeddable/GlassFishVariable.java
OndroMih: OK. I'll just inline this methog SystemTasksImpl and remove it from GlassFishVariable. I added it to GlassFish variable because it makes the code cleaner in SystemTasksImpl, `HOST_NAME.getSystemPropertyValue().isEmpty()` instead of `System.getProperty(HOST_NAME.getPropertyName()) != null)`, but it's not much different and still readable.

OndroMih: In fact, if I change the code to respect `forceOverwrite`, then I don't even need this at all :)

OndroMih: I might need it to skip getting hostname if the property is already set and forceOverwrite is false.

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/SystemTasksImpl.java
OndroMih: I changed it so that it uses only attempts to compute hostname and set the value if either forceOverwrite=true or if the value is not set. To avoid computing hostname if it's going to be ignored in the end. A very small optimization for an edge case, I know :)

nucleus/common/simple-glassfish-api/src/main/java/org/glassfish/embeddable/GlassFishVariable.java
dmatej: I still don't know what to do if the property value is an empty string. Is it the same as an empty string (explicit reset)? Or it can be taken as an error. Maybe the first option is correct, then the optional doesn't work. As a paradox Optional can make it [more complicated](https://stackoverflow.com/questions/28322464/in-java-8-transform-optionalstring-of-an-empty-string-in-optional-empty) compared to old and ugly "null or empty" check.

OndroMih: I removed this method already

OndroMih: And I only check for null, not empty. Empty means a value, the only reason why it would mean the property is not set would be if we want to allow deleting the property by setting it to an empty string. Makes sense in some cases, but I wouldn't bother here.

dmatej: Yeah, but empty hostname doesn't sound something valid for me.

dmatej (approved):

avpinchuk (approved):

arjantijms (approved):

PR #25728, opened 2025-10-08
Fix bug where policy context was not cleared before policy retranslate

arjantijms:
When a servlet or filter is added dynamically, internally GlassFish triggers a Policy retranslate. This means indexing of URLs for permissions etc is done again.

Jersey happens to do such dynamic add, and thus triggers this retranslation.

We had a bug there, where the previous context was not properly cleared. This particularly meant that the "http request capturing lambda" was still the one from the first translation. After the retranslation, the lambda of the second context was given the http request, but the first one was still being used when application code asked for it. The first one was never given the request, so would always return null.

The request is used currently for the SecurityContext implementation of Soteria (although we are bound to change this in a later version). The included test therefor also looks at this stored request directly.


appserver/tests/application-tests/security/policy-retranslate/pom.xml
```diff
@@ -0,0 +1,137 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.shrinkwrap</groupId>
+            <artifactId>shrinkwrap-impl-base</artifactId>
+        </dependency>
+
+        <!--
+            The Arquillian connector that starts GlassFish and deploys archives to it.
+        -->
+        <dependency>
```
dmatej, line 112: This creates cyclic dependency. It depends on GlassFish project, now Glassfish project depends on the container too. It seems to me that there is yet bit more redundant things.
Can I merge it to application-tests instead? We also don't need to have new pom (or two) for every test. 

EDIT: Ok, we already have the dependency in TCK submodules ... I forgot about this aspect, however as the plugin can support multiple GF versions, it is probably ok.
However I would still move it to application-tests.

appserver/tests/security/policy-retranslate/pom.xml
```diff
@@ -0,0 +1,137 @@
...
+        
+        This action triggers policy retranslate. In the REST resource we use get caller principal
+        from the security context, which on its turn depends upon the request set by authorization.
+        
+        If the retranslate setup doesn't clear the previous translation, we will not get to see
+        this request and hence not the principal.
+    </description>
+
+    <dependencies>
+        <dependency>
```
dmatej: I prefer to explicitly declare which specs we need here ...

appserver/tests/security/policy-retranslate/src/main/webapp/WEB-INF/beans.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<beans xmlns="https://jakarta.ee/xml/ns/jakartaee"
```
dmatej: Do we still need the beans.xml here?

appserver/tests/security/policy-retranslate/pom.xml
arjantijms: I took this particular app from another demo app I was working on, where just the EE dependency was more convenient. For here I can specify the exact APIs used indeed.

appserver/tests/application-tests/security/policy-retranslate/pom.xml
arjantijms: > Ok, we already have the dependency in TCK submodules 

yeah, it's modelled exactly after the TCK submodule. We don't have application-tests btw, but I think it's appserv-tests you mean then?

appserver/tests/security/policy-retranslate/src/main/webapp/WEB-INF/beans.xml
arjantijms: Nope, we indeed don't need it. Left-over from the app I took it from.

appserver/tests/application-tests/security/policy-retranslate/pom.xml
dmatej: We do for some 3 years ;-)
https://github.com/eclipse-ee4j/glassfish/tree/master/appserver/tests/application

PR #25733, opened 2025-10-10
Merge from master

dmatej:
* Multimode and autocomplete
* Disabled staging in Jenkinsfile
* Fixes
* Updates
* Depends on https://github.com/eclipse-ee4j/glassfish-concurro/pull/148 - instead of release, for now we can temporarily depend on snapshots until Concurro deploys to Maven Central.

Have to be released:
* Soteria
  * https://github.com/eclipse-ee4j/soteria/pull/401
* Derby repackaged
* Concurro
* `org.glassfish.external:schema2beans:jar:6.7` ... new needs Antlr4, and problem is just with the `go-offline`, so we can keep it yet for a while.
* Ant - we can stay with the old one, but once I release Derby, I can release this one too.
  * EDIT: We must stay:
    ```
    org.glassfish.main.persistence.cmp.support-ejb [48]
    missing requirement
        &(package = org.netbeans.modules.schema2beans) (password = GlassFish) (version >= 6.5.0) (!(version >= 7.0.0))
        caused by:
            Unable to resolve
                org.glassfish.external.schema2beans [88]
                missing requirement
                    package = org.apache.tools.ant
                    caused by:
                        Unable to resolve
                            org.glassfish.external.ant [303]
                            missing requirement
                                package = com.sun.tools.javah)]]
    ```
* All snapshots were temporarily deployed to verify that this PR passes on Jenkins.

### DIFF
--- a/appserver/pom.xml
+++ b/appserver/pom.xml
@@ -194,7 +194,8 @@
         <!-- Other -->
         <dbschema.version>2.7.0-SNAPSHOT</dbschema.version>
         <schema2beans.version>2.7.0-SNAPSHOT</schema2beans.version>
-        <derby.version>10.17.1.0-SNAPSHOT</derby.version>
+        <derby.version>10.17.1.0</derby.version>
+        <derby.repackaged.version>10.17.1.0-SNAPSHOT</derby.repackaged.version>
         <wsdl4j.version>1.6.3</wsdl4j.version>
         <maven-rar-plugin.version>3.0.0</maven-rar-plugin.version>
 
@@ -700,7 +701,7 @@
             <dependency>
                 <groupId>org.glassfish.external</groupId>
                 <artifactId>derby</artifactId>
-                <version>${derby.version}</version>
+                <version>${derby.repackaged.version}</version>
                 <type>zip</type>
             </dependency>
             <dependency>
--- a/appserver/pom.xml
+++ b/appserver/pom.xml
@@ -98,6 +98,9 @@
         <jakarta.concurrent-api.version>3.1.1</jakarta.concurrent-api.version>
         <concurro.version>3.1.0-M6</concurro.version>
 
+        <!-- Jakarta Interceptors -->
+        <jakarta.interceptor-api.version>2.2.0</jakarta.interceptor-api.version>
+
         <!-- Jakarta Security + Authentication/Authorization -->
         <!-- APIs -->
         <jakarta.security-api.version>4.0.0</jakarta.security-api.version>
@@ -107,7 +110,6 @@
         <soteria.version>4.0.1</soteria.version>
         <exousia.version>3.0.0</exousia.version>
         <epicyro.version>3.1.0</epicyro.version>
-        <!-- Dependencies -->
         <nimbus.version>10.5</nimbus.version>
         <jcip.version>1.0.2</jcip.version>
 
@@ -159,9 +161,6 @@
         <jakarta.cdi-api.version>4.1.0</jakarta.cdi-api.version>
         <weld.version>6.0.3.Final</weld.version>
         <jboss.classfilewriter.version>1.3.1.Final</jboss.classfilewriter.version>
-
-        <!-- Jakarta Interceptors -->
-        <jakarta.interceptor-api.version>2.2.0</jakarta.interceptor-api.version>
 
         <!-- Jakarta MVC -->
         <jakarta.mvc-api.version>3.0.0</jakarta.mvc-api.version>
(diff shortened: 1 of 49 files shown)


appserver/pom.xml
```diff
@@ -106,8 +109,7 @@
         <!-- Implementations -->
         <soteria.version>4.0.1</soteria.version>
         <exousia.version>3.0.0</exousia.version>
-        <epicyro.version>3.1.0</epicyro.version>
-        <!-- Dependencies -->
+        <epicyro.version>3.1.1</epicyro.version>
```
pzygielo, line 112: Where I could find this version? In Central there is only 3.0.0 and 3.1.1 is not even [staged](https://jakarta.oss.sonatype.org/content/groups/staging/org/glassfish/epicyro/epicyro/) (which would be also problematic for the release).

pzygielo: It's taking its time, but 3.1.1 appears slowly in central now.

pzygielo: Funny - that's what I see in browser:

Image

And that's `maven-metadata.xml`:
```xml
<metadata>
  <groupId>org.glassfish.epicyro</groupId>
  <artifactId>epicyro</artifactId>
  <versioning>
    <latest>3.1.1</latest>
    <release>3.1.1</release>
    <versions>
      <version>3.0.0</version>
      <version>3.1.1</version>
    </versions>
    <lastUpdated>20251008222333</lastUpdated>
  </versioning>
</metadata>
```
(few days old, with 3.1.1 already there).

dmatej: https://mvnrepository.com/artifact/org.glassfish.epicyro/epicyro/3.1.1

dmatej: Staging is dead, and it is part of the problem. Now fails certificate validation of staging for some reason, it seems. Some artifacts were just staged and nobody pushed them out, next item is Concurro.
However Epicyro 3.1.1 is in Maven Central now. I am writing an email to dev mailing list ... it may take a while, I want to collect somehow if there are yet some "holes".
For Epicyro 3.1.0 @arjantijms  created this issue, but it is not moving at all: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/6729

dmatej: Yeah, I noticed that too, first it makes artifacts downloadable including the metadata file, however the UI index is updated much later. When I tried last time, it was MAYBE after I tried to use search form on Maven Central, visited the page, and it maybe notified something to update the index. Or it was just a coincidence.

appserver/pom.xml
```diff
@@ -96,7 +96,10 @@
 
         <!-- Jakarta Concurrency -->
         <jakarta.concurrent-api.version>3.1.1</jakarta.concurrent-api.version>
-        <concurro.version>3.1.0-M6</concurro.version>
+        <concurro.version>3.1.0-SNAPSHOT</concurro.version>
```
pzygielo, line 99: Only a bit worse than using staging repo. IMO - such change (using SNAPSHOT) should never land in public repository.

dmatej: It can land to Maven Central Snapshots, that is the purpose. It is not even worse than using staging, because I had to delete the maven cache on CI manually whenever some dependency was rereleased to the the staging. Sometimes we even did not notice it, it made it to Central and I think you was usually the first person noticing that hashes differ. CycloneDX is now allergic to these differences if I remember it correctly.

So this unblocks daily deployments to snapshots, but continues to block releases. Also maven release plugin would scream and shout if we would have snapshot dependencies in the release, that worked already in Maven 2 ... 

What I like about this is that it is pretty clear and visible.

pzygielo: By "repository" I meant "source code git repository" as that is the place where "such change (using SNAPSHOT)" is offered in this PR.

dmatej: It is a compromise, better than blocked development. The only problem is rebuilding the commit in the future, but that is usually not hard to resolve by using already released artifact instead.

For now we need to prove at least that the whole system is ok and then we will release all these snapshots and update the PR again. Or we can merge it as it is (if it passes checks), but at this point Arjan cannot work on 8.0 at all as staging is somehow broken and there is also a risk that it is not all we have found yet.

Again, previous system with staging which allowed redeployments was even worse because what was a final dependency 1.2.3 as of 31.8.2023 was sometimes completely different jar file than 1.2.3 as of 1.1.2025. With snapshots everyone can see that and has to respect that it is not stable dependency.

pzygielo: > update the PR again.

:+1: 

> what was a final dependency 1.2.3 as of 31.8.2023 was sometimes completely different jar file than 1.2.3 as of 1.1.2025. With snapshots everyone can see that and has to respect that it is not stable dependency.

Right, indeed. 

But - IMO using unreleased artifacts (staged or SNAPSHOTs) is justified only when done locally.

arjantijms (approved):

PR #25738, opened 2025-10-19
Merge from master

dmatej:
Following https://github.com/eclipse-ee4j/glassfish/pull/25733

* Fixed featureset web pom.xml
* Merge form master to 8.0


avpinchuk (approved):

OndroMih (approved):

dmatej:
Next: I am releasing external dependencies so we would not use snapshots any more. It will take a while so I will create a new PR for that when I finish.

PR #25745, opened 2025-10-22
Runnable Embedded GlassFish improvements (stop on error, print message at the end)

OndroMih:
* stop automatically on error when executing commands
* print Welcome message at the beginning
* print message after shutdown, as the last message. Either OK message or error message.

### DIFF
--- a/nucleus/core/kernel/src/main/java/org/glassfish/runnablejar/InfoPrinter.java
+++ b/nucleus/core/kernel/src/main/java/org/glassfish/runnablejar/InfoPrinter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024 Contributors to the Eclipse Foundation
+ * Copyright (c) 2024, 2025 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -21,7 +21,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
 
-import org.glassfish.embeddable.GlassFishException;
 import org.glassfish.grizzly.config.dom.NetworkListener;
 import org.glassfish.grizzly.config.dom.Protocol;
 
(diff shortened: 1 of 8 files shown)


nucleus/core/kernel/src/main/java/org/glassfish/runnablejar/UberMain.java
```diff
@@ -213,23 +252,38 @@ private void addShutdownHook() {
 
             @Override
             public void run() {
-                if (glassFish != null) {
-                    stopGlassFish();
+                // We restore the handlers because they are already removed from the logger in the shutdown hook
```
dmatej: This is quite certainly bad idea, I will remember later why.
Btw logging should not be used for communication with the user, but we have this problem all over the sources. I already created an issue for that.

OndroMih: This is not communication with the user, it’s a proper log message at the end. It can happen in Kubernetes for example, if the service is misconfigured.

what else do you suggest? System.err?

nucleus/core/kernel/src/main/java/org/glassfish/runnablejar/UberMain.java
```diff
@@ -80,6 +81,11 @@ public class UberMain {
 
     GlassFish glassFish;
     CommandRunner commandRunner;
+    String fatalErrorMessage = null;
+    String goodByeMessage = "GlassFish shut down. See you soon!";
+    boolean shutdownRequested = false;
+    // to be used in a shutdown hook when handlers are cleared already
+    Handler[] shutdownkLogHandlers;
```
pzygielo: _k_?

OndroMih: Thanks

nucleus/core/kernel/src/main/java/org/glassfish/runnablejar/InfoPrinter.java
```diff
@@ -124,9 +124,12 @@ private int processApplications(List<Application> applications, final List<Strin
             output.append(BORDER_CHARACTER)
                     .append(" ")
                     .append(app.getName())
-                    .append(" deployed at ")
-                    .append(appUrls)
-                    .append(" ")
+                    .append(" deployed");
```
pzygielo: Could the chain of `.append`s be kept unbroken? Maybe with `?`?

OndroMih: OK

dmatej (changes requested):
Shutdown hooks are there to do the cleanup. When they stop logging, all output properly closed, going against the stream is wrong. In extreme it can cause some damage.

If I remember well, the last message is already written by the shutdown hook, if it is not, we can fix it, but this is really a bad hack.

Handlers are also closeable (but don't implement Closeable), so they should not be used after the logging system was stopped.

I will take a look later ... embedded is bit different than Server, ie. shutdown hooks might not be executed as embedded was started inside existing JVM ... so perhaps something really has to be done, but not this way.

dmatej (approved):

PR #25747, opened 2025-10-23
Allow to specify options without equals in Glassfish Embedded

Thihup:


nucleus/core/kernel/src/test/java/org/glassfish/runnablejar/commandline/CommandLineParserTest.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    static Stream<org.junit.jupiter.params.provider.Arguments> portValues() {
+        return Stream.of(
+                of((Object) new String[]{"--port=9090"}),
+                of((Object) new String[]{"--port", "9090"})
+        );
+    }
+
+    static Stream<org.junit.jupiter.params.provider.Arguments> verboseValues() {
```
OndroMih: would be better to move this method just above the `parseVerboseOption` method, where it's used.

OndroMih (approved):

nucleus/core/kernel/src/test/java/org/glassfish/runnablejar/commandline/CommandLineParserTest.java
```diff
@@ -0,0 +1,84 @@
+package org.glassfish.runnablejar.commandline;
```
pzygielo, line 16: Can't see _This program and the accompanying materials are made available under the ..._.

Thihup: Done!

PR #25750, opened 2025-10-24
Modernized admingui module

samsonsouley:
- Create Logger settings page
- Add primefaces


appserver/admingui/war/pom.xml
```diff
@@ -65,6 +65,17 @@
...
+            <groupId>${project.groupId}</groupId>
+            <artifactId>console-hello-world-plugin</artifactId>
+            <version>${project.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
```
OndroMih, line 78:
```suggestion
        </dependency>
        <dependency>
            <groupId>${project.groupId}</groupId>
            <artifactId>modernized-admingui</artifactId>
            <version>${project.version}</version>
        </dependency>
```

This should be without exclusions, then it will bring also primefaces to the admin app.

PR #25751, opened 2025-10-25
Fix GlassFish static shell bootstrap setup

OndroMih:
* Was missing several JARs in the JAR classpath after they were relocated earlier
* Behaved as DAS. Now behaves as Embedded
* Added static shell as another target for Embedded tests

### DIFF
--- a/appserver/tests/embedded/runnable/src/test/java/org/glassfish/tests/embedded/runnable/TestArgumentProviders.java
+++ b/appserver/tests/embedded/runnable/src/test/java/org/glassfish/tests/embedded/runnable/TestArgumentProviders.java
@@ -22,6 +22,7 @@ import java.util.stream.Stream;
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.ArgumentsProvider;
+import org.junit.jupiter.params.support.ParameterDeclarations;
 
 /**
  *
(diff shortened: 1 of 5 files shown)


OndroMih:
There are tests failing related to Embedded GF in normal server. Looks like the Embedded runtime is also used in server in some obscure scenario. I will need to find a better way to detect when `EmbeddedGlassFishRuntime` runs as a static shell and should report Embedded in the environment.

@dmatej, do know know under which situation is `EmbeddedGlassFishRuntime` used in Server (not Embedded GlassFish) ?

appserver/tests/embedded/runnable/src/test/java/org/glassfish/tests/embedded/runnable/TestArgumentProviders.java
```diff
@@ -32,15 +33,27 @@ public class TestArgumentProviders {
     public static class GfEmbeddedJarNameProvider implements ArgumentsProvider {
 
         @Override
-        public Stream<? extends Arguments> provideArguments(ExtensionContext ec) throws Exception {
+        public Stream<? extends Arguments> provideArguments(ParameterDeclarations declarations, ExtensionContext ec) throws Exception {
             List<Arguments> arguments = new ArrayList<>();
             arguments.add(Arguments.of("glassfish-embedded-all.jar"));
             if (!GfEmbeddedUtils.isDebugEnabled()) {
                 arguments.add(Arguments.of("glassfish-embedded-web.jar"));
+                arguments.add(Arguments.of("glassfishXX/glassfish/lib/embedded/glassfish-embedded-static-shell.jar"
```
dmatej, line 41: The root directory can have any name. Subdirectories should not be changed, so we can rely on `glassfish`, but its parent doesn't have any convention, people do rename it.

dmatej:
Static is not really embedded, it is rather a bastard made just for few TCK tests if I remember well. It should not be used at all, and maybe for those TCKs we should enforce real embedded jar instead and kill this monster. It is a monster because it uses "random" jar files and files spread over standard Eclipse GlassFish server.
I guess server doesn't use it at all ...?

dmatej:
Aha, I was wondering what was the motivation behind this, I do remember we were talking about it.

PR #25752, opened 2025-10-25
Fixes in monitoring, enable monitoring in Embedded GlassFish

OndroMih:
Ensures that the reported current thread count and busy thread count is exactly the number of threads reported by the pool. Even if monitoring is started while thread pool is actively processing requests.

A few improvements in Embedded GlassFish to enable monitoring and access it via JMX (it was completely broken before):

* if the flashlight-agent is already loaded, don't attempt to load it - that would fail on newer Java versions
* concatenate `probe-provider-class-names` MANIFEST elements from dependencies, using gmaven plugin script at build time. It references all probes that should be activated

Embedded GlassFish requires running with flashlight-agent on command line with -javaagent:flashlight-agent.jar on newer Java versions as it doesn't contain the agent. Dynamic loading of agent is not a good practice with newer Java versions anyway.

When started with command "enable-monitoring --mbean true", metric will be available via JMX beans:

Example to start Embedded GlassFish with thread pool and http service metrics:

```
java -javaagent:/path/to/flashlight-agent.jar -jar glassfish-embedded-all.jar 'enable-monitoring --modules thread-pool:http-service'
```

Or, alternatively, using the following properties in `glassfish.properties` in the current directory:

```
configs.config.server-config.monitoring-service.module-monitoring-levels.thread-pool=HIGH
configs.config.server-config.monitoring-service.module-monitoring-levels.http-service=HIGH
```

And command:

```
java -javaagent:/path/to/flashlight-agent.jar -jar glassfish-embedded-all.jar
```

Then in VisualVM (with MBeans extension), in the `amx` node:

image

If you don't see the `amx` node, you need to execute the `bootAMX` operation from VisualVM first (the added test demonstrates how to do that in code):

image


nucleus/flashlight/framework/src/main/java/org/glassfish/flashlight/provider/FlashlightProbe.java
```diff
@@ -148,11 +150,15 @@ public void fireProbe(Object[] params) {
...
-                invoker.invoke(params);
+        try {
+            for (int i=0; i<sz; i++) {
+                ProbeClientInvoker invoker = invokerList.get(i);
+                if(invoker != null) {
+                    invoker.invoke(params);
+                }
             }
+        } catch (IndexOutOfBoundsException e) {
+            logger.log(FINE, "", e);
```
dmatej: Really? Maybe it would be better to clone the list and then use just foreach.

appserver/extras/embedded/pom.xml
```diff
@@ -52,4 +56,62 @@
         </dependencies>
     </dependencyManagement>
 
+    <build>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>org.codehaus.gmaven</groupId>
+                    <artifactId>groovy-maven-plugin</artifactId>
```
dmatej, line 64: I would prefer using java and exec plugin; not sure if it is doable. 
The groovy is not so well readable for us and we tend to do evil things: catching exceptions without processing them, closing jars just when their processing doesn't throw exceptions instead of try-with ...

nucleus/flashlight/framework/src/main/java/org/glassfish/flashlight/provider/FlashlightProbe.java
OndroMih: I know it's not nice. But I suspect  this situation happens when the list is modified in parallel. And when a probe is removed, we don't want to execute it, do we?

OndroMih: Or, actually, I didn't realize that a probe can be removed also from the middle. So you are right. I'll do as you suggest. It doesn't matter if the probe is execute a few moments after it was removed.

dmatej: Yeah, in such cases we cannot be sure at any point about the state of the list or even its content, so first we should make iterating safe, then check every element.

appserver/extras/embedded/pom.xml
OndroMih: I was lazy here and asked AI. It couldn't figure out how to do it using Ant. It kept suggesting using exec plugin with a shell script, which obviously wouldn't work on Windows. Then I decided to use Groovy, which works everywhere. Another option would be to create a new maven plugin, but that would have to be a separate project, or part of the buildhelper plugin, which is external and we would have to release it first. 

Another option is to create a script in Java and execute it via the exec plugin, maybe that's what you mean. If we can compile it first or execute it directly as a script, then it should work too.

OndroMih: I tried. It's just too complicated. Groovy script is the best option. The only other option is to create a separate maven plugin, because we need to set a maven project property, which is used in another plugin to build the final MANIFEST. And I don't want to bother with a separate Maven plugin. The Groovy script is almost like Java and it's pretty simple.

appserver/extras/embedded/all/pom.xml
```diff
@@ -127,6 +127,17 @@
                 </executions>
             </plugin>
 
+            <plugin>
```
pzygielo, line 130: What is the purpose of this plugin being declared here? Is this project parent of some other module?

appserver/extras/embedded/web/pom.xml
```diff
@@ -113,6 +113,17 @@
                 </executions>
             </plugin>
 
+            <plugin>
```
pzygielo, line 116: as https://github.com/eclipse-ee4j/glassfish/pull/25752/files#r2484673287

appserver/extras/embedded/pom.xml
pzygielo, line 64: What is the version of the plugin used here?

appserver/extras/embedded/all/pom.xml
OndroMih: The plugin's defaults are set in the parent, in pluginManagement. To actually run it, the execution `merge-manifest-values-to-properties`  is bound here to the `prepare-package` phase. The same in the `web/pom.xml`

pzygielo: Why not just (bind it in mgmt) and
```xml
<plugin>
  <groupId>org.codehaus.gmaven</groupId>
  <artifactId>groovy-maven-plugin</artifactId>
</plugin>
```
then? The same in the `web/pom.xml`.

appserver/extras/embedded/pom.xml
pzygielo: c7d4a9b73f965d96ca7fbd5d009af9717908b305

appserver/extras/embedded/all/pom.xml
OndroMih: This is the pattern I usually use, so that the downstream module decides. But I can do as you suggest.

.github/workflows/build-linux.yml
```diff
@@ -28,3 +28,8 @@ jobs:
         name: server-logs
         path: '**/server.log*'
         retention-days: 3
+    - name: Publish Test Report
+      uses: mikepenz/action-junit-report@v5
+      if: success() || failure() # always run even if the previous step fails
```
pzygielo: Why not `always()` and no comment? (I find the current comment as misleading a bit as it says _always_, but `cancelled` is not covered...)

appserver/extras/embedded/all/pom.xml
pzygielo: If it's only me - let's keep it the way it is now.
Thanks.

.github/workflows/build-linux.yml
OndroMih: I copied it from the example: https://github.com/mikepenz/action-junit-report?tab=readme-ov-file#configure-the-workflow

 I assume the example meant that tests are not gathered if the build is cancelled and I think it's reasonable. But the comment is misleading, true, I'll change it

pzygielo: I'd prefer no comment here at all :smile:

OndroMih: I chose a compromise and simplified the comment to "# do not run if build cancelled"

OndroMih:
It seems that each time when the Windows build got cancelled after 6 hours, it was stuck in the MonitoringTest for Embedded GlassFish, which starts GlassFish with the flashlight agent. For now, I disabled that test on WIndows. In the future, we want to refactor the monitoring engine and get rid of the agent anyway.

.github/workflows/build-linux.yml
```diff
@@ -3,6 +3,8 @@ name: Build GlassFish on Ubuntu
 on:
   pull_request:
     branches: [ "master", "7.0", "8.0" ]
+permissions:
```
pzygielo: Is this required, unlike in other workflows?

dmatej: First idea was that the windows container is so slow, so Ondro tried to find some solution to improve its performance using ram disks etc. Now I am slowly fixing some problematic code pieces and adding logs to find out why the execution just times out after several hours and just with Windows GHA container.
I will try to run that yet on Windws Server 25 in VirtualBox for sure, because this is really awful behavior, looks like some black hole network communication.

pzygielo: My question is for
```yaml
permissions:
  checks: write
```

I've seen commit with that setting being removed (undone) from other workflow files, but it's still in linux one.

OndroMih: I think @dmatej you reverted too much in this commit: https://github.com/eclipse-ee4j/glassfish/pull/25752/commits/3fb9a43c0f9a5f8f91a95826d81fd820dda8b564

I added 
```
permissions:
  checks: write
```
for the `action-junit-report` step, although I added it with a commit that only mentions adding RAM disk. Though it seems that have this permission block doesn't have any effect.

I don't understand how the `action-junit-report` exactly works. Now I tried to enable annotation-only mode, hopefully it will provide us what I seek for - information about which tests fail if some tests fail. Right now, this step only provides the number of passed and failed tests.

dmatej: It is a work in progress, I will go over those commits later one by one and move some changes to own PR. First I need to understand what is really happening and why the build freezes just on GHA.

dmatej:
Status: I have rebased #25768 to this, fixed conflicts (I made them). Now I will create another PR just with Ondro's changes in GHA, then I will rebase again, I will try yet to improve the performance of those new tests in application-tests, so we could avoid those tags. I have a new weapon for that, learned new usages of the Stream API :-)

Otherwise, nothing new about Jenkins performance, unfortunately.

arjantijms (approved):

PR #25754, opened 2025-10-28
Add Jenkins Checkstyle plugin to show info when checkstyle fails the build

OndroMih:


dmatej:
I have rebased your branch, I will try to finish and merge it asap.

dmatej:
Quite good :-)
image
image

dmatej (approved):

OndroMih:
Though, by looking at the failed job, it's not clear that it failed because of checkstyle error: https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/job/PR-25754/4

Or am I missing something, @dmatej ?

dmatej:
I guess we would have to migrate to standard maven based job. The groovy script based is bit limited. I am already thinking about it, but I don't want to burn too much time on things which could just vanish after we finally migrate those ant based tests to maven; then I would completely rewrite the way how we process PRs to align with Maven better.

Actually opening the log file from the pipeline view is quite quick. You can also switch in your Jenkins profile to the new UI which is bit faster in this.

PR #25756, opened 2025-10-28
Fixed NPEs in JDBC pooling + extended performance test

dmatej:
### Test
- Can use embedded OR server as an alternative
- Parameters were extracted to own class, makes experiments easy
- create+list separated
- Using ShrinkWrap instead of the dependency plugin to resolve dependencies makes possible also running the test in IDE.

### ConnectionPool
- extracted methods made code safer to NPEs
- System.Logger
- fixed several issues with wait-notify; could also time out


dmatej:
#25752 is related a bit - with that I could enable the monitoring for embedded too :-)

arjantijms (approved):

PR #25768, opened 2025-11-05
Fixing embedded monitoring on windows

dmatej:
Clone of #25752 to avoid conflicts; needed as some issues are reproducible just on CI.


nucleus/common/amx-core/src/main/java/org/glassfish/admin/amx/impl/config/AMXConfigLoader.java
```diff
@@ -170,12 +172,11 @@ private void sortAndDispatch(
                 // ADD: a new ConfigBean was added
                 final ConfigBeanProxy cbp = (ConfigBeanProxy) newValue;
                 final ConfigBean cb = asConfigBean(Dom.unwrap(cbp));
-                final boolean doWait = amxIsRunning();
-                if (handleConfigBean(cb, doWait)) // wait until registered
-                {
+                // wait until registered
+                if (handleConfigBean(cb, true)) {
```
OndroMih, line 176: Seems to me that this method is now always called with true, so maybe remove the second parameter completely?

nucleus/common/amx-core/src/main/java/org/glassfish/admin/amx/impl/config/ConfigBeanJMXSupport.java
```diff
@@ -526,6 +527,7 @@ private void findStuff(
                 attrs.add(a);
                 if ( a.returnType() != String.class )
                 {
+                    // FIXME DEADLOCK
```
OndroMih: Can you add some more details about the deadlock?

appserver/tests/embedded/runnable/src/test/java/org/glassfish/tests/embedded/runnable/tool/EGFContainer.java
```diff
@@ -0,0 +1,117 @@
...
+import static java.lang.System.Logger.Level.DEBUG;
+import static java.lang.System.Logger.Level.ERROR;
+import static java.lang.System.Logger.Level.INFO;
+
+/**
+ * Wrapper for Embedded GlassFish process.
+ *
+ * @author David Matejcek
+ */
+public final class EGFContainer {
```
OndroMih: I would rather use full name not just letters: `EmbeddedGlassFishContainer`

appserver/tests/embedded/runnable/src/test/java/org/glassfish/tests/embedded/runnable/MonitoringTest.java
```diff
@@ -0,0 +1,179 @@
...
+    private JMXConnector jmxConnector;
+    private Path warFile;
+    private EGFContainer egf;
+
+
+    @BeforeEach
+    void init() throws Exception {
+        // an app needs to be deployed to initialize request monitoring
+        warFile = createEmptyApp();
+        egf = new EGFContainer("glassfish-embedded-all.jar", warFile);
```
OndroMih: All the other tests in other test classes use 
```
    @ParameterizedTest
    @ArgumentsSource(GfEmbeddedJarNameProvider.class)
```
to run the tests against both All and Web distribution. Why not do it also here?

dmatej: Web is not supported here, you used test parameters everywhere, but in this case when web came, you skipped the test. So I made it explicit.
```
  assumeTrue(!gfEmbeddedJarName.endsWith("web.jar"),
                "AMX is not supported by glassfish-embedded-web.jar, skipping this test scenario");
```

nucleus/common/amx-core/src/main/java/org/glassfish/admin/amx/impl/config/ConfigBeanJMXSupport.java
dmatej: No, it was just a note committed by mistake. There is a lot of synchronization, so I reproduced few deadlocks already, but there's a lot more work to do around these modules.

nucleus/common/amx-core/src/main/java/org/glassfish/admin/amx/impl/config/AMXConfigLoader.java
dmatej: It was rather a quick save and bit of changes. I left those parameters there, I noticed some warning that some methods can be accessed by reflection or something like that. This was effectively a constant, so I made it visible, and ran away from it for now :-)

dmatej: Well, we are both blind, there is one usage with false :)

nucleus/common/amx-core/src/main/java/org/glassfish/admin/amx/impl/config/ConfigBeanJMXSupport.java
dmatej: Removed

dmatej:
Ok, moving back to Ondro's PR.

PR #25773, opened 2025-11-08
Optimizing CI

dmatej:
As Eclipse CI is twice slower than before the migration, we are touching the PR validation again. Related issue: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/6822

Note that Eclipse still works on the issue and also on issues reported by other people, so whatever we do, builds now take too much time and can be unpredictably unstable.

These are the most important changes:

# Splitting integration tests to smaller blocks on Jenkins

* admin-tests, application-tests and embedded-tests have own block now
* embedded was renamed to embedded-tests to avoid conflicts with the compilation of the module

# Ant builds

* Historically, gftest.sh were somehow merged features of CI and local builds. Both parts had unpredictable dependencies based on variables and "heuristic" ensuring that it worked somehow. Every execution created mess on local filesystem, which was not a problem on CI, but locally it was.
* So once I created runtests.sh - it mostly worked, except QL. Now it works for QL too.
* The runtests.sh script sets most of variables, prepares dependencies, executes tests. Therefore I could remove these from CI now, which just uses different two variables, and the script respects that, same other scripts.
* Ant stages were bit reorganized because parallel stages look better this way in Jenkins Pipeline Overview.

# Cancelling running builds when pushing changes

* Now enabled both on Jenkins and GHA
* It should reduce the load of the CI a bit

# Archivation

* Embedded is now in bundles directory too. The archive is used just for distribution using Jenkins API, while tests use normal Maven dependencies to resolve their dependencies.


.lgtm.yml
pzygielo, line 1: With other files removed in this PR - perhaps this one could go as well. LGTM.com is no longer, and GF didn't migrate to GitHub code scanning anyway.

OndroMih (approved):

PR #25789, opened 2025-11-14
ProgressCustomCommand now uses Java Time api

dmatej:
- The former syntax was confusing and overcomplicated
- the test detachOnesAttachMulti did not check job result, now does
- Stabilized tests

### DIFF
--- a/appserver/tests/admin/admin-extension/src/main/java/com/sun/enterprise/tests/progress/ProgressCustomCommand.java
+++ b/appserver/tests/admin/admin-extension/src/main/java/com/sun/enterprise/tests/progress/ProgressCustomCommand.java
@@ -21,7 +21,6 @@ import java.lang.System.Logger;
 import java.time.Duration;
 import java.util.Arrays;
 import java.util.List;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.glassfish.api.ActionReport;
@@ -96,14 +95,14 @@ public class ProgressCustomCommand implements AdminCommand {
 
 
     /**
-     * @param content time of one interval in milliseconds
-     * @param count of intervals
+     * @param timespan total time to split to 100 intervals
      * @return String which can be used as an argument.
      */
-    public static String generateRegularIntervals(long content, int count) {
-        final long[] intervals = new long[count];
-        Arrays.fill(intervals, content);
-        return generateIntervals(intervals);
+    public static String generateRegularIntervals(long timespan) {
+        final int steps = 10;
+        final String[] intervals = new String[steps];
+        Arrays.fill(intervals, Duration.ofMillis(timespan <= steps ? steps : (timespan/steps)).toString());
+        return String.join(",", intervals);
     }
 
     /**
@@ -111,7 +110,8 @@ public class ProgressCustomCommand implements AdminCommand {
      * @return String which can be used as an argument.
      */
     public static String generateIntervals(long... intervals) {
-        return Arrays.stream(intervals).mapToObj(Duration::ofMillis).map(Duration::toString)
-            .collect(Collectors.joining(","));
+        final String[] durations = new String[intervals.length];
+        Arrays.setAll(durations, i -> Duration.ofMillis(intervals[i]).toString());
+        return String.join(",", durations);
     }
 }
--- a/appserver/tests/admin/admin-extension/src/main/java/com/sun/enterprise/tests/progress/ProgressCustomCommand.java
+++ b/appserver/tests/admin/admin-extension/src/main/java/com/sun/enterprise/tests/progress/ProgressCustomCommand.java
@@ -17,9 +17,10 @@
 
 package com.sun.enterprise.tests.progress;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.regex.Pattern;
+import java.time.Duration;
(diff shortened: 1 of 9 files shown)


appserver/tests/admin/admin-extension/src/main/java/com/sun/enterprise/tests/progress/ProgressCustomCommand.java
```diff
@@ -17,174 +17,101 @@
...
-            return origInterval;
-        }
-
-    }
-
-    private Collection<Interval> parsedIntervals;
+    /**
+     * Value must be in for {@code [Nx][MINSEC-]MAXSEC}
+     */
+    @Param(primary = true, separator = ',')
```
OndroMih, line 58: Before it was optional and now it's mandatory? Do I understand it correctly?

OndroMih: Or the `separator` argument means it's a list, and if it's not defined, the array is empty?

appserver/tests/admin/admin-extension/src/main/java/com/sun/enterprise/tests/progress/ProgressCustomCommand.java
```diff
@@ -17,174 +17,101 @@
...
+        final ActionReport report = context.getActionReport();
+        LOG.log(DEBUG, () -> "Intervals: " + Arrays.toString(intervals));
+        if (intervals == null || intervals.length == 0) {
             report.setMessage("Done command process without waiting interval.");
             return;
         }
-        ps.setTotalStepCount(getStepCount());
+
+        final ProgressStatus progressStatus = context.getProgressStatus();
+        final List<Duration> parsedIntervals = Stream.of(intervals).map(Duration::parse).toList();
```
OndroMih, line 72: What happens if Duration.parse cannot parse input?

Can we cover this situation with tests? And also a situation when intervals are not specified, if this isn't covered already.

appserver/tests/admin/admin-extension/src/main/java/com/sun/enterprise/tests/progress/ProgressCustomCommand.java
```diff
@@ -17,174 +17,101 @@
...
             } else {
-                ps.progress(1, "Finished unparsable block [" +blockId + "] " + interval);
+                progressStatus.progress(0, "Starting block [" + blockId + "] " + intervalText);
+                try {
+                    Thread.sleep(interval.toMillis());
+                } catch (InterruptedException ex) {
+                    Thread.currentThread().interrupt();
+                }
+                String message = "Finished block [" + blockId + "] " + intervalText;
+                LOG.log(INFO, message);
```
OndroMih, line 89: There was no log statement before. Shouldn't it be DEBUG instead of INFO?

appserver/tests/admin/admin-extension/src/main/java/com/sun/enterprise/tests/progress/ProgressCustomCommand.java
```diff
@@ -17,174 +17,101 @@
...
+
+    /**
+     * @param content time of one interval in milliseconds
+     * @param count of intervals
+     * @return String which can be used as an argument.
+     */
+    public static String generateRegularIntervals(long content, int count) {
+        final long[] intervals = new long[count];
+        Arrays.fill(intervals, content);
+        return generateIntervals(intervals);
```
OndroMih: Isn't it better to skip creating array and generate a stream right away?

```suggestion
        return IntStream.range(0, count)
        .map(i -> content)
        .mapToObj(Duration::ofMillis)
        .map(Duration::toString)
            .collect(Collectors.joining(","))
```

appserver/tests/admin/admin-extension/src/main/java/com/sun/enterprise/tests/progress/ProgressCustomCommand.java
dmatej: This command is here just for tests, the separator is default the same, however it did not work with duration strings, so tests send just a single argument string internally using the comma as a separator.
I could delete it from the annotation, but I did not.

appserver/tests/admin/admin-extension/src/main/java/com/sun/enterprise/tests/progress/ProgressCustomCommand.java
dmatej: It should crash and burn. The command is just for tests.

appserver/tests/admin/admin-extension/src/main/java/com/sun/enterprise/tests/progress/ProgressCustomCommand.java
dmatej: No, because then there is nothing in server.log to check when test fails. And we are talking about race conditions, so these logs are important.

OndroMih: OK, I didn't realize it's just for tests :)

OndroMih (approved):

PR #25790, opened 2025-11-15
More flexibility to control server start

dmatej:
### New parameters

* User has control what will be considered as the "server successfully started" state.
  * pid file
  * process state
  * any of admin ports
  * all of custom endpoints (might be useful when we can access some public endpoint and we would need to block some script until the endpoint is publicly visible; I doubt if I would ever use it, but ...)
* printServerOutput - explicitly ask to print the process output in the report. If not set, it is printed just if the terse was not set to true and if the start failed.

### Logging

* If terse not set, all start and restart commands now print some basic info about the server
  * Can be extended later
* When Restart Domain and Restart Local Instance now fail, they print the output of the failed start of the server. The start of the server already knows if it is restarting or just starting, so it also knows if there's some parent able to read the output. If there is no such parent, it will set the log handler and the local command then can read it.

### Fixed Bugs

* Admin endpoints can be a list, but we used just first item. At least one should be accessible. 
* Race conditions between processes
  * Not sure if I can remember well and describe them well
  * TBD, I have to read diff

### DIFF
--- a/appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/asadmin/Asadmin.java
+++ b/appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/asadmin/Asadmin.java
@@ -242,7 +242,7 @@ public class Asadmin {
      */
     private AsadminResult exec(final Integer timeout, final boolean detachedAndTerse, final String... args) {
         final List<String> parameters = Arrays.asList(args);
-        LOG.log(TRACE, "exec(timeout={0}, detached={1}, args={2})", timeout, detachedAndTerse, parameters);
+        LOG.log(INFO, "exec(timeout={0}, detached={1}, args={2})", timeout, detachedAndTerse, parameters);
         final List<String> command = new ArrayList<>();
         if (asadmin.getName().endsWith(".java")) {
             command.add(JAVA_EXECUTABLE);
--- a/appserver/orb/orb-connector/src/main/java/org/glassfish/enterprise/iiop/api/GlassFishORBHelper.java
+++ b/appserver/orb/orb-connector/src/main/java/org/glassfish/enterprise/iiop/api/GlassFishORBHelper.java
@@ -97,6 +97,14 @@ public class GlassFishORBHelper implements ORBLocator {
             // Still, threads already working with the instance will have it unstable.
             final ORB destroyedOrb = orb;
             orb = null;
+            // FIXME: com.sun.corba.ee.impl.transport.AcceptorImpl.getAcceptedSocket(AcceptorImpl.java:127)
+            //        can still be blocked in standalone thread, that would lead to its failure
+            //        and cascade leading sockets open. Restart of the server could fail then.
+            try {
+                Thread.sleep(1000L);
+            } catch (InterruptedException e) {
+                // We don't want to interrupt here.
+            }
             destroyedOrb.destroy();
         }
     }
--- a/appserver/orb/orb-connector/src/main/java/org/glassfish/enterprise/iiop/api/GlassFishORBHelper.java
+++ b/appserver/orb/orb-connector/src/main/java/org/glassfish/enterprise/iiop/api/GlassFishORBHelper.java
@@ -17,21 +17,19 @@
 
 package org.glassfish.enterprise.iiop.api;
 
-import com.sun.logging.LogDomains;
-
 import jakarta.annotation.PostConstruct;
-import jakarta.annotation.PreDestroy;
 import jakarta.ejb.Singleton;
 import jakarta.inject.Inject;
 import jakarta.inject.Provider;
 
+import java.lang.System.Logger;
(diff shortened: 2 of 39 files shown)


arjantijms (approved):

appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/asadmin/Asadmin.java
```diff
@@ -242,7 +242,7 @@ private File getPasswordFile() {
      */
     private AsadminResult exec(final Integer timeout, final boolean detachedAndTerse, final String... args) {
         final List<String> parameters = Arrays.asList(args);
-        LOG.log(TRACE, "exec(timeout={0}, detached={1}, args={2})", timeout, detachedAndTerse, parameters);
+        LOG.log(INFO, "exec(timeout={0}, detached={1}, args={2})", timeout, detachedAndTerse, parameters);
```
dmatej, line 245: Not seeing this when I had issues on GHA was very confusing.

appserver/orb/orb-connector/src/main/java/org/glassfish/enterprise/iiop/api/GlassFishORBHelper.java
```diff
@@ -75,17 +77,35 @@ public class GlassFishORBHelper implements ORBLocator {
 
     @PostConstruct
     public void postConstruct() {
+        // WARN: Neither PreDestroy annotation nor interface worked!
```
dmatej, line 80: PreDestroy annotation doesn't work here.
PreDestrou interface doesn't work too.
This was here on 2024, in one change I removed it, I was wrong.

appserver/orb/orb-connector/src/main/java/org/glassfish/enterprise/iiop/api/GlassFishORBHelper.java
```diff
@@ -75,17 +77,35 @@ public class GlassFishORBHelper implements ORBLocator {
...
-        LOG.log(Level.CONFIG, "ORB Shutdown started");
-        if (orb != null) {
-            orb.destroy();
+        LOG.log(INFO, "ORB shutdown started");
+        if (this.orb != null) {
+            // First remove, then destroy.
+            // Still, threads already working with the instance will have it unstable.
+            final ORB destroyedOrb = orb;
             orb = null;
+            // FIXME: com.sun.corba.ee.impl.transport.AcceptorImpl.getAcceptedSocket(AcceptorImpl.java:127)
```
dmatej, line 100: #25804 will target this.

appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IIOPSSLSocketFactory.java
```diff
@@ -251,25 +253,23 @@ public void setORB(ORB orb) {
...
         if (orb.getORBData().acceptorSocketType().equals(SOCKETCHANNEL)) {
             ServerSocketChannel serverSocketChannel = ServerSocketChannel.open();
             serverSocket = serverSocketChannel.socket();
         } else {
             serverSocket = new ServerSocket();
         }
-
-         serverSocket.bind(inetSocketAddress);
-         return serverSocket;
+        final Action<ServerSocket> action = () -> {
```
dmatej, line 268: Desperate action, I will recheck in #25804 if it helps anything. When GF + Corba will reliably close all open ports on shutdown, this should be removed.

appserver/tests/admin/tests/src/test/java/org/glassfish/main/admin/test/AsadminLoggingITest.java
```diff
@@ -69,11 +71,16 @@ public class AsadminLoggingITest {
...
 
+    /** Fill up the server log. */
     @BeforeAll
-    public static void fillUpServerLog() {
-        // Fill up the server log.
-        AsadminResult result = ASADMIN.exec("restart-domain", "--timeout", "60");
-        assertThat(result, asadminOK());
+    public static void fillUpServerLog() throws Exception {
+        if (OS.isWindowsForSure()) {
+            // For some reason windows can collide on debug port.
```
dmatej, line 78: I don't experience blocked `--debug` ports on Linux in `restart-domain` command, but GHA on Windows does it. Seems like it doesn't close debug port when one process spawns another - the new then tries to open the same port, could reuse it, which is not a good idea, but ... 
I think we had this problem on Solaris many years ago too.

appserver/tests/appserv-tests/devtests/ejb/ports.java
```diff
@@ -0,0 +1,76 @@
+/*
```
dmatej, line 1: I tried to use AI to generate free ports, but finally this was easier.

PR #25796, opened 2025-11-21
Trivial smoke test for Jakarta Data

arjantijms:
Very simple smoke test for Jakarta Data using the default data source and Jakarta Persistence.


appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/GlassFishTestEnvironment.java
```diff
@@ -268,6 +268,22 @@ public static <T extends HttpURLConnection> T openConnection(final boolean secur
         return connection;
     }
 
+    /**
+     * Creates a {@link HttpResponse} for the default port using unsecured HTTP.
+     *
+     * @param context - part of the url behind the <code>http://localhost:[port]</code>
+     * @return a new disconnected {@link HttpURLConnection}.
```
pzygielo: 👀

PR #25807, opened 2025-11-28
Documentation for --server-output, plus fixes after testing

dmatej:
- Refactoring around GFLauncher and both children
- Fixed printing output for restart-domain
- Fixed links to JPDA
- Improved output of start/restart commands
  - ie. Konsole in KDE now allows to click directly on the admin endpoint link and open it in a default browser
- I did not document other parameters added in #25790 , because it is not much safe to use them as I tried locally and on Windows machines. They might be useful sometimes, however, ie. when we start docker container, and we don't care about the admin port, just about PID. However there we would still use rather start-serv ... it depends, because sometimes I am restarting domain in container, and in such case I prefer asadmin. We have pretty flexible monster ;-)

```
/usr/lib/jvm/jdk25/bin/java ./glassfish7/bin/asadmin.java restart-domain 
Server is not running, will attempt to start it...
Waiting until start of domain domain1 completes.. ... finished after 1446 ms.

Successfully started the domain domain1.
  Location: /app/appservers/glassfish7/glassfish/domains/domain1
  Log File: /app/appservers/glassfish7/glassfish/domains/domain1/logs/server.log
  The pid file /app/appservers/glassfish7/glassfish/domains/domain1/config/pid contains pid 2462207.
  Process with pid 2462207 is alive
  Admin Endpoints:
    http://localhost:4848 is reachable.

Command restart-domain executed successfully.

```

### DIFF
--- a/docs/reference-manual/src/main/asciidoc/copy-config.adoc
+++ b/docs/reference-manual/src/main/asciidoc/copy-config.adoc
@@ -79,8 +79,10 @@ asadmin-options::
     requires superuser privileges.
   `JAVA_DEBUGGER_PORT`;;
     This property specifies the port number of the port that is used for
-    connections to the Java Platform Debugger Architecture (JPDA)
-    (https://docs.oracle.com/en/java/javase/17/docs/specs/jpda/jpda.html)
+    connections to the Specifies whether the domain is restarted with
+  https://docs.oracle.com/en/java/javase/17/docs/specs/jpda/jpda.html[Java Platform Debugger Architecture (JPDA)]
+  debugging enabled.
+
     debugger. Valid values are 1-65535. On UNIX, creating sockets that
     listen on ports 1-1024 requires superuser privileges.
   `JMS_PROVIDER_PORT`;;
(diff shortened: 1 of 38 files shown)


arjantijms (approved):

docs/reference-manual/src/main/asciidoc/copy-config.adoc
```diff
@@ -79,8 +79,10 @@ asadmin-options::
     requires superuser privileges.
   `JAVA_DEBUGGER_PORT`;;
     This property specifies the port number of the port that is used for
-    connections to the Java Platform Debugger Architecture (JPDA)
-    (https://docs.oracle.com/en/java/javase/17/docs/specs/jpda/jpda.html)
+    connections to the Specifies whether the domain is restarted with
+  https://docs.oracle.com/en/java/javase/17/docs/specs/jpda/jpda.html[Java Platform Debugger Architecture (JPDA)]
+  debugging enabled.
+
```
pzygielo, line 85: Seems that https://glassfish.org/docs/SNAPSHOT/reference-manual.html#copy-config is not updated with that (yet?) - so can't check how this is rendered - but I don't understand that part in **bold** already:

> This property specifies the port number of the port that is used for
> **connections to the Specifies whether** the domain is restarted with

For me, there is something not right with that sentence...

dmatej: It is on my list as next item 🙂

dmatej: Well, I had it open in my browser the whole time ...

PR #25809, opened 2025-11-30
Merged latest master

dmatej:


pzygielo:
- This would bring no changes after #25810 was merged.

PR #25811, opened 2025-12-02
Added disconnecting HttpURLConnection and replaced ProcessUtils.isListening body

dmatej:
When we do so many checks, the URLConnection caches connections. 
But the biggest issue was SocketChannel.

In both cases when server stopped, the cleanup of 15000 connections was done by the operating system about 1-2 minutes after the end of tests.

Now it is much less, there are still some, perhaps later we should switch to java.net.HttpClient in most cases.

The issue was detected by @arjantijms and probably also by Windows GHA CI.


OndroMih:
```
WebTest.testWeb:37->goGet:83 expected: <1> but was: <0>
```
In all the failed builds (Jenkins, Linux, Windows)

arjantijms (approved):

PR #25818, opened 2025-12-04
MicroProfile Config instance always scoped per application

OndroMih:
This PR addresses 2 things:

1. Optimization: Disable MicroProfile Config extension if Config is not used in the app, because the CDI beans won't be used
2. An improvement needed for later JNoSQL integration: Enforce MicroProfile Config instance to be unique in different apps at all times

**Background:** 

Normally, MP Config instances are not shared by applications, if the application requests them. But, as I was implementing NoSQL support via JNoSQL I found out, that JNoSQL requests Config and stores it in a static variable. Since, in this context, jNoSQL is going to be an internal GlassFish component, the static variable is shared by all applications and is not released by undeploying the applications.

**Solution:**

Wrap the Config instance into a stateless instance, which retrieves the config from application context and delegates to it. Similar to ApplicationScoped CDI proxies. 

In fact, I also tried to implement it with CDI ApplicationScoped bean and retrieve it via CDI.current(). It worked, but it's even more magic - CDI scans classloaders and tries to determine the beanManager from a classloader that defines the class, by loading the class name via reflection. And I also had to export the CDI bean from the GlassFish Config OSGi package so that Weld can see it. In terms of performance, it looked to me that it was even worse than searching for ApplicationInfo from the current invocation but I didn't measure.

### DIFF
--- a/appserver/microprofile/config/src/main/java/org/glassfish/microprofile/config/util/LazyProperty.java
+++ b/appserver/microprofile/config/src/main/java/org/glassfish/microprofile/config/util/LazyProperty.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+package org.glassfish.microprofile.config.util;
+
+import java.util.function.BiConsumer;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+/**
+ *
+ * @author Ondro Mihalyi
+ */
+public class LazyProperty {
+
+    public static <OBJ, PROP> PROP getOrSet(OBJ object, Function<OBJ, PROP> getter, Supplier<PROP> propertyProducer, BiConsumer<OBJ, PROP> setter) {
+        PROP result = getter.apply(object);
+        if (result == null) {
+            synchronized (object) {
+                result = propertyProducer.get();
+                setter.accept(object, result);
+            }
+        }
+        return result;
+    }
+
+}
--- a/appserver/microprofile/config/src/main/java/org/glassfish/microprofile/config/util/LazyProperty.java
+++ b/appserver/microprofile/config/src/main/java/org/glassfish/microprofile/config/util/LazyProperty.java
@@ -20,12 +20,30 @@ import java.util.function.Function;
 import java.util.function.Supplier;
 
 /**
+ * Utility class for lazy evaluation
  *
  * @author Ondro Mihalyi
  */
-public class LazyProperty {
+public final class LazyProperty {
 
-    public static <OBJ, PROP> PROP getOrSet(OBJ object, Function<OBJ, PROP> getter, Supplier<PROP> propertyProducer, BiConsumer<OBJ, PROP> setter) {
+    private LazyProperty() {
+    }
+
+    /**
+     * Returns a value lazily. Never returns {@code null}. If it would be {@code null}, it creates and sets the value
+     * in a thread-safe way and returns the newly created value.
+     * It's expected that the value set by the {@setter} will be returned by a subsequent call of the {@code getter}.
+     * Therefore subsequent calls to this method would return the value created and set in a previous call.
+     *
+     * @param <OBJ> Type of the object that getter and setter get value from or set value to
+     * @param <PROP> Type of the value to be returned by this method (type of the property to get or set)
+     * @param object Object to get value from or set value to
+     * @param getter Will be used to get the value from the {@code object} argument
+     * @param propertyProducer Produces a new value if {@code getter} returns {@code null}.
+     * @param setter Will be used to set the value produced by {@code propertyProducer} before the value is returned
+     * @return Value returned by {@code getter} or by {@code propertyProducer} if {@code getter} returns {@code null}
+     */
+    public static <OBJ, PROP> PROP getOrCreateAndSet(OBJ object, Function<OBJ, PROP> getter, Supplier<PROP> propertyProducer, BiConsumer<OBJ, PROP> setter) {
         PROP result = getter.apply(object);
         if (result == null) {
             synchronized (object) {
--- a/appserver/microprofile/config/src/main/java/org/glassfish/microprofile/config/util/LazyProperty.java
+++ b/appserver/microprofile/config/src/main/java/org/glassfish/microprofile/config/util/LazyProperty.java
@@ -47,8 +47,10 @@ public final class LazyProperty {
         PROP result = getter.apply(object);
         if (result == null) {
             synchronized (object) {
-                result = propertyProducer.get();
-                setter.accept(object, result);
+                if (result == null) {
+                    result = propertyProducer.get();
+                    setter.accept(object, result);
+                }
             }
         }
         return result;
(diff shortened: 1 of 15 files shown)


nucleus/common/internal-api/src/main/java/org/glassfish/internal/data/ApplicationRegistry.java
```diff
@@ -54,4 +66,26 @@ public Set<String> getAllApplicationNames() {
         return deployedApplications.keySet();
     }
 
+    /**
+     * Return current application or null, if we're in context of the server, outside of a context of any application.
+     *
+     * @return Information about the current application
+     */
+    public ApplicationInfo getCurrentApplicationInfo() {
```
pzygielo: New method. Could it return `Optional` instead and have it exploited at the call-site?

appserver/microprofile/config/src/main/java/org/glassfish/microprofile/config/util/LazyProperty.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.function.BiConsumer;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+/**
+ *
+ * @author Ondro Mihalyi
+ */
+public class LazyProperty {
```
pzygielo: Looks like new utility class. Could it be made non-instantiable?

nucleus/common/internal-api/src/main/java/org/glassfish/internal/data/ApplicationRegistry.java
OndroMih: Thanks, @pzygielo , done.

appserver/microprofile/config/src/main/java/org/glassfish/microprofile/config/util/LazyProperty.java
OndroMih: Thanks, @pzygielo , done.

appserver/microprofile/config/src/main/java/org/glassfish/microprofile/config/util/LazyProperty.java
```diff
@@ -0,0 +1,57 @@
...
+     * It's expected that the value set by the {@setter} will be returned by a subsequent call of the {@code getter}.
+     * Therefore subsequent calls to this method would return the value created and set in a previous call.
+     *
+     * @param <OBJ> Type of the object that getter and setter get value from or set value to
+     * @param <PROP> Type of the value to be returned by this method (type of the property to get or set)
+     * @param object Object to get value from or set value to
+     * @param getter Will be used to get the value from the {@code object} argument
+     * @param propertyProducer Produces a new value if {@code getter} returns {@code null}.
+     * @param setter Will be used to set the value produced by {@code propertyProducer} before the value is returned
+     * @return Value returned by {@getter} or by {@code propertyProducer} if {@code getter{ returns null
```
pzygielo: bad closing for `@code getter`

OndroMih: Thanks, @pzygielo , fixed.

appserver/microprofile/config/src/main/java/org/glassfish/microprofile/config/util/LazyProperty.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param object Object to get value from or set value to
+     * @param getter Will be used to get the value from the {@code object} argument
+     * @param propertyProducer Produces a new value if {@code getter} returns {@code null}.
+     * @param setter Will be used to set the value produced by {@code propertyProducer} before the value is returned
+     * @return Value returned by {@code getter} or by {@code propertyProducer} if {@code getter} returns {@code null}
+     */
+    public static <OBJ, PROP> PROP getOrCreateAndSet(OBJ object, Function<OBJ, PROP> getter, Supplier<PROP> propertyProducer, BiConsumer<OBJ, PROP> setter) {
+        PROP result = getter.apply(object);
+        if (result == null) {
+            synchronized (object) {
```
pzygielo, line 49: > it creates and sets the value
> in a thread-safe way

What if N (N > 1) `Thread`s arrive at this point? How many times the block
```java
  result = propertyProducer.get();
  setter.accept(object, result);
```
will be executed?

appserver/tests/application/src/main/java/org/glassfish/main/test/app/mpconfig/lib/ConfigHolder.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+
+package org.glassfish.main.test.app.mpconfig.lib;
+
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.ConfigProvider;
+
+public class ConfigHolder {
```
pzygielo: Could this be final? Non-instantiable?

appserver/tests/application/src/test/java/org/glassfish/main/test/app/mpconfig/ApplicationScopedMpConfigTest.java
```diff
@@ -0,0 +1,130 @@
...
+import org.junit.jupiter.api.TestMethodOrder;
+import org.junit.jupiter.api.io.TempDir;
+
+import static java.lang.System.Logger.Level.INFO;
+import static org.glassfish.main.itest.tools.GlassFishTestEnvironment.openConnection;
+import static org.glassfish.main.itest.tools.asadmin.AsadminResultMatcher.asadminOK;
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+@TestMethodOrder(OrderAnnotation.class)
```
pzygielo: Can't see any `@Order` here, so what is the meaning of this?

appserver/tests/application/src/main/java/org/glassfish/main/test/app/mpconfig/lib/ConfigHolder.java
OndroMih: Why not. I did it. But it's just a test class so it's not really necessary

appserver/tests/application/src/test/java/org/glassfish/main/test/app/mpconfig/ApplicationScopedMpConfigTest.java
OndroMih: Thanks. There used to be methods with @Order, but I merged 2 methods into a single one and this is a leftover.

appserver/tests/application/src/main/java/org/glassfish/main/test/app/mpconfig/lib/ConfigHolder.java
pzygielo: > But it's just a test class

Ouch, I've missed that!

> so it's not really necessary

True!

appserver/microprofile/config/src/main/java/org/glassfish/microprofile/config/util/LazyProperty.java
OndroMih: Of course, thanks a lot for spotting this! I'm sure I had another null check there at some point but I must have removed it by accident when I was moving things around.

OndroMih: I'm still thinking refactoring this to use Reentrant lock instead of synchronized, just in case it causes some issues with Virtual threads on Java 21.

OndroMih: In the end, I believe there's no need to get rid of `synchronized`. The operations in the synchronized block are not blocking, they are very fast, just creating simple objects. So `synchronized` is OK even with virtual threads on Java 21.

arjantijms:
> This is happening also in the 8.0 branch, without this PR. So not introduced here.

Let's be sure to investigate this (in another issue).

appserver/microprofile/config/src/main/java/org/glassfish/microprofile/config/util/LazyProperty.java
```diff
@@ -0,0 +1,59 @@
...
+                if (result == null) {
```
pzygielo, line 50: I don't get it. How local variable `result` could not be `null` here? Has it been reassigned to anything else since line 47?

OndroMih: Oh, yes, you’re right. The code should call the getter again

OndroMih: Arjan already merged this, so I’ll raise another PR to fix this

OndroMih:
The `ServiceRegistryException` is fixed in https://github.com/eclipse-ee4j/glassfish/pull/25882

PR #25821, opened 2025-12-05
Explicit version of o.g.hk2:osgiversion-m-p

pzygielo:
Currently, underspecified version (depending on build environment) is being used for `glassfish-jar` packaging (`:compute-osgi-version` bound by extension):
```
[2025-12-04T23:21:48.302Z] [INFO] --- osgiversion:4.0.0-M3:compute-osgi-version (default-compute-osgi-version) @ jsr109-impl ---
[2025-12-04T23:21:52.258Z] [INFO] --- osgiversion:4.0.0-M3:compute-osgi-version (default-compute-osgi-version) @ webservices.security ---
```

### DIFF
--- a/nucleus/parent/pom.xml
+++ b/nucleus/parent/pom.xml
@@ -871,6 +871,17 @@
                     </configuration>
                 </plugin>
                 <plugin>
+                    <!--
+                         osgiversion-maven-plugin is used implicitly in modules
+                         of glassfish-jar packaging, as configured by
+                         glassfish-build-maven-plugin (GlassFishJarLifecycle.java).
+                         Managing it here allows its version to be under control.
+                    -->
+                    <groupId>org.glassfish.hk2</groupId>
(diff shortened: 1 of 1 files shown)


dmatej:
Can you add some comments, describe what is it for there? I removed the plugin here: https://github.com/eclipse-ee4j/glassfish/pull/25591 because it seemed it is redundant, so ... what problem the removal caused and why do we need this change? Does it really do something?

nucleus/parent/pom.xml
```diff
@@ -870,6 +870,11 @@
                         <includeDate>false</includeDate>
                     </configuration>
                 </plugin>
+                <plugin>
```
OndroMih, line 873: Can we add a comment with a reason to specify this? E.g. where it's used? Since it's not used anywhere in GF source code but in some dependency.

pzygielo: Added comment to `plugin` section.

OndroMih (approved):

PR #25823, opened 2025-12-06
Do not set fixed values for MP TCKs via Arquillian XML

arjantijms:
Don't need to fix the port explicitly to the default

### DIFF
--- a/appserver/tests/tck/microprofile/arquillian.xml
+++ b/appserver/tests/tck/microprofile/arquillian.xml
@@ -9,11 +9,7 @@
 
     <container qualifier="arquillian-glassfish">
         <configuration>
-            <property name="debug">true</property>
             <property name="allowConnectingToRunningServer">false</property>
-            <property name="adminHost">localhost</property>
-            <property name="adminPort">4848</property>
-            <property name="enableDerby">${enableDerby:false}</property>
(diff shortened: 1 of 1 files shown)


appserver/tests/tck/microprofile/arquillian.xml
```diff
@@ -9,11 +9,7 @@
 
     <container qualifier="arquillian-glassfish">
         <configuration>
-            <property name="debug">true</property>
             <property name="allowConnectingToRunningServer">false</property>
-            <property name="adminHost">localhost</property>
-            <property name="adminPort">4848</property>
-            <property name="enableDerby">${enableDerby:false}</property>
```
OndroMih, line 16: I assume it's still possible to overwrite these values by system properties, e.g. by `glassfish.enableDerby`, is that right?

OndroMih (approved):

appserver/tests/tck/microprofile/arquillian.xml
arjantijms: You set them by by system properties indeed. Arquillian.xml is the final override, so by having adminPort 4848 in Arquillian.xml it would be fixed to that value. Maybe we should reverse that in the plug-in; make sure Arquillian.xml only sets the base values, with system properties determining the final value.

PR #25833, opened 2025-12-11
Fixes for MP Config

arjantijms:
This changes deployer "requires" semamtics to "requires if available", in other words specifies a loading preference when available.

### DIFF
--- a/appserver/web/war-util/src/main/java/org/glassfish/web/loader/JarFileManager.java
+++ b/appserver/web/war-util/src/main/java/org/glassfish/web/loader/JarFileManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Eclipse Foundation and/or its affiliates. All rights reserved.
+ * Copyright (c) 2023, 2025 Eclipse Foundation and/or its affiliates.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -23,8 +23,11 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.System.Logger;
+import java.net.JarURLConnection;
 import java.net.MalformedURLException;
+import java.net.URI;
 import java.net.URL;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


appserver/web/war-util/src/main/java/org/glassfish/web/loader/JarFileManager.java
```diff
@@ -318,13 +330,45 @@ private static void closeJarFiles(List<JarResource> files) {
             if (jarResource.jarFile == null) {
                 continue;
             }
+
             final JarFile toClose = jarResource.jarFile;
             jarResource.jarFile = null;
+
+            // We need to close the Jar file via its URL, so we get hold
+            // of the potentially cached version. Caching of jarFiles in the
```
OndroMih, line 338: This wording is not clear to me. I would change to:
```suggestion
            // We need to close the Jar file via its URL, so we release
            // a potentially cached version. Caching of jarFiles in the
```

appserver/web/war-util/src/main/java/org/glassfish/web/loader/JarFileManager.java
```diff
@@ -318,13 +330,45 @@ private static void closeJarFiles(List<JarResource> files) {
...
+            // OpenJDK is done by sun.net.www.protocol.jar.JarFileFactory
+            // Of course this is an implementation detail, but in the Open JDK
+            // it has worked like this for a long time and will likely stay for
+            // some time to come.
+            closeJarFileViaURL(jarResource.file);
```
OndroMih, line 343: I would move this call to `closeJarFileViaURL` into the `closeJarFile` method, so that it encapsulates the whole logic of closing a JARfile. `closeJarFile` would then need to be changed to accept `JarResource` instead of just `JarFile`.

appserver/web/war-util/src/main/java/org/glassfish/web/loader/JarFileManager.java
arjantijms: > We need to close the Jar file via its URL, so we get hold of the

I will reword that indeed (in a next PR, your feedback and the merge just crossed) and maybe move the comment.

What is important to realize is that when looking at the code below:

```
jarURLConnection.getJarFile()
```

This one has to return the cached jarFile. Once we have the actual cached jarFile (and not a new instance that just happens to point to the same jar on disk), we can clean it from the cache.

appserver/web/war-util/src/main/java/org/glassfish/web/loader/JarFileManager.java
arjantijms: Sure, I'll do a followup up PR for that.

OndroMih: I've raised a PR here: https://github.com/eclipse-ee4j/glassfish/pull/25834
Including my other changes I had already lying around.

PR #25889, opened 2026-01-21
startserv script didn't work if path contained spaces

OndroMih:
E.g. if GlassFish installed in /path/to/with spaces/glassfish

Needed to add '' around each argument and run it using eval apply the quotes, simple exec would ignore the quotes.

Also added some simple clean up of code without functional changes.


appserver/distributions/glassfish-common/src/main/resources/bin/startserv
```diff
@@ -16,9 +16,9 @@
 # SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
 #
 
-startAsMainProcess() {
+startAsMainProcess () {
     if [[ "$@" == "--help" ]] || [[ "$@" == "--help=true" ]] || [[ "$@" == "-?" ]]; then
-      exec "$JAVA" ${ASADMIN_JVM_OPTIONS} --module-path "${ASADMIN_MODULEPATH}" --add-modules ALL-MODULE-PATH -cp "$ASADMIN_CLASSPATH" org.glassfish.admin.cli.AsadminMain start-domain --help
+      exec "$JAVA" ${ASADMIN_JVM_OPTIONS} --module-path "${ASADMIN_MODULEPATH}" --add-modules ALL-MODULE-PATH -cp "$ASADMIN_CLASSPATH" com.sun.enterprise.admin.cli.AdminMain start-domain --help
```
dmatej: AsadminMain is the right one, AdminMain is just for nadmin scripts, and at some point we could remove it in the future.

appserver/distributions/glassfish-common/src/main/resources/bin/startserv
```diff
@@ -49,14 +48,13 @@ startAsMainProcess() {
...
     local FINAL_COMMAND=(${DRY_RUN_OUTPUT[@]:0:${OUTPUT_LENGTH}-2})
-    exec "${FINAL_COMMAND[@]}"
+    eval "exec ${FINAL_COMMAND[@]}"
 
 }
 
 AS_CONFIG="$(dirname "$(realpath "$0")")/../glassfish/config"
 AS_CONFIG_SH="${AS_CONFIG}/config.sh"
 source "${AS_CONFIG_SH}" "$AS_CONFIG" || { echo "${AS_CONFIG_SH} not found" >&2; exit 1; }
-export ASADMIN_CLASSPATH="$AS_INSTALL/appserver-cli.jar:${ASADMIN_CLASSPATH}"
```
dmatej, line 59: appserv-cli contains AsadminMain - you are under appserver directory, so this part should use it.

nucleus/distributions/nucleus-common/src/main/resources/bin/startserv
```diff
@@ -49,13 +48,14 @@ startAsMainProcess () {
...
     local FINAL_COMMAND=(${DRY_RUN_OUTPUT[@]:0:${OUTPUT_LENGTH}-2})
-    exec "${FINAL_COMMAND[@]}"
+    eval "exec ${FINAL_COMMAND[@]}"
 
 }
 
 AS_CONFIG="$(dirname "$(realpath "$0")")/../config"
 AS_CONFIG_SH="${AS_CONFIG}/config.sh"
 source "${AS_CONFIG_SH}" "$AS_CONFIG" || { echo "${AS_CONFIG_SH} not found" >&2; exit 1; }
+export ASADMIN_CLASSPATH="$AS_INSTALL/appserver-cli.jar:${ASADMIN_CLASSPATH}"
```
dmatej: Here in nucleus should be the opposite situation, nucleus doesn't have appserver-cli and uses just AdminMain class (as you left it here, that is correct). AdminMain is in admin-cli.jar, which is on every ASADMIN_CLASSPATH.

dmatej (changes requested):

nucleus/distributions/nucleus-common/src/main/resources/bin/startserv
OndroMih: It's a bit confusing that otherwise the same script is different in this. But OK, I'll keep it as it is.

appserver/distributions/glassfish-common/src/main/resources/bin/startserv
OndroMih: What about asadmin.bat and startserv.bat? Should they use `AsadminMain` or `AdminMain`? Why do `nadmin` scripts use a different class?

OndroMih: I got confused. Here's how it really is:

* scripts in `glassfish-common` module use `org.glassfish.admin.cli.AsadminMain` -> they end up in the top-level `bin` directory in installed GlassFish, plus there's another `asadmin` script in `resources/glassfish/bin` which ends up in `glassfish/bin` in the installation
* scripts in `nucleus-common` use `com.sun.enterprise.admin.cli.AdminMain` -> `nadmin`, `startserv` and `stopserv`, including *.java and *.bat variants

Why does `nadmin` use `AdminMain`? And should `startserv` and `stopserv` scripts in `nucleus-common` also use it or they should instead use `org.glassfish.admin.cli.AsadminMain` as the same scripts in the `glassfish-common` module?

dmatej: Nucleus, nadmin, doesn't know anything about appserv, appserv depends on nucleus.

dmatej (approved):

PR #25902, opened 2026-02-02
Update various versions in pom

arjantijms:


nucleus/parent/pom.xml
```diff
@@ -1316,6 +1316,35 @@
...
+                                            <ignoreVersions>
+                                                <ignoreVersion>
+                                                    <type>regex</type>
+                                                    <version>.*-M.*</version>
+                                                </ignoreVersion>
+                                                <ignoreVersion>
+                                                    <type>regex</type>
+                                                    <version>.*-beta-.*</version>
+                                                </ignoreVersion>
+                                                <!-- Should work but, of course, does not -->
```
OndroMih, line 1329: What is this, @arjantijms ? Is this comment necessary?

OndroMih (approved):

nucleus/parent/pom.xml
dmatej: I would not even add it, I hope that the release will enforce Central to regenerate those metadata files, so they will be correct again.

dmatej (approved):

PR #25904, opened 2026-02-02
Avoid JAR file locking in `WebappClassLoader`

avpinchuk:
The `antiJARlocking` property is always false (is newer set).

 `WebappClassLoader` throws `IllegalStateException` which is swallowing somewhere.

The main idea of this patch is to move the processing of this property from Catalina `StandardContext` to the `WarHandler`. This sets the `antiJARLocking` property before starting webapp class loader and no exception is thrown.


avpinchuk:
> Seems it might be related to issues of GitHub, see https://www.githubstatus.com/
> It refuses to build too.

My Eclipse Account is valid. I logged in without any problems. ECA signed. Also I logged in without any problems to Jenkins. Let's just wait for now.

dmatej (approved):

arjantijms (approved):

PR #25906, opened 2026-02-02
Updated references to Jakarta EE and GlassFish versions

dmatej:


avpinchuk (approved):

appserver/admin/cli-optional/src/main/manpages/com/sun/enterprise/admin/cli/optional/backup-domain.1
```diff
@@ -79,4 +79,4 @@ EXIT STATUS
 SEE ALSO
        restore-domain(1), list-backups(1), suspend-domain(1), resume-domain(1)
 
-Jakarta EE 10                         22 Dec 2010                 backup-domain(1)
+Jakarta EE 11                         22 Dec 2010                 backup-domain(1)
```
OndroMih, line 82: The references o Jakarta EE in every man page are meaningless, isn't it better to just replace them with "Eclipse GlassFish"? So that we don't need to update it with every new Jakarta EE release?

OndroMih: Even better "Eclipse GlassFish 8.0.0" - though I would only do it if we can generate the version based on the pom.xml properties and that's for another PR.

There reference to the date is also meaningless if it's static, It's here only to match the usual footer of man pages. I would rather generate it based on the last change from git history. But again, for another PR.

appserver/connectors/admin/src/main/manpages/org/glassfish/connectors/admin/cli/flush-connection-pool.1
```diff
@@ -32,7 +32,7 @@ DESCRIPTION
        The jndi-name for application-scoped-resources or
        module-scoped-resources are specified using the format
        java:app/jdbc/myDataSource or java:module/jdbc/myModuleLevelDataSource.
-       This naming scope is defined in the Jakarta EE 10 Specification
+       This naming scope is defined in the Jakarta EE 11 Specification
        (https://javaee.github.io/javaee-spec/javadocs/).
```
OndroMih:
```suggestion
       (https://jakarta.ee/specifications/enterprise-beans/4.0/jakarta-enterprise-beans-spec-core-4.0#syntax).
```

nucleus/deployment/admin/src/main/manpages/org/glassfish/deployment/admin/deploy.1
```diff
@@ -515,7 +515,7 @@ EXAMPLES
                asadmin> deploy --deploymentorder 110 --name horse horse.war
                ...
 
-       Example 7, Deploying an Application Using an Alternate Jakarta EE 10
+       Example 7, Deploying an Application Using an Alternate Jakarta EE 11
```
OndroMih, line 518:
```suggestion
       Example 7, Deploying an Application Using an Alternate Jakarta EE
```

I would just remove the Jakarta EE version here, it's useless.

OndroMih (approved):

appserver/admin/cli-optional/src/main/manpages/com/sun/enterprise/admin/cli/optional/backup-domain.1
dmatej: It is a man page format definition, I checked how it looks for other programs. 
```
# GIT
Git 2.43.0     07/02/2025      GIT(1)
MAN
2.12.0    2023-09-23    MAN(1)

```
There is always a program name and version (so GlassFish 8.0.0 would be better), date of last revision (with unstable formatting) and some paging.

I would prefer the man page date format.

We can do that in another PR, this is just a minor update and even here I had to review all files and fix "false positives". In that PR we have to touch all those line again, so it will be a PR of the same size.

arjantijms: > We can do that in another PR,

Okay, so not to let PRs open too long, I'll merge this one and we'll see the further updates in a next pr

PR #25907, opened 2026-02-03
Document that Admin Console shouldn't be enabled in production

OndroMih:


avpinchuk (approved):

avpinchuk:
I have a very early draft of how to protect the console from brute force attacks using a _device cookies_.

dmatej (approved):

PR #25908, opened 2026-02-03
Replace hard-coded version references in docs with existing properties defined in pom.xml

OndroMih:
Moved some properties from appserver/pom.xml to nucleus parent so that they are available in nucleus docs.

Will conflict with https://github.com/eclipse-ee4j/glassfish/pull/25906, which just updates hard-coded values. But doesn't replace it completely.

### DIFF
--- a/appserver/pom.xml
+++ b/appserver/pom.xml
@@ -81,93 +81,6 @@
     <properties>
         <glassfish.version>${project.version}</glassfish.version>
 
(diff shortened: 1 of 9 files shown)


OndroMih:
> Did you check if all that filtering does not damage anything, ie. examples in the documentation?

Yes, I checked the final HTML, all looks good

arjantijms (approved):

PR #25935, opened 2026-02-24
Deployment performance and resource usage optimizations

OndroMih:
Check each commit individually:

* [Imrpoved thread local cleanup, cleanup deployment info not needed after deployment](https://github.com/eclipse-ee4j/glassfish/commit/eb89f7d43be5ee9a23a99da1cad868339832cc78)
* [Jakarta Data optimizations](https://github.com/eclipse-ee4j/glassfish/commit/427d39261d1d514e6a086f0358d56916165d0193) - avoid init if not needed, release type info if not needed
* [Simple clean up](https://github.com/eclipse-ee4j/glassfish/commit/6f269c13a0b2c704438e376102762b316999c24e)

### DIFF
--- a/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/util/ResourceValidator.java
+++ b/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/util/ResourceValidator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022, 2026 Contributors to the Eclipse Foundation
  * Copyright (c) 2017, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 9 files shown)


appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/util/ResourceValidator.java
```diff
@@ -161,16 +165,23 @@ public void event(Event<?> event) {
...
+                    return;
+                }
+                AppResources appResources = new AppResources();
+                parseResources(appResources);
+                validateResources(appResources);
+            } finally {
+                // cleanup variables scoped to this method so that they don't remain in HK2 cache
+                dc = null;
+                application = null;
+                target = null;
```
avpinchuk: Maybe it would be better to convert these three fields into local variables (because the default scope for this validator is singleton)?

OndroMih: I thought it would be too much work, the whole class is huge. But I can have a look if I can do it easily.

avpinchuk: @OndroMih, A quick look shows that they can be passed via method parameters

OndroMih: Yes, they can. But I think they need to be passed to many methods. I thought about extracting the logic to a separate class, create an instance of it in the event method and the throw it away.

OndroMih: I committed the change - wrapped all private methods and fields into a separate nested object, which gets released after every call to `event()`.

avpinchuk: Excellent except for checkstyle violations

OndroMih: I force pushed the checkstyle fixes

avpinchuk (approved):

PR #25942, opened 2026-03-03
Add info about Embedded GlassFish maven plugin to website (and disable tests on docs changes)

OndroMih:
Also modifies PR pipeline to skip running tests when the only changes are in the docs directory.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -297,32 +297,19 @@ pipeline {
                   if (env.CHANGE_TARGET) {
                      echo "PR build detected, checking if only docs changed..."
 
-                     def changedFiles = sh(
+                     def relevantChanges = sh(
                         script: '''
-                           git diff --name-only origin/${CHANGE_TARGET}...HEAD || echo "all"
+                           (git diff --exit-code --name-only origin/${CHANGE_TARGET}...HEAD && echo "all") | sed '/^docs[/]/d'
                         ''',
                         returnStdout: true
                      ).trim()
 
-                     echo "Changed files:\n${changedFiles}"
 
-                     if (changedFiles != "" && changedFiles != "all") {
-                        // Check if all changes are in docs/ directory or subdirectories
-                        def nonDocsChanges = sh(
-                           script: """
-                              echo '${changedFiles}' | grep -v '^docs/' || true
-                           """,
-                           returnStdout: true
-                        ).trim()
-
-                        if (nonDocsChanges == "") {
-                           env.SKIP_TESTS = "true"
-                           echo "✓ Only docs/ changes detected - tests will be skipped"
-                        } else {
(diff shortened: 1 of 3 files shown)


avpinchuk (approved):
Please remove garbage from description.

Funny thing - `pr-merge` for the doc updates always fails. xD

OndroMih:
> Please remove garbage from description.

I hoped that the comments get removed automatically, not the case. I wish Github supported that.

OndroMih:
I'm trying changes in Jenkinsfile to skip running tests if only docs change. I moved to draft for now to revent merging this PR.

Jenkinsfile
```diff
@@ -285,6 +285,52 @@ pipeline {
...
+                  // Default: run tests
+                  env.SKIP_TESTS = "false"
+
+                  // Only check for docs-only changes in PR builds
+                  if (env.CHANGE_TARGET) {
+                     echo "PR build detected, checking if only docs changed..."
+
+                     def changedFiles = sh(
+                        script: '''
+                           git diff --name-only origin/${CHANGE_TARGET}...HEAD || echo "all"
```
dmatej: Why you grep those files when you could probably simplify that to (I used your branch to see what exactly it does):
```
git diff --name-only eclipse/main..ondro/ondromih-2026-03-website-egf-maven-plugin -- docs || echo "all"
```
And yet one problem, it seems git diff always returns 0 error code, so it never echoes "all". I tried with this:
```
git diff --name-only eclipse/main..eclipse/main  || echo "all"
```

dmatej (changes requested):

Jenkinsfile
dmatej: This would work:
```
dmatej@dmatej-tux:~/work/repo/git/glassfish$ git diff --exit-code --name-only eclipse/main..eclipse/main -- docs && echo "all"
all

dmatej@dmatej-tux:~/work/repo/git/glassfish$ git diff --exit-code --name-only eclipse/main..ondro/ondromih-2026-03-website-egf-maven-plugin -- docs && echo "all"
Jenkinsfile
docs/website/src/main/resources/_includes/documentation-toc.html
docs/website/src/main/resources/documentation.md
```

> --exit-code Make the program exit with codes similar to diff(1). That is, it exits with 1 if there were differences and 0 means no differences.

OndroMih: If I use `-- docs`, then the diff is only in the docs directory. If doesn't detect if changes are both in the `docs` directory and somewhere else - e.g. the PR contains both code and docs changes.

I want the opposite - create a diff of all files outside of docs directory, and if there are any changed files, run teh tests, otherwise don't run them.

That's why I first list all the changes and then exclude all docs changes with grep. Though I can do it in a single step.

dmatej (approved):

PR #25951, opened 2026-03-12
Refactoring of utilities to make them usable outside of GlassFish

dmatej:
* glassfish-itest-tools
  * Separated GlassFishEnvironment from GlassFishITestEnvironment - doesn't change the domain, doesn't install or update anything, just provides easy access to server's file system.
  * Asadmin can be configured to use different JDK
  * Useful for users which would want to write tests and even for Eclipse IDE plugin.
* common-util and jdk-extensions
  * ProcessManager moved to the lower level - now we can avoid the whole dependency tree brought by common-util.
* rest-client
  * Removed RestClientLogging and dependency on common-util
  * Bit of a cleanup, it would need to do much more work.
* Fixed JdbcPoolPerformanceIT and tests in the same block
  * Updated dependencies
  * Removed usage of deprecated classes
  * Fixed work with case sensitive db schema
  * Fixed dependency between tests
  * Fixed NPE
* Refactored MarshallingUtils + marked as deprecated
  * It is used just internally, I tried to reduce it a bit.
  * Swallowing exceptions replaced, seems it doesn't break anything.


nucleus/admin/rest/rest-client/src/main/java/org/glassfish/admin/rest/client/utils/MarshallingUtils.java
```diff
@@ -55,29 +60,27 @@ public static List<Map<String, String>> getPropertiesFromJson(String json) {
...
 
     public static List<Map<String, String>> getPropertiesFromXml(String xml) {
         List<Map<String, String>> list = new ArrayList<>();
-        InputStream input = null;
-        try {
             XMLInputFactory inputFactory = XMLInputFactory.newFactory();
             inputFactory.setProperty(XMLInputFactory.IS_SUPPORTING_EXTERNAL_ENTITIES, Boolean.FALSE);
-            input = new ByteArrayInputStream(xml.trim().getBytes("UTF-8"));
+            try (InputStream input = new ByteArrayInputStream(xml.trim().getBytes(UTF_8))) {
             XMLStreamReader parser = inputFactory.createXMLStreamReader(input);
```
avpinchuk, line 79: Indentation

appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/asadmin/Asadmin.java
```diff
@@ -100,6 +99,20 @@ public Asadmin(final File asadmin, final String adminUser, final File adminPassw
...
+    }
+
+    /**
+     * Sets Java home directory used to set <code>AS_JAVA</code> and <code>JAVA_HOME</code>
+     * environment variables and also to run <code>asadmin.java</code>.
+     *
+     * @param javaHome if null, uses current <code>java.home</code> system property.
+     * @return this
+     */
+    public Asadmin withJavaHome(@SuppressWarnings("hiding") final Path javaHome) {
```
avpinchuk: Does this warning only apply to Eclipse?

appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/asadmin/StartServ.java
```diff
@@ -76,7 +78,7 @@ public StartServ withEnv(final String name, final String value) {
      * @param textToWaitFor
      * @return this
      */
-    public StartServ withTextToWaitFor(final String textToWaitFor) {
+    public StartServ withTextToWaitFor(@SuppressWarnings("hiding") final String textToWaitFor) {
```
avpinchuk: And here?

appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/GlassFishEnvironment.java
```diff
@@ -0,0 +1,231 @@
...
+     * Sets credentials for the {@link Asadmin} command api.
+     * Creates an internal temporary password file.
+     *
+     * @param username
+     * @param password
+     * @return this
+     * @throws IllegalStateException if the temporary password file could not be created.
+     */
+    public GlassFishEnvironment withCredentials(
+        @SuppressWarnings("hiding") String username,
```
avpinchuk: Again 'hiding'

appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/GlassFishEnvironment.java
```diff
@@ -0,0 +1,231 @@
...
+        }
+    }
+
+    /**
+     * Sets the JDK home directory.
+     *
+     * @param javaHome if null, the value is set from the system property <code>java.home</code>
+     * @return this
+     */
+    public GlassFishEnvironment withJavaHome(@SuppressWarnings("hiding") Path javaHome) {
```
avpinchuk: Here

appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/GlassFishEnvironment.java
```diff
@@ -0,0 +1,231 @@
...
+     * Sets credentials for the {@link Asadmin} command api.
+     *
+     * @param username
+     * @param passwordFile
+     * @return this
+     */
+    public GlassFishEnvironment withCredentials(
+        @SuppressWarnings("hiding") String username,
+        @SuppressWarnings("hiding") File passwordFile) {
+        this.username = username;
```
avpinchuk, line 96: And here

nucleus/admin/rest/rest-client/src/main/java/org/glassfish/admin/rest/client/utils/MarshallingUtils.java
dmatej: Oh, thanks, I see!

appserver/itest-tools/src/main/java/org/glassfish/main/itest/tools/asadmin/Asadmin.java
dmatej: It seems so. I am still thinking about it, because it is really useful to warn about this when doing refactoring. Another option is using `this.` for fields so when I remove field or parameter, it doesn't automatically take the other one. Sure, warnings work just when there's not a hundred of ignored warnings.
The `this.` is better, because then it doesn't even compile when I remove the field.

But yeah, suppressing warnings ... I will remove it, it is usually a bad practice to ignore any warnings.

avpinchuk (approved):

PR #25962, opened 2026-03-30
Support @EJB(beanName=...) for app clients

arjantijms:


appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/ApplicationClientDescriptor.java
```diff
@@ -149,11 +151,12 @@ public Vector<NamedReferencePair> getNamedReferencePairs() {
      */
     @Override
     public Set<EnvironmentProperty> getEnvironmentProperties() {
-        if (this.environmentProperties == null) {
-            this.environmentProperties = new OrderedSet<>();
+        if (environmentProperties == null) {
+            environmentProperties = new OrderedSet<>();
         }
-        this.environmentProperties = new OrderedSet<>(this.environmentProperties);
```
dmatej, line 155: Do we really wrap it in each and every call? Also, I would prefer keeping this. for fields, because without it it easily causes disasters when refactoring (creating/renaming variables). For methods it is ok to remove it, but for fields it helps. Ie. here the new code looks like if it would be a local variable.

appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/ApplicationClientDescriptor.java
```diff
@@ -179,43 +183,40 @@ public EnvironmentProperty getEnvironmentPropertyByName(String name) {
...
      * Return the set of references to ejbs that I have.
      */
     @Override
     public Set<EjbReferenceDescriptor> getEjbReferenceDescriptors() {
-        if (this.ejbReferences == null) {
-            this.ejbReferences = new OrderedSet<>();
+        if (ejbReferences == null) {
+            ejbReferences = new OrderedSet<>();
         }
-        this.ejbReferences = new OrderedSet<>(this.ejbReferences);
```
dmatej, line 201: Same problem.

appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/ApplicationClientDescriptor.java
```diff
@@ -318,11 +323,12 @@ public ServiceReferenceDescriptor getServiceReferenceByName(String name) {
 
     @Override
     public Set<MessageDestinationReferenceDescriptor> getMessageDestinationReferenceDescriptors() {
-        if (this.messageDestReferences == null) {
-            this.messageDestReferences = new OrderedSet<>();
+        if (messageDestReferences == null) {
+            messageDestReferences = new OrderedSet<>();
         }
-        this.messageDestReferences = new OrderedSet<>(this.messageDestReferences);
```
dmatej, line 324: And again

appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/ApplicationClientDescriptor.java
```diff
@@ -360,34 +367,34 @@ public MessageDestinationReferenceDescriptor getMessageDestinationReferenceByNam
      */
     @Override
     public Set<ResourceEnvReferenceDescriptor> getResourceEnvReferenceDescriptors() {
-        if (this.resourceEnvReferences == null) {
-            this.resourceEnvReferences = new OrderedSet<>();
+        if (resourceEnvReferences == null) {
+            resourceEnvReferences = new OrderedSet<>();
         }
-        return this.resourceEnvReferences = new OrderedSet<>(this.resourceEnvReferences);
```
dmatej, line 366: Even worse, set and return on a single line ...

dmatej (changes requested):
Wow, this class was horrible, always was, good you touched it. I will commit yet once to resolve my review.

dmatej (approved):
Hmm, the `OrderedSet` creates a copy of the collection. I don't like it, but it should not cause a bug I originally thought, if it always replaces the original.
I will try to touch it in standalone PR.

appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/ApplicationClientDescriptor.java
```diff
@@ -453,11 +461,13 @@ public List<InjectionCapable> getInjectableResourcesByClass(String className) {
 
     @Override
     public EjbReferenceDescriptor getEjbReference(String name) {
-        for (EjbReferenceDescriptor ejr : this.getEjbReferenceDescriptors()) {
```
dmatej, line 456: So this is the fix/improvement.

appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/ApplicationClientDescriptor.java
arjantijms: I removed "this", it's implicit. The IDE will colour it accordingly to what it is (may be a setting in the IDE) anyway. The less "clutter" the better the readability. With less clutter, we can focus on what really matters, and give those things proper names instead of using obscure abbreviations like "erd", "epr", etc.

arjantijms:
The beanName is used, in the TCK, in files like this:

```xml
<!DOCTYPE glassfish-application-client PUBLIC "-//GlassFish.org//DTD GlassFish Application Server 3.1 Java EE Application Client 6.0//EN" "http://glassfish.org/dtds/glassfish-application-client_6_0-1.dtd">
<glassfish-application-client>
    <ejb-ref>
        <ejb-ref-name>EnvEntryFieldBean</ejb-ref-name>
        <jndi-name>ejb3_bb_stateless_enventry_EnvEntryFieldBean</jndi-name>
    </ejb-ref>
    <ejb-ref>
        <ejb-ref-name>EnvEntrySetterBean</ejb-ref-name>
        <jndi-name>ejb3_bb_stateless_enventry_EnvEntrySetterBean</jndi-name>
    </ejb-ref>
    <ejb-ref>
        <ejb-ref-name>EnvEntryTypeBean</ejb-ref-name>
        <jndi-name>ejb3_bb_stateless_enventry_EnvEntryTypeBean</jndi-name>
    </ejb-ref>
</glassfish-application-client>
```

Before the change, this only worked for beans defined with `@EJB(name=`. Appearently other servers expect this to work for `@EJB(beanName=`. 

See also
* https://github.com/jakartaee/platform-tck/commit/378fe74f26504063d9eff02a8a4186eb50b3f053
* https://github.com/jakartaee/platform-tck/pull/2666

The different overloads of the EJB annnotation are not so nice and hard to remember. `name`, `beanName`, `mappedName`...

PR #25964, opened 2026-03-31
Fixed ConfigProfileResolver service selection

dmatej:
- The order suddenly changed just by changing the version number to 8.0.1
- Sometimes our resolver was even missing, not sure why 
- Changing order of dependencies still did not ensure order of resolvers, so I excluded the service and explicitly set the one we want.

In the future we should write our own descriptor handler instead of usage `metaInf-services` which merges all it can see, but cannot filter them or order them.

https://maven.apache.org/plugins/maven-assembly-plugin/examples/single/using-container-descriptor-handlers.html


dmatej:
I discovered yet one issue with HK2 - when I added test-utils to the classpath, the scope doesn't matter, the dependency is processed and HK2 uses its services despite I needed just one utility method without any HK2 support.

avpinchuk (approved):

PR #25971, opened 2026-04-06
Fixes authentication with PAM login module

avpinchuk:


docs/reference-manual/src/main/asciidoc/create-auth-realm.adoc
```diff
@@ -55,7 +55,7 @@ asadmin-options::
   `com.sun.enterprise.security.auth.realm.certificate.CertificateRealm`,
   `com.sun.enterprise.security.ee.authentication.glassfish.jdbc.JDBCRealm`,
   `com.sun.enterprise.security.auth.realm.ldap.LDAPRealm`,
```
OndroMih, line 57: Shouldn’t other realms be also in the ee package?

nucleus/admin/template/src/main/resources/config/login.conf
OndroMih, line 19: What about other login modules? Shouldn’t they be in the ee package too?

docs/reference-manual/src/main/asciidoc/create-auth-realm.adoc
OndroMih: OK, seems the other realms are in different packages as documented.

nucleus/admin/template/src/main/resources/config/login.conf
OndroMih: OK, they really are in different packages and the paths are correct.

dmatej (approved):

PR #25972, opened 2026-04-08
Improved behavior related to InetAddress, hostname resolution, added tests

dmatej:
Story: I moved to another wifi, but forgot I used some hardcoded mapping in `/etc/hosts`. GlassFish then could not start, because it could not resolve 192.x.x.x address, telling me that the admin port is already taken (false).

Then I have found also that IPv6 was another issue.

Then different combinations of hostnames, ipv6 and ipv4 addresses, localhost and loopback addresses, and soon it became more and more complicated.

Finally I have removed the `isThisHostLocal` method as it had same results as improved `isLocal`, which is more simple, because Java does most of the job.


appserver/tests/quicklook/amx/src/test/amx/AMXTestBase.java
```diff
@@ -120,7 +120,7 @@ void setUpEnvironment(
 
         mDebug = debug;
         mPort = port;
-
+        System.err.println("host=" + mHost + ", port=" + mPort + ", debug=" + mDebug);
```
OndroMih, line 123: Is this necessary? And all the other prints to err in other tests? Do you want to keep them or they are no longer needed?

nucleus/admin/util/src/main/java/com/sun/enterprise/admin/util/RemoteInstanceCommandHelper.java
```diff
@@ -53,12 +53,11 @@ public RemoteInstanceCommandHelper(ServiceLocator habitatIn) {
...
     public final String getHost(final String serverName) {
-        String host = null;
         Server server = getServer(serverName);
-        if (server != null) {
-            host = server.getAdminHost();
+        if (server == null) {
+            return null;
         }
-        return host;
+        return server.getAdminHost();
```
OndroMih: Could be also written as 


```suggestion
    return (server == null) ? null : server.getAdminHost();
```

nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/CreateInstanceCommand.java
```diff
@@ -371,48 +377,47 @@ private SecureAdminBootstrapHelper createBootstrapHelper(SSHLauncher sshL) throw
...
-            ActionReport report = ctx.getActionReport();
-            report.setActionExitCode(ActionReport.ExitCode.SUCCESS);
-
-            NodeUtils nodeUtils = new NodeUtils(habitat);
-            Server dasServer = servers.getServer(SystemPropertyConstants.DAS_SERVER_NAME);
-            String dasHost = dasServer.getAdminHost();
-            String dasPort = Integer.toString(dasServer.getAdminPort());
-
-            ArrayList<String> command = new ArrayList<>();
+        if (!theNode.isLocal() && !"SSH".equals(theNode.getType())) {
```
OndroMih: Maybe extract `"SSH"` into a constant?

OndroMih: The `"SSH"` literal is used in several places in this file, and is also used in `ClusterHandler.java`, and also in some other *.java files.

OndroMih (approved):

appserver/tests/quicklook/amx/src/test/amx/AMXTestBase.java
dmatej: In these tests are any logs necessary, otherwise you have no clue what is happening when they fail.

nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/CreateInstanceCommand.java
dmatej: There's even a `RemoteType` enum for it, but it is not visible from everywhere ... I will take a look.

avpinchuk (approved):

PR #25973, opened 2026-04-08
Support for graceful socket closes when checking server endpoint listening via the admin CLI

carryel:
For macOS, immediately closing a client socket results in an incomplete socket state on the server, causing unnecessary server errors. Please refer to the issue below:

- https://github.com/eclipse-ee4j/glassfish-grizzly/issues/2285

### DIFF
--- a/nucleus/admin/cli/src/main/java/com/sun/enterprise/admin/cli/CLIConstants.java
+++ b/nucleus/admin/cli/src/main/java/com/sun/enterprise/admin/cli/CLIConstants.java
@@ -37,6 +37,7 @@ public final class CLIConstants {
     public static final String EOL = System.lineSeparator();
 
     public static final Duration WAIT_FOR_DAS_TIME_MS = getEnv(TIMEOUT_START_SERVER);
+    public static final Duration WAIT_FOR_DAS_POLLING_INTERVAL = Duration.ofMillis(100);
     public static final int RESTART_NORMAL = 10;
     public static final int RESTART_DEBUG_ON = 11;
     public static final int RESTART_DEBUG_OFF = 12;
--- a/nucleus/common/common-util/src/main/java/com/sun/enterprise/universal/process/ProcessUtils.java
+++ b/nucleus/common/common-util/src/main/java/com/sun/enterprise/universal/process/ProcessUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2025 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022, 2026 Contributors to the Eclipse Foundation
  * Copyright (c) 2009, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -166,7 +166,7 @@ public final class ProcessUtils {
         final Supplier<Boolean> action = () -> {
             try (Socket server = new Socket()) {
                 // Force RST on close
-                server.setSoLinger(true, 0);
+                server.setSoLinger(true, 1);
                 server.setSoTimeout(timeout == null ? 0 : (int) timeout.toMillis());
                 try {
                     server.connect(endpoint.toInetSocketAddress(), SOCKET_CONNECT_TIMEOUT);
@@ -214,7 +214,7 @@ public final class ProcessUtils {
     public static boolean isListening(HostAndPort endpoint) {
         try (Socket socket = new Socket()) {
             // Force RST on close
-            socket.setSoLinger(true, 0);
+            socket.setSoLinger(true, 1);
             socket.connect(endpoint.toInetSocketAddress(), SOCKET_CONNECT_TIMEOUT);
             return true;
         } catch (IOException e) {
@@ -274,15 +274,36 @@ public final class ProcessUtils {
      * @return true if the sign returned true before timeout.
      */
     public static boolean waitFor(Supplier<Boolean> sign, Duration timeout, boolean printDots) {
-        LOG.log(DEBUG, "waitFor(sign={0}, timeout={1}, printDots={2})", sign, timeout, printDots);
+        return waitFor(sign, timeout, null, printDots);
+    }
+
+    /**
+     * @param sign logic defining what we are waiting for.
+     * @param timeout can be null to wait indefinitely.
+     * @param interval interval for polling until the timeout. can be null to poll without an interval.
+     * @param printDots print dot each second and new line in the end.
+     * @return true if the sign returned true before timeout.
+     */
+    public static boolean waitFor(Supplier<Boolean> sign, Duration timeout, Duration interval, boolean printDots) {
+        LOG.log(DEBUG, "waitFor(sign={0}, timeout={1}, interval={2}, printDots={3})", sign, timeout, interval, printDots);
         final Instant start = Instant.now();
         final Instant deadline = timeout == null ? null : start.plus(timeout);
         final Supplier<Boolean> action = () -> {
-            while (deadline == null || Instant.now().isBefore(deadline)) {
+            Instant current = null;
+            while (deadline == null || (current = Instant.now()).isBefore(deadline)) {
                 if (sign.get()) {
                     return true;
                 }
-                Thread.onSpinWait();
+                final Duration pollingInterval = calculatePollingInterval(interval, deadline, current);
+                if (pollingInterval != null) {
+                    try {
+                        Thread.sleep(interval);
+                    } catch (InterruptedException e) {
+                        throw new RuntimeException(e);
+                    }
+                } else {
+                    Thread.onSpinWait();
+                }
             }
             return false;
         };
@@ -293,6 +314,23 @@ public final class ProcessUtils {
         LOG.log(DEBUG,
             () -> "Waiting finished after " + millis + " ms. Action " + (result ? "succeeded." : "timed out."));
         return result;
+    }
+
+
+    private static Duration calculatePollingInterval(final Duration interval, final Instant deadline, Instant current) {
+        if (interval == null) {
+            return null;
+        }
+        if (deadline == null) {
+            return interval;
+        }
+        if (current == null) {
+            current = Instant.now();
+        }
+        if (current.isAfter(deadline)) {
+            return null;
+        }
+        return current.plus(interval).isBefore(deadline) ? interval : Duration.between(current, deadline);
     }
 
 
--- a/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/ServerLifeSignChecker.java
+++ b/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/ServerLifeSignChecker.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Contributors to the Eclipse Foundation.
+ * Copyright (c) 2025, 2026 Contributors to the Eclipse Foundation.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -16,6 +16,7 @@
 
 package com.sun.enterprise.admin.servermgmt.cli;
 
+import com.sun.enterprise.admin.cli.CLIConstants;
 import com.sun.enterprise.admin.servermgmt.util.CommandAction;
 import com.sun.enterprise.universal.process.ProcessUtils;
 import com.sun.enterprise.util.HostAndPort;
@@ -126,7 +127,7 @@ public class ServerLifeSignChecker {
             }
             return true;
         };
-        return ProcessUtils.waitFor(signOfFinishedStartup, timeout, verbose);
+        return ProcessUtils.waitFor(signOfFinishedStartup, timeout, CLIConstants.WAIT_FOR_DAS_POLLING_INTERVAL, verbose);
     }
 
     private ServerLifeSigns createTimeoutReport(final ServerLifeSigns signs) {
--- a/nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/StartInstanceCommand.java
+++ b/nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/StartInstanceCommand.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, 2025 Contributors to the Eclipse Foundation
+ * Copyright (c) 2023, 2026 Contributors to the Eclipse Foundation
  * Copyright (c) 2008, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -70,6 +70,7 @@ import static org.glassfish.embeddable.GlassFishVariable.TIMEOUT_START_SERVER;
         })
 })
 public class StartInstanceCommand implements AdminCommand {
+    private static final Duration PORT_CHECK_SLEEP = Duration.ofMillis(100);
     @Inject
     ServiceLocator locator;
 
@@ -197,7 +198,7 @@ public class StartInstanceCommand implements AdminCommand {
 
         if (report.getActionExitCode() == ActionReport.ExitCode.SUCCESS) {
             // Make sure instance is listening
-            if (ProcessUtils.waitFor(instance::isListeningOnAdminPort, getTimeout(timeout), false)) {
+            if (ProcessUtils.waitFor(instance::isListeningOnAdminPort, getTimeout(timeout), PORT_CHECK_SLEEP, false)) {
                 return;
             }
             report.setMessage(Strings.get("start.instance.timeout", instanceName));
--- a/nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/StopInstanceCommand.java
+++ b/nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/StopInstanceCommand.java
@@ -86,7 +86,7 @@ import static org.glassfish.embeddable.GlassFishVariable.TIMEOUT_STOP_SERVER;
         })
 })
 public class StopInstanceCommand extends StopServer implements AdminCommand {
-
+    private static final Duration PORT_CHECK_SLEEP = Duration.ofMillis(100);
     @Inject
     private ServiceLocator locator;
     @Inject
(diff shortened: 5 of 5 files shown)


OndroMih (approved):

nucleus/common/common-util/src/main/java/com/sun/enterprise/universal/process/ProcessUtils.java
```diff
@@ -212,9 +212,20 @@ public static boolean waitWhileListening(HostAndPort endpoint, Duration timeout,
...
+     */
+    public static boolean isListening(final HostAndPort endpoint, final boolean graceful) {
         try (Socket socket = new Socket()) {
-            // Force RST on close
-            socket.setSoLinger(true, 0);
+            if (!graceful) {
+                // Force RST on close
+                socket.setSoLinger(true, 0);
+            }
             socket.connect(endpoint.toInetSocketAddress(), SOCKET_CONNECT_TIMEOUT);
```
avpinchuk, line 218: What will happen if we set some timeout after connection and save SO_LINGER option?

The SO_LINGER option was introduced to close a bunch of ephemeral ports in the TIME_WAIT state.

avpinchuk: See
* #25811

dmatej:
The problem is that without the RST those connections will not be closed immediately and it brings issues with restarts and number of allocated ports.
There are at least 5 calls of the original method, you replaced just 3 of them. Why? If we don't want to use the RST signal, we should remove it completely.

However looking at https://ipwithease.com/tcp-rst-flag/ I would say that the original was correct, and server should handle that. Probably log it just as a trace/finest level as it may happen any time with any other client.

> If there is a router doing NAT, especially a low end router with few resources, it will age the oldest TCP sessions first. To do this it sets the RST flag in the packet that effectively tells the receiving station to (very ungracefully) close the connection. this is done to save resources.

> Some firewalls do that if a connection is idle for x number of minutes. Some ISPs set their routers to do that for various reasons as well.

Without that I was able to exhaust some limits.

So I believe the server should just ignore this error.

I have yet one thought ... in the past Java cached connections internally, now it should be possible to close them explicitly. Then it might be possible that we would not need to use the `socket.setSoLinger(true, 0);` if we would always disconnected. I will check that ... I have to remember how I was counting those connections ...

dmatej:
I asked Claude, it confirmed both mentioned problems - Mac issue and TIMED_WAITING without RST and offered this solution:
```Java
public boolean isReachable(Endpoint endpoint) {
    try (SocketChannel channel = SocketChannel.open();
         Selector selector = Selector.open()) {
        
        channel.configureBlocking(false);
        channel.register(selector, SelectionKey.OP_CONNECT);
        channel.connect(endpoint.toInetSocketAddress());

        if (selector.select(SOCKET_CONNECT_TIMEOUT) == 0) {
            return false; // timed out, no established connection → no TIME_WAIT
        }

        SelectionKey key = selector.selectedKeys().iterator().next();
        if (key.isConnectable()) {
            return channel.finishConnect(); // true = success
        }
        return false;
    } catch (IOException e) {
        LOG.log(TRACE, "Socket check to " + endpoint + " failed", e);
        return false;
    }
}
```
That is very close to the original code I replaced in 55797c6bc5d53278de02f49c489a3d20e9494716 which probably did not work properly too as it was missing the `key.isConnectable` part. Now the question is if the Claude's code is completely correct.

dmatej (changes requested):
The problem is that "graceful" disconnect leads to thousands of connections in a state TIMED_WAITING. 
I will try the Claude's recommendation, perhaps it could resolve both problems.

dmatej:
Testing:
```bash
watch -n 1 "ss -tan state time-wait | grep ':4848' | wc -l"
```
```bash
# Wait until watch goes to 0, then restart the domain and watch the watch. 
asadmin restart-domain
```

Claude's solution: 900-3000 connections.
Your solution (no RST): 1022-2500
RST with 0 timeout: 2
RST with 1s timeout: 107

Finally Claude recommended just to reduce the logging level in Grizzly. 

This "vector" causes the problem:
`ServerLifeSignChecker.waitFor -> isListeningOnAnyEndpoint(adminEndpointsSupplier.get()) -> isListening(HostAndPort endpoint)`

carryel:
I just checked the past history. Thanks!

> There are at least 5 calls to the original method; you replaced just 3 of them. Why? If we don't want to use the RST signal, we should remove it completely.

Considering potential side-effects I might not be aware of, I only changed the minimum number of calls (the ones I was able to verify).

> Claude's solution: 900-3000 connections.
> Your solution (no RST): 1022-2500
> RST with 0 timeout: 2
> RST with 1s timeout: 107

Now that I know about that part as well, I’ll look into it a bit more with you.

carryel:
> What will happen if we set a timeout after the connection and save the SO_LINGER option?

I will check this as well. It is presumed that doing this will resolve the original issue to some extent. This is because the problem occurred when a connection was closed in an incomplete state quickly due to a large number of connection requests over a very short period, so it is estimated that even a slight delay will prevent the issue.

PS)
I just tested it.

`socket.setSoLinger(true, 1);`

The `java.net.SocketException: Invalid argument` error does not disappear, but it is effective as expected, and occurs intermittently in small amounts. However, the large number of TIME_WAIT instances has not improved.

dmatej:
I nearly forgot about it, but @avpinchuk's note and link refreshed my memory. After all I would really just reduced the log level of the warning, because any client/proxy/firewall can do that too (close without FIN) and Grizzly doesn't have any options to do something except throwing the connection away.

Me:
> Is there some way to reduce the TIMED_WAITING time? It seems it is around a minute on my system.

Claude:
> Yes, but with an important caveat — TIME_WAIT duration is hardcoded in the Linux kernel as 2 × MSL (Maximum Segment Lifetime) = 60 seconds and is not directly configurable via sysctl.

carryel:
I also confirmed that an unnecessarily large number of TIME_WAIT states were generated when running commands like `restart-domain` locally. (Approximately 100 to 300 connections, matching the number of connections reproduced here.)

- https://github.com/eclipse-ee4j/glassfish-grizzly/issues/2285#issuecomment-4204744402

Shifting my perspective slightly, I would like to discuss the issue regarding the excessive TIME_WAIT states mentioned in the following issue.

> 1. Repeated port listening checks for short periods.

If improvements are made in the following area as well, there will be no unnecessary (meaningless) connection requests, and the resulting TIME_WAIT requests will also decrease.

https://github.com/eclipse-ee4j/glassfish/blob/b9e2987822b0753da91b34ecebc3dc4ab6128fb5/nucleus/common/common-util/src/main/java/com/sun/enterprise/universal/process/ProcessUtils.java#L281-L286

I thought it would be good if the while loop and `Thread.onSpinWait()` were improved for I/O operations such as socket connect. This is because, practically speaking, `Thread.onSpinWait()` is a very short duration, resulting in many unnecessary connect requests within that short period.

Even when only a 100ms sleep was applied for a brief test, excessive connection requests and the resulting TIME_WAIT did not occur.
Of course, there is likely a better way. I just added a sleep just for verification purposes.

dmatej:
That is exactly what I would like to avoid - to sleep for no reason. We don't need the connection, we really want to close it asap, not to rely on implicit cleanup made by OS after 60 seconds.

EDIT: yet one idea, what if we would use UDP for that? I will try. 

EDIT2: That looks quite well, 17. I will try to create a PR so you could test it on Mac - I am not sure how Grizzly can react on UDP ...
Also, for remote connections the UDP might be forbidden, so probably we will need some switcher. However these restarts are usually local.

EDIT3: UDP has new problem - the socket timeout is always there. So I discussed that with Claude again:

> Hmm, the problem here is that the SOCKET_CONNECT_TIMEOUT is set to 1 second, and it really times out after one second. On the other hand shorter timeout might be  problem on really slow networks ....

Claude:
> <p class="font-claude-response-body break-words whitespace-normal leading-[1.7]">Exactly — you've hit the fundamental asymmetry of UDP for this use case:</p>
Scenario | TCP | UDP |
-- | -- | -- |
Port closed | Instant ConnectionRefused | Instant PortUnreachableException |
Port open | Instant success | Must wait for timeout |

> <p class="font-claude-response-body break-words whitespace-normal leading-[1.7]">With TCP you get an <strong>immediate answer in both cases</strong>. With UDP you only get an immediate answer when the port is closed — open requires waiting for silence.</p>
> <h2 class="text-text-100 mt-3 -mb-1 text-[1.125rem] font-bold">So UDP Doesn't Actually Help You Here</h2>
> <p class="font-claude-response-body break-words whitespace-normal leading-[1.7]">Your polling loop is:</p>
> <div role="group" aria-label="Code" tabindex="0" class="relative group/copy bg-bg-000/50 border-0.5 border-border-400 rounded-lg focus:outline-none focus-visible:ring-2 focus-visible:ring-accent-100"><div class="sticky opacity-0 group-hover/copy:opacity-100 group-focus-within/copy:opacity-100 top-2 py-2 h-12 w-0 float-right"><div class="absolute right-0 h-8 px-2 items-center inline-flex z-10"><button class="inline-flex
>   items-center
>   justify-center
>   relative
>   isolate
>   shrink-0
>   can-focus
>   select-none
>   disabled:pointer-events-none
>   disabled:opacity-50
>   disabled:shadow-none
>   disabled:drop-shadow-none border-transparent
>           transition
>           font-base
>           duration-300
>           ease-[cubic-bezier(0.165,0.85,0.45,1)] h-8 w-8 rounded-md backdrop-blur-md _fill_56vq7_9 _ghost_56vq7_96" type="button" aria-label="Copy to clipboard" data-state="closed"><div class="relative"><div class="transition-all opacity-100 scale-100" style="width: 20px; height: 20px; display: flex; align-items: center; justify-content: center;"><svg width="20" height="20" viewBox="0 0 20 20" fill="currentColor" xmlns="http://www.w3.org/2000/svg" class="transition-all opacity-100 scale-100" aria-hidden="true" style="flex-shrink: 0;"><path d="M12.5 3A1.5 1.5 0 0 1 14 4.5V6h1.5A1.5 1.5 0 0 1 17 7.5v8a1.5 1.5 0 0 1-1.5 1.5h-8A1.5 1.5 0 0 1 6 15.5V14H4.5A1.5 1.5 0 0 1 3 12.5v-8A1.5 1.5 0 0 1 4.5 3zm1.5 9.5a1.5 1.5 0 0 1-1.5 1.5H7v1.5a.5.5 0 0 0 .5.5h8a.5.5 0 0 0 .5-.5v-8a.5.5 0 0 0-.5-.5H14zM4.5 4a.5.5 0 0 0-.5.5v8a.5.5 0 0 0 .5.5h8a.5.5 0 0 0 .5-.5v-8a.5.5 0 0 0-.5-.5z"></path></svg></div><div class="absolute inset-0 flex items-center justify-center"><div class="transition-all opacity-0 scale-50" style="width: 20px; height: 20px; display: flex; align-items: center; justify-content: center;"><svg width="20" height="20" viewBox="0 0 20 20" fill="currentColor" xmlns="http://www.w3.org/2000/svg" class="transition-all opacity-0 scale-50" aria-hidden="true" style="flex-shrink: 0;"><path d="M15.188 5.11a.5.5 0 0 1 .752.626l-.056.084-7.5 9a.5.5 0 0 1-.738.033l-3.5-3.5-.064-.078a.501.501 0 0 1 .693-.693l.078.064 3.113 3.113 7.15-8.58z"></path></svg></div></div></div></button></div></div><div class="overflow-x-auto"><pre class="code-block__code !my-0 !rounded-lg !text-sm !leading-relaxed p-3.5" style="color: rgb(234, 236, 240); background: transparent; font-family: var(--font-mono);"><code style="color: rgb(234, 236, 240); background: transparent; font-family: var(--font-mono); white-space: pre-wrap;"><span><span>server down → server starting → server up</span></span></code></pre></div></div>
> <p class="font-claude-response-body break-words whitespace-normal leading-[1.7]">The slow case you care about is <strong>detecting server up as fast as possible</strong>. UDP makes that <em>worse</em> — you have to wait the full timeout every time the server is up, which is the hot path you're polling for.</p>
> <h2 class="text-text-100 mt-3 -mb-1 text-[1.125rem] font-bold">Back to TCP — The Real Options</h2>
> <p class="font-claude-response-body break-words whitespace-normal leading-[1.7]">You've essentially proven empirically that:</p>
> <ol class="[li_&amp;]:mb-0 [li_&amp;]:mt-1 [li_&amp;]:gap-1 [&amp;:not(:last-child)_ul]:pb-1 [&amp;:not(:last-child)_ol]:pb-1 list-decimal flex flex-col gap-1 pl-8 mb-3">
> <li class="whitespace-normal break-words pl-2"><strong>RST linger=0</strong> — perfect behavior, Grizzly warns on Mac</li>
> <li class="whitespace-normal break-words pl-2"><strong>Everything else</strong> — either slow or accumulates connections</li>
> </ol>
> <p class="font-claude-response-body break-words whitespace-normal leading-[1.7]">The options remaining are:</p>
> <ul class="[li_&amp;]:mb-0 [li_&amp;]:mt-1 [li_&amp;]:gap-1 [&amp;:not(:last-child)_ul]:pb-1 [&amp;:not(:last-child)_ol]:pb-1 list-disc flex flex-col gap-1 pl-8 mb-3">
> <li class="whitespace-normal break-words pl-2"><strong>Suppress Grizzly's warning</strong> — simplest, no code change</li>
> <li class="whitespace-normal break-words pl-2"><strong>Contribute a fix to Grizzly</strong> — handle the race condition where socket is closed before options are set</li>
> <li class="whitespace-normal break-words pl-2"><strong>Accept the warnings</strong> — they're cosmetic, not functional</li>
> </ul>
> <p class="font-claude-response-body break-words whitespace-normal leading-[1.7]">Is the Grizzly warning actually causing any real problem beyond noise in the logs? Because RST linger=0 with 2 connections is genuinely the correct solution for what you're doing.</p>
> 

The point is that those warnings really can be just suppressed, they don't mean any warning about disaster closing. Just the channel was unexpectedly closed from the other side while Grizzly tried to use it, and it should just reflect that probably by releasing own resources (if there are some).

carryel:
I am not sure if my intention was conveyed clearly, so I will explain a little further. 

My question was whether there is a reason for checking whether a (local or remote) server is listening to be done urgently within a very short time (about 130 connection attempts on Mac; about 1000 attempts on Windows?). 

I was not saying that the socket should be closed immediately with `RST` or gracefully, regardless of whether the connection succeeds or fails.

I discovered this while investigating why a large number of connect requests were coming in on port 4848 during the process of starting/stopping/restarting the domain.

carryel:
PS) Additionally, I plan to submit a pull request to adjust Grizzly's log level. Since it outputs actual errors as warning logs but ignores them to proceed, I don't see any reason to keep it at the warning level.

carryel:
I have additionally modified the excessive port listening check that occurred for a short duration during the `start-domain` process mentioned above.

Furthermore, considering the issue of unnecessary persistence of TIME_WAIT, I have applied `socket.setSoLinger(true, 1);`.

With these modifications, the TIME_WAIT level increased to an understandable level, and no errors occurred in Grizzly on my macOS.

Would it be possible to verify this patch version in a Windows environment, where TIME_WAIT occurs frequently?

nucleus/admin/cli/src/main/java/com/sun/enterprise/admin/cli/CLIConstants.java
```diff
@@ -36,6 +36,7 @@ public final class CLIConstants {
     public static final String EOL = System.lineSeparator();
 
     public static final Duration WAIT_FOR_DAS_TIME_MS = getEnv(TIMEOUT_START_SERVER);
+    public static final Duration WAIT_FOR_DAS_POLLING_INTERVAL_MS = Duration.ofMillis(100);
```
pzygielo: I see the old ones above and below - have `_MS` as well. But why have it in new one? Isn't this just _some_ `Duration`? Is the fact that it's initialized from millis important later?

carryel: Correct. I don't think it is necessary to specify MS.
I just included it to comply with the convention above.

However, the WAIT_FOR_DAS_TIME_MS above is actually incorrect as well. In reality, the Duration for WAIT_FOR_DAS_TIME_MS is generated using seconds.

https://github.com/eclipse-ee4j/glassfish/blob/b9e2987822b0753da91b34ecebc3dc4ab6128fb5/nucleus/admin/cli/src/main/java/com/sun/enterprise/admin/cli/CLIConstants.java#L86

I will correct the new one.

dmatej:
> My question was whether there is a reason for checking whether a (local or remote) server is listening to be done urgently within a very short time (about 130 connection attempts on Mac; about 1000 attempts on Windows?).

Because we want to make restarts fast as much as possible.
Reason 1: Every waiting counts, in tests we do a lot of restarts.
Reason 2: No additional latency here led also to a lot of related discoveries of resource leaks, timing issues etc.

Reason 3: Perhaps Grizzly should react to these events perfectly too. I have yet another half baked PR on Corba-ORB project with similar issues, race conditions, where listeners don't handle start/stop/restart too well. I believe Grizzly is in better shape.

On production GF/Grizzly has to handle thousands of clients in parallel with very different use cases, behaviors, so this is not something what should cause any problems. With any latency and graceful FIN we would leave hundreds open ports, which doesn't make any sense to me. It seems as a step back to me.

avpinchuk:
> * **Contribute a fix to Grizzly** — handle the race condition where socket is closed before options are set

Unfortunately, Java does not have a reliable way to determine whether a socket has been closed or has received an RST.

We can't get the OS EINVAL error code without using native tools. And using native tools in the fast NIO framework is very expensive. In Windows it's even worse.

dmatej:
Sorry for the conflicting changes made in #25972 , the Constants class was removed. 

Yet another idea - create some PORT_CHECK_SLEEP (maybe a better name) environment option. Unset by default, so it would be so aggressive as now, but it would be possible to slow it down.

Another option is to open own server socket and wait for callback; because we have no problems with shutdown, the problem here is the detection that the server is listening to following asadmin+REST commands. That is maybe the correct solution. So we would not check if the 4848 is listening, but starting server would report itself at the point when it would cross some boundary of the startup process. The socket would be open just for loopback host as the server must be always started by the local asadmin command... then we would check the port just once and we don't even need the RST.

carryel:
I have resubmitted the PR after resolving the conflict and modifying the variable names as suggested.

However, unfortunately, I was unable to implement the environment option method you suggested. This is because the existing logic references various files such as `GlassfishVariable.java`, `SystemPropertyConstants.java`, and `CLIConstants.java`, and some of them are deprecated, so I was unsure of the proper approach.

OndroMih:
We need to rethink how we address this - I moved this PR to draft to indicate that it's not to be merged for now.

dmatej:
I plan to work on this in 8.0.4 - the callback solution instead of "spamming" the server.

PR #25981, opened 2026-04-15
Fix MP Config packaging

OndroMih:
Refactors Embedded GF All distribution module to avoid duplicating the patched Helidon class.

Fixes Embedded GF Web, which didn't prevent Helidon's service file occasionally win over the patched file.

There's still a warning from MP Config OSGi bundle build about split packages. I don't know how to fix it. Our bundle overrides/patches a few classes in the packages that those 2 modules export and bundle plugin complains about split packages provided by both our module and the Helidon artifacts. I tried `-split-package:=merge-first` but it didn't help. Probably the only way to fix it is to remove those 2 modules from the manifest and copy over classes from them to our bundle and then overwrite them with our classes. But I'm not going to do it in this PR.

### DIFF
--- a/appserver/extras/embedded/all/src/assembly/package.xml
+++ b/appserver/extras/embedded/all/src/assembly/package.xml
@@ -9,6 +9,11 @@
     <includeBaseDirectory>false</includeBaseDirectory>
 
     <dependencySets>
+        <!--
+            microprofile-config contains GlassFish-specific overrides of Helidon resources. 
+            It must be unpacked before all other jars so its classes take precedence.
+            Note that the default behavior of assembly plugin is that it skips duplicates, first occurence wins.
+        -->
         <dependencySet>
             <unpack>true</unpack>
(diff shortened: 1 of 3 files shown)


dmatej (approved):
Not sure which strategy is the best, but in my experience is really the best to maintain own service files perfectly maintained for the JAR we distribute.
On the other hand internal service files which we "take as they were supplied" is not worth to maintain on our own, so ... it depends.

OndroMih:
My strategy usually is to use the 3rd party service files and change them only if needed. Thrid party maintainers usually maintain them better than us. But sometimes we need to make changes.

Here I didn’t change anything, I just removed our deuplicate service file and copied the same service file from another GlassFish module. The microprofile-config module as a GlssFish module and already contains the service file.

The end result is the same as without this PR, besides the fix for Web distribution. For the All distro it’s just a refactoring.

PR #25982, opened 2026-04-16
Update several versions in pom.xml

arjantijms:


nucleus/parent/pom.xml
```diff
@@ -202,7 +202,7 @@
         <tyrus.version>2.3.0-M1</tyrus.version>
 
         <!-- Jakarta Concurrency -->
-        <jakarta.concurrency-api.version>3.2.0-M3</jakarta.concurrency-api.version>
+        <jakarta.concurrency-api.version>3.2.0-M2</jakarta.concurrency-api.version>
```
pzygielo, line 205: What's the justification for downgrade?

pzygielo: I see - no such thing as M3 there (in Central).

PR #25997, opened 2026-04-19
Extract console-mail-plugin and add button "send test email"

bvfalcon:
This PR contains several adjustments for Admin Console application:

1. mail resource configuration pages are extracted into separate `console-mail-plugin`.
This point makes features No 2-3 from this list available for Glassfish version 7.0 and later
2. all entries of "JavaMail" are renamed to "Jakarta Mail": labels, help pages, links to jakarta.ee instead of oracle.com
3. new feature: added new button "send test email" to check correctness of mail resource configuration
Sends test message from email "from" to email "from" (to himself) using configured server.


OndroMih:
Hi @bvfalcon , is there a reason why this is raised against the 7.x branch and not against the main branch?

We usually first accept changes into the main branch, and then backport them to older branches if the branch is still supported. The reason is that the main branch receives regular attention and is regularly released, while the other branches are released when time allows or an OmniFish customer requests it.

appserver/admingui/mail-plugin/pom.xml
```diff
@@ -0,0 +1,103 @@
...
+                        <supportedProjectType>glassfish-jar</supportedProjectType>
+                        <supportedProjectType>jar</supportedProjectType>
+                        <supportedProjectType>bundle</supportedProjectType>
+                    </supportedProjectTypes>
+                    <instructions>
+                        <!-- _include does not work. See Felix-620.
+                        <_include>osgi.bundle</_include>
+                        -->
+                        <Export-Package>org.glassfish.javamail.admingui</Export-Package>
+                    </instructions>
```
OndroMih, line 80: What is this about? There's no osgi.bundle file, is there? And there's also no package org.glassfish.javamail.admingui to export, or is there?

appserver/admingui/mail-plugin/src/main/resources/jakartaMailResources.jsf
OndroMih: Shouldn't this be "jakartaMailResources" now?

```suggestion
    guiTemplateOnLoad="admingui.nav.refreshTree('treeForm:tree:resources:jakartaMailResources');
```


Without this, the sidebar isn't updated after a new mail resource is created and the list of mail resources in the sidebar doesn't show the new resource until page reload: 

Image

appserver/admingui/mail-plugin/src/main/resources/mailResourceNew.jsf
OndroMih: Should probably be:
```suggestion
"    <script type="text/javascript">admingui.nav.selectTreeNodeById(admingui.nav.TREE_ID+":resources:jakartaMailResources");</script>
```

appserver/admingui/mail-plugin/src/main/resources/jakartaMailResourceLink.jsf
```diff
@@ -0,0 +1,30 @@
...
+
+<!initPage
+    setResourceBundle(key="i18njmail" bundle="org.glassfish.jmail.admingui.Strings")
+/>
+
+<sun:property>
+    <sun:hyperlink 
+        toolTip="$resource{i18njmail.tree.javaMailSessions.tooltip}"
+        url="/jmail/jakartaMailResources.jsf" >
+        <sun:image url="/resource/images/javaMailSession.gif" />
```
OndroMih: Other places reference the image via `/resource/jmail/images/javaMailSession.gif`, is this correct to use `/resource/images/javaMailSession.gif` (without `jmail`) here?

OndroMih (changes requested):

OndroMih:
I tested the new "Send test email" button against a fake SMTP server (https://hub.docker.com/r/haravich/fake-smtp-server) and it works well.

I raised a few comments because I found some little bugs here and there.

And I still wonder whether this should rather be raised against the main branch and then backported to 7.x to follow our usual workflow. What do you think, @dmatej ?

dmatej:
> And I still wonder whether this should rather be raised against the main branch and then backported to 7.x to follow our usual workflow. What do you think, @dmatej ?

It would be better, then we would backport it, however if it doesn't become a history of many commits and especially conflicts, I can exclusively do that the opposite way too.

appserver/admingui/mail-plugin/src/main/resources/jakartaMailResourceLink.jsf
bvfalcon: Yes, I agree. gif-file location was changed, but link was not. I change the link too with next commit

appserver/admingui/mail-plugin/pom.xml
bvfalcon: I suppose, correct package name is `org.glassfish.jmail.admingui` -- package with localization resources. I change it with next commit too

pzygielo:
Why include empty commit https://github.com/eclipse-ee4j/glassfish/pull/25997/commits/a3d04a2b915df6e58b57b843b7b6922e3e7cb4c4?

bvfalcon:
> Why include empty commit [a3d04a2](https://github.com/eclipse-ee4j/glassfish/commit/a3d04a2b915df6e58b57b843b7b6922e3e7cb4c4)?

artifacts of git merge. solved

OndroMih (approved):
Thanks!

OndroMih:
I'll merge this even without the Jenkins build passing - the changes are only in Admin Console and that's not even covered by tests.

OndroMih:
I created a PR to port this to the main branch: https://github.com/eclipse-ee4j/glassfish/pull/26003

PR #25998, opened 2026-04-21
Updated Ant, Schema2beans to 2.9.0, dbschema to 2.9.0, Grizzly to 5.0.1, Grizzly NPN to 2.0.1, HK2 to 4.0.1

dmatej:
* Fixes #25870 
* Fixes also grizzly-bom issue mentioned [here](https://github.com/eclipse-ee4j/glassfish/pull/26003#issuecomment-4341923502), which blocks all builds.
* See https://github.com/eclipse-ee4j/glassfish-repackaged/pull/30 and https://github.com/eclipse-ee4j/glassfish-repackaged/pull/31

Upgrades:

* HK2 from 4.0.0 to 4.0.1
* Grizzly from 5.0.0 to 5.0.1
* Grizzly NPN from 2.0.0 to 2.0.1
* Netbeans DB Schema from 6.7 (old versioning) to 2.9.0 (new versioning under Apache Foundation)
* Netbeans schema2beans from 6.7 (old versioning) to 2.9.0 (new versioning under Apache Foundation)

Other changes

* Added logs to see problems at least on FINEST level
* Removed Ant dependency, we can use just one, Ant OSGi 
* Reconfigured maven-antrun-plugin
* Some code cleanup made when I tried to find the error or the place where it can be reported.

PR #26002, opened 2026-04-25
CMP Refactoring

dmatej:
- Logging simplified to today's pattern using System.Logger
- generics
- `@Overrides`
- Some formatting to make the code reasonably readable
- Removal of NOI18N comments
- Added System.exit(1) to Main if there is an exception.
- Removed DumpMapping class - unused

Notes
- Those cache implementations are problematic, I doubt they work perfectly. I added generics, but that made visible some issues. There were also already some comments related, but as this part of the technology is obsoleted, I would not burn more time with that.
- Old Ant based tests are covering these features. Testing:
```
./rm-generated.sh && mvn clean install -T4C -Pfastest && ./runtests.sh ejb_group_2
./rm-generated.sh && mvn clean install -T4C -Pfastest && ./runtests.sh connector_group_4
./rm-generated.sh && mvn clean install -T4C -Pfastest && ./runtests.sh connector_group_1
```


arjantijms:
CMP being arguably THE more obscure thing ever to come out of J2EE (beating e.g. app client by a large margin), I wonder if we should actually need to keep this.

avpinchuk:
Moreover, the CMP JDO is not even fully compliant with the old JDO specification. It looks like there is no actual reference manual for it.

dmatej:
At least it will stop slowing us down for a while :)
Pushed again - I forgot to remove one resource bundle instance together with the file loaded. It is interesting how it broke completely the synchronization, because it just logged the exception and used transactions without semaphores - it is so old that it uses own implementation of a semaphore, it has two - one is SemaphoreImpl, another is NullSemaphore which is used as a backup, which was really terrible idea :D

Note that most of the code was written between 1997 and 2001.

Originally I made this under #25998 , but new libraries have some internally incompatible changes and I could not find what was wrong. So I was digging and doing some maintenance, and the amount of changes grown quickly. Nothing too complicated, but I still did not see what I should do with the test failure.

Then I was stuck so I decided to move the maintenance changes to own PR, fix whatever mistake I made, and then rebase the original PR to this.

Fortunately the coverage is not so bad here, Sun was rich company at that time :-)

dmatej:
Note about the Jenkins link: I tried to abort `#2`, but jenkins aborted `#3` instead, so I deleted it and restarted the build again.

arjantijms (approved):

PR #26029, opened 2026-05-11
Scope java:global/ resource pools per declaring app

BalusC:
ComponentEnvManagerImpl.getResourceId returned "" for ScopeType.GLOBAL, so deriveResourceName produced an un-prefixed pool name and every app declaring the same java:global/X (e.g. @DataSourceDefinition) shared one physical connection pool. Connections opened against the first app's classloader stayed in the shared pool after that app undeployed, blowing up the next app's lookups with "ClassLoader is not in expected state" wrapped as SQLException: GeneralError.

Return the application name for GLOBAL too, matching the behaviour already in place for APP scope. Each app now gets its own physical pool keyed under __SYSTEM/.../<appName>/java:global/X — symmetric with __SYSTEM/.../<appName>/java:app/X.

The empty fall-through has been there since the helper was introduced in 1230a81f8f (2012-07-25, Java EE 7 Mail Resource Annotation); GLOBAL was simply omitted from the COMPONENT/MODULE/APP cascade. No deliberate semantics to preserve.

Discovered while debugging recurring "GeneralError" failures in jakartaee/security#368 (Jakarta Security TCK
arquillian-glassfish-server-pool migration), where multiple app-db-* modules each declare @DataSourceDefinition(name = "java:global/securityAPIDB", ...).

### DIFF
--- a/appserver/common/container-common/src/main/java/com/sun/enterprise/container/common/impl/ComponentEnvManagerImpl.java
+++ b/appserver/common/container-common/src/main/java/com/sun/enterprise/container/common/impl/ComponentEnvManagerImpl.java
@@ -314,10 +314,13 @@ public class ComponentEnvManagerImpl implements ComponentEnvManager {
         if (dependencyAppliesToScope(desc, ScopeType.MODULE)) {
             return getApplicationName(env) + "/" + getModuleName(env);
         }
-        if (dependencyAppliesToScope(desc, ScopeType.APP)) {
-            return getApplicationName(env);
-        }
-        return "";
+        // APP and GLOBAL scopes: scope the underlying physical resource (pool
+        // name, etc.) to the declaring app. For java:global/ this is what
+        // prevents two apps declaring the same name from sharing one physical
+        // pool — which would let connections opened against the first app's
+        // classloader linger after that app undeploys, blowing up the next
+        // app's lookups with "ClassLoader is not in expected state".
+        return getApplicationName(env);
     }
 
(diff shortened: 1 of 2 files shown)


dmatej (approved):

OndroMih:
Tests in ClusterITest.java fail, I'm not sure whether it's an infra issue or a problem introduced by these changes.

OndroMih:
The test failures are probably because of infra issues, the same tests failed here: https://github.com/eclipse-ee4j/glassfish/pull/26026

I restarted the test suite.

dmatej:
Now it seems it is some dependency issue (missing nosql class), seems @arjantijms recently fixed something relevant, so I rebased this PR to the latest 9.0 now.

dmatej:
The problem is that 9.0 depends on a snapshot - which changes a lot now:
`<jnosql.version>1.2.0-SNAPSHOT</jnosql.version>`

arjantijms:
Hmmm, the "same" snapshot does pass locally. We should probably pin a version though.

The snapshot was needed since there were no milestone releases for JNoSql extensions. Maybe @OndroMih can do such milestone release?

dmatej:
> Hmmm, the "same" snapshot does pass locally. We should probably pin a version though.
> 
> The snapshot was needed since there were no milestone releases for JNoSql extensions. Maybe @OndroMih can do such milestone release?

No, it doesn't, failed on my computer too.

arjantijms (approved):

PR #26030, opened 2026-05-11
Fix NetUtilsIT to detect the actual container IP at runtime

BalusC:
getContainerHostIP() was hard-coded to the default Docker bridge prefix "172.17.". On machines where Docker assigns containers from a different subnet (e.g. 172.26.0.0/16, which happens when 172.17.0.0/16 is already in use by a VPN or pre-existing network), every assertion using stringContainsInOrder(getContainerHostIP(), ...) fails because the expected substring never appears in the actual IP string.

Replace the hard-coded prefix with a live lookup via testcontainers' NetworkSettings.getIpAddress() after the container has started, falling back to iterating the networks map if the primary IP is empty (the original FIXME's complaint about getGlobalIPv6Address() returning blank).

### DIFF
--- a/nucleus/common/common-util/src/test/java/com/sun/enterprise/util/net/NetUtilsIT.java
+++ b/nucleus/common/common-util/src/test/java/com/sun/enterprise/util/net/NetUtilsIT.java
@@ -317,9 +317,16 @@ public class NetUtilsIT {
     }
 
     private String getContainerHostIP() {
-        // FIXME: The following call is unstable, randomly returns an empty string.
-//        return container.getContainerInfo().getNetworkSettings().getGlobalIPv6Address();
-        return "172.17.";
+        var networkSettings = container.getContainerInfo().getNetworkSettings();
+        String ip = networkSettings.getIpAddress();
+        if (ip != null && !ip.isEmpty()) {
+            return ip;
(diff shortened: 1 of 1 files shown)


dmatej (approved):
The FIXME was added in f63f085b9fe79eb39e5020d57502337fe4a8d20e. Seems originally I wanted ipv6, not 4, and then I have up. However maybe I was just trying some shotgun development with the GlobalIPv6Address, I tried many combinations of settings just to confirm it behaves well.

However choosing "any" address is still better than the hardcoded.

BalusC:
The hardcoded return string was IPv4 not IPv6. So I went for IPv4.

Let me know if you'd like to see an IPv6 solution.

dmatej:
> The hardcoded return string was IPv4 not IPv6. So I went for IPv4.
> 
> Let me know if you'd like to see an IPv6 solution.

I believe it is alright, I believe at the point I wrote that comment with FIXME I did not know yet that there are special rules around loopback vs ipv4 vs ipv6 vs local host vs localhost, and all can be somehow touched by the etc host content ordering, every detail matters. 
I remember yet one experience, that the test containers/docker just appends the hosts file. Logical, but good to know :-)

PR #26031, opened 2026-05-11
Generate the embedded HTTPS keystore at runtime so HTTPS works without shipping a private key

BalusC:
Commit 6b062d0d27 ("Removed nonexisting keystore.jks and cacerts.jks", Jun 2025) deliberately stopped shipping the s1as keystore inside kernel.jar / glassfish-embedded-all.jar so the published artifacts would not leak a publicly-known private key. That was the right call, but no runtime replacement was wired up: `EmbeddedGlassFishRuntime.createTempInstanceRoot()` now creates an instance root with no keystore at all. The embedded domain.xml still declares `-Djavax.net.ssl.keyStore=${com.sun.aas.instanceRoot}/config/keystore.p12`, but those `<jvm-options>` only take effect when the JVM is forked - in-process they are ignored. `SecuritySupport` reads `javax.net.ssl.keyStore` as null and falls back to an empty `KeyStore`. JSSE logs `"s1as is not a private key entry"` and the TLS 1.3 handshake aborts with `HANDSHAKE_FAILURE (No available authentication scheme)`; the `HttpURLConnection` client sees `"Remote host terminated the handshake"`.

The CI build masked this for ~10 months because the nine maven-plugin test modules that run before `secureWebApp` each invoke the embedded plugin in the same Maven JVM. The first of them sets `javax.net.ssl.keyStore` to its own pre-generated test keystore, and every subsequent module's GlassFish silently inherits that JVM-wide property - giving `jsptest`, `multipleApps`, and others a usable keystore. `secureWebApp` itself was also a victim: `maven-embedded-glassfish-plugin.setSystemProperties` only sets a key when `System.getProperty` returns null, so `secureWebApp`'s own `systemPropertiesFile` (which would have wired in its `testkeystore.p12`) was silently skipped, and the dangling pointer from the very first module won. Run any of these tests standalone and the handshake fails; run `mvn clean install` from the repo root and `secureWebApp`'s https test fails too, once a later commit on this PR makes the leaked property point at a now-deleted instanceRoot.

All four commits live in `EmbeddedGlassFishRuntime` and preserve the original intent of not shipping a private key:

1. **Generate the keystore at runtime** — `createTempInstanceRoot` now calls `generateEmbeddedKeyPair`, which creates a fresh self-signed `s1as` PKCS12 keystore and matching `cacerts` truststore under `${instanceRoot}/config/` using the existing `KeyTool` utility. A new key pair is generated for every embedded JVM, so `kernel.jar` and `glassfish-embedded-all.jar` still contain no key material - nothing is leaked through the build artifacts. A WARNING log states clearly that the auto-generated cert is for development and testing only, and points to the supported override paths (custom domain.xml, `javax.net.ssl.keyStore` system property, custom instanceRoot). `setEnv` sets `javax.net.ssl.keyStore` and `javax.net.ssl.trustStore` to the generated files with `force=false`, so JSSE in the embedded JVM actually picks them up. Pre-set values from the caller are respected.

2. **Skip generation when the caller already configured one** — if `javax.net.ssl.keyStore` is set on the JVM before `bootstrap()` (typical when wiring in a production cert), `generateEmbeddedKeyPair` bails out early. Avoids ~1s of spurious `keytool` work and suppresses the dev-cert WARNING in scenarios where JSSE never loads the auto-generated keystore.

3. **Log a warning when a configured `javax.net.ssl.keyStore` / `javax.net.ssl.trustStore` is missing** — if the caller pointed JSSE at a path that does not exist, `SecuritySupport.getKeyStore` silently substitutes an empty `KeyStore` and the misconfiguration only surfaces later as an opaque handshake failure. `setEnv` now logs a WARNING up front naming the property and the offending path, mirroring the existing pattern in `SecuritySupport.getKeyStore` that also warns and falls back.

4. **Clear the JSSE properties on dispose** — `setEnv` now returns an `OwnedSslProperties` record listing the paths it set for `javax.net.ssl.keyStore` / `javax.net.ssl.trustStore`. The `onDispose` lambda in `newGlassFish` clears those properties when the instance shuts down, *only* if they still equal the path we set - so a caller's later override is never clobbered. Without this, the embedded instanceRoot is auto-deleted on dispose but the system property keeps pointing at the now-missing `keystore.p12`; the next embedded GlassFish in the same JVM then sees a non-null `javax.net.ssl.keyStore`, `generateEmbeddedKeyPair` bails out (per `#2` above), `setEnv`'s `force=false` preserves the dangling value, and `SecuritySupport` silently substitutes an empty keystore. This is exactly the chain that breaks `secureWebApp` under `mvn clean install` from the repo root, and it's also what restores the plugin's per-module `setSystemProperties` so `secureWebApp`'s `testkeystore.p12` actually gets wired in.

The embedded https-listener remains opt-in (`enabled="false"` in domain.xml); this change only makes it functional when the user does opt in.

### DIFF
--- a/nucleus/core/bootstrap-osgi/src/main/java/org/glassfish/main/boot/embedded/EmbeddedGlassFishRuntime.java
+++ b/nucleus/core/bootstrap-osgi/src/main/java/org/glassfish/main/boot/embedded/EmbeddedGlassFishRuntime.java
@@ -43,6 +43,7 @@ import org.glassfish.embeddable.GlassFish;
 import org.glassfish.embeddable.GlassFishException;
 import org.glassfish.embeddable.GlassFishProperties;
 import org.glassfish.embeddable.GlassFishRuntime;
+import org.glassfish.embeddable.GlassFishVariable;
 import org.glassfish.hk2.api.ServiceLocator;
 import org.glassfish.hk2.utilities.DuplicatePostProcessor;
 import org.glassfish.main.boot.log.LogFacade;
@@ -147,7 +148,12 @@ class EmbeddedGlassFishRuntime extends GlassFishRuntime {
         // The embedded domain.xml declares javax.net.ssl.keyStore/trustStore as jvm-options, but
         // those only take effect when the JVM is forked (standalone GlassFish). For in-process
         // embedded use we set them programmatically here, deferring to the caller's choice if
-        // they've already configured a keystore.
+        // they've already configured a keystore. If the caller set a path but the file does not
+        // exist we log a WARNING up front: SecuritySupport will still substitute an empty
+        // KeyStore (matching its own behaviour), so the handshake will still fail later, but the
+        // log now names the property and the offending path instead of an opaque TLS error.
+        validateConfiguredStorePath(KEYSTORE_FILE);
+        validateConfiguredStorePath(TRUSTSTORE_FILE);
         File embeddedKeyStore = new File(instanceRoot, "config/keystore.p12");
         if (embeddedKeyStore.isFile()) {
             setProperty(KEYSTORE_FILE.getSystemPropertyName(), embeddedKeyStore.getAbsolutePath(), false);
@@ -234,6 +240,16 @@ class EmbeddedGlassFishRuntime extends GlassFishRuntime {
         String autoDelete = gfProps.getProperties().getProperty(AUTO_DELETE, "true");
         gfProps.setProperty(AUTO_DELETE, autoDelete);
         return instanceRoot.getAbsolutePath();
+    }
+
+    private static void validateConfiguredStorePath(GlassFishVariable variable) {
+        String configured = System.getProperty(variable.getSystemPropertyName());
+        if (configured != null && !new File(configured).isFile()) {
+            LOG.log(WARNING, "Configured {0}={1} does not exist or is not a regular file;"
+                + " SecuritySupport will fall back to an empty in-memory keystore and"
+                + " TLS handshakes will fail with an opaque error.",
+                new Object[] {variable.getSystemPropertyName(), configured});
+        }
     }
 
     private static void generateEmbeddedKeyPair(File instanceConfigDir) {
--- a/nucleus/core/bootstrap-osgi/src/main/java/org/glassfish/main/boot/embedded/EmbeddedGlassFishRuntime.java
+++ b/nucleus/core/bootstrap-osgi/src/main/java/org/glassfish/main/boot/embedded/EmbeddedGlassFishRuntime.java
@@ -237,6 +237,11 @@ class EmbeddedGlassFishRuntime extends GlassFishRuntime {
     }
 
     private static void generateEmbeddedKeyPair(File instanceConfigDir) {
+        if (System.getProperty(KEYSTORE_FILE.getSystemPropertyName()) != null) {
+            // The caller already pointed JSSE at a keystore (e.g. via -Djavax.net.ssl.keyStore);
+            // don't generate a dev cert they won't use.
+            return;
+        }
         File keyStore = new File(instanceConfigDir, "keystore.p12");
         File trustStore = new File(instanceConfigDir, "cacerts.p12");
         if (keyStore.isFile() && trustStore.isFile()) {
--- a/nucleus/core/bootstrap-osgi/src/main/java/org/glassfish/main/boot/embedded/EmbeddedGlassFishRuntime.java
+++ b/nucleus/core/bootstrap-osgi/src/main/java/org/glassfish/main/boot/embedded/EmbeddedGlassFishRuntime.java
@@ -93,7 +93,7 @@ class EmbeddedGlassFishRuntime extends GlassFishRuntime {
             properties.putAll(glassFishProperties.getProperties());
 
             final GlassFishProperties gfProps = new GlassFishProperties(properties);
-            setEnv(gfProps);
+            final OwnedSslProperties ownedSslProps = setEnv(gfProps);
 
             final StartupContext startupContext = new StartupContext(gfProps.getProperties());
             final ModulesRegistry modulesRegistry = AbstractFactory.getInstance().createModulesRegistry();
@@ -103,7 +103,10 @@ class EmbeddedGlassFishRuntime extends GlassFishRuntime {
             final ModuleStartup gfKernel = main.findStartupService(modulesRegistry, serviceLocator, null,
                 startupContext);
 
-            final Consumer<GlassFish> onDispose = gf -> glassFishInstances.remove(gfProps.getInstanceRoot());
+            final Consumer<GlassFish> onDispose = gf -> {
+                glassFishInstances.remove(gfProps.getInstanceRoot());
+                ownedSslProps.clearIfStillOwned();
+            };
             final GlassFish glassFish = new AutoDisposableGlassFish(gfKernel, serviceLocator, gfProps, onDispose);
             glassFishInstances.put(gfProps.getInstanceRoot(), glassFish);
 
@@ -134,7 +137,7 @@ class EmbeddedGlassFishRuntime extends GlassFishRuntime {
         }
     }
 
-    private void setEnv(GlassFishProperties gfProps) throws Exception {
+    private OwnedSslProperties setEnv(GlassFishProperties gfProps) throws Exception {
         String instanceRootValue = gfProps.getInstanceRoot();
         if (instanceRootValue == null) {
             instanceRootValue = createTempInstanceRoot(gfProps);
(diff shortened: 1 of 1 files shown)


OndroMih:
To be honest, I don't like this solution much. It forces generation of a certificate even if it's not needed. It would be better to generate the certificate only when https is enabled. It's a waste if Embedded GF is started, e.g. in a Docker container, and HTTPS is not needed.

nucleus/core/bootstrap-osgi/src/main/java/org/glassfish/main/boot/embedded/EmbeddedGlassFishRuntime.java
```diff
@@ -204,6 +264,13 @@ private String createTempInstanceRoot(GlassFishProperties gfProps) throws Except
             if(configFileURI != null) {
                 copy(configFileURI.toURL(), new File(instanceConfigDir, "domain.xml"), true);
             }
+
+            // Generate a self-signed s1as keypair for the embedded HTTPS listener. Each embedded
+            // JVM gets its own fresh private key, so the private key is never shipped in the jar.
+            // The embedded domain.xml expects keystore.p12 + cacerts.p12 in config/, and JSSE
+            // loads them via the system properties set in setEnv. Users replace the cert by
+            // providing a custom instanceRoot or domain.xml.
+            generateEmbeddedKeyPair(instanceConfigDir);
```
OndroMih, line 273: Should be generated only if needed, if all of the following are true:
* if https enabled
* if system properties are not set or point to the default path and the files don't exist

If https is not enabled or system properties point somewhere else, do not generate a key pair.

nucleus/core/bootstrap-osgi/src/main/java/org/glassfish/main/boot/embedded/EmbeddedGlassFishRuntime.java
```diff
@@ -141,6 +148,21 @@ private void setEnv(GlassFishProperties gfProps) throws Exception {
...
+        // KeyStore (matching its own behaviour), so the handshake will still fail later, but the
+        // log now names the property and the offending path instead of an opaque TLS error.
+        validateConfiguredStorePath(KEYSTORE_FILE);
+        validateConfiguredStorePath(TRUSTSTORE_FILE);
+        File embeddedKeyStore = new File(instanceRoot, "config/keystore.p12");
+        String ownedKeyStore = setStoreSystemPropertyIfUnset(KEYSTORE_FILE, embeddedKeyStore);
+        File embeddedTrustStore = new File(instanceRoot, "config/cacerts.p12");
+        String ownedTrustStore = setStoreSystemPropertyIfUnset(TRUSTSTORE_FILE, embeddedTrustStore);
+        OwnedSslProperties ownedSslProps = new OwnedSslProperties(ownedKeyStore, ownedTrustStore);
+
```
OndroMih, line 165: Move this into a searate method and shorten the comment to the bare minimum.

All the code and comments are very confusing, while it basically just the keystore and trustore properties if not set and the files exist.

nucleus/core/bootstrap-osgi/src/main/java/org/glassfish/main/boot/embedded/EmbeddedGlassFishRuntime.java
```diff
@@ -141,6 +148,21 @@ private void setEnv(GlassFishProperties gfProps) throws Exception {
...
+        // The embedded domain.xml declares javax.net.ssl.keyStore/trustStore as jvm-options, but
+        // those only take effect when the JVM is forked (standalone GlassFish). For in-process
+        // embedded use we set them programmatically here, deferring to the caller's choice if
+        // they've already configured a keystore. If the caller set a path but the file does not
+        // exist we log a WARNING up front: SecuritySupport will still substitute an empty
+        // KeyStore (matching its own behaviour), so the handshake will still fail later, but the
+        // log now names the property and the offending path instead of an opaque TLS error.
+        validateConfiguredStorePath(KEYSTORE_FILE);
+        validateConfiguredStorePath(TRUSTSTORE_FILE);
+        File embeddedKeyStore = new File(instanceRoot, "config/keystore.p12");
```
OndroMih, line 160: I believe that "config/keystore.p12" are already somewhere defined as constants. They shouldn't be hardcoded here.

nucleus/core/bootstrap-osgi/src/main/java/org/glassfish/main/boot/embedded/EmbeddedGlassFishRuntime.java
```diff
@@ -141,6 +148,21 @@ private void setEnv(GlassFishProperties gfProps) throws Exception {
...
+        String ownedKeyStore = setStoreSystemPropertyIfUnset(KEYSTORE_FILE, embeddedKeyStore);
+        File embeddedTrustStore = new File(instanceRoot, "config/cacerts.p12");
```
OndroMih, line 162: Same here, cacerts shouldn't be hardcoded. The path is probably already defined somewhere else in the existing code.

OndroMih (commented):
I think that the code and comments can definitely be improved to make the code more readable and concise.

OndroMih (changes requested):

nucleus/core/bootstrap-osgi/src/main/java/org/glassfish/main/boot/embedded/EmbeddedGlassFishRuntime.java
dmatej: See `GlassFishVariable.TRUSTSTORE_FILE`, but I agree with Ondro, I doubt this is the right way. Another problem will be in my next comment.

nucleus/core/bootstrap-osgi/src/main/java/org/glassfish/main/boot/embedded/EmbeddedGlassFishRuntime.java
```diff
@@ -213,6 +280,44 @@ private String createTempInstanceRoot(GlassFishProperties gfProps) throws Except
...
+        }
+        File keyStore = new File(instanceConfigDir, "keystore.p12");
+        File trustStore = new File(instanceConfigDir, "cacerts.p12");
+        if (keyStore.isFile() && trustStore.isFile()) {
+            return;
+        }
+        try {
+            char[] password = "changeit".toCharArray();
+            KeyTool keyTool = new KeyTool(keyStore, "PKCS12", password);
+            keyTool.generateKeyPair("s1as", "CN=localhost,OU=GlassFish,O=Eclipse Foundation", "RSA", 3650);
```
dmatej, line 307: Using localhost in CN is wrong, the certificate is worthless then. I am thinking about proper solution right now, because I collided with this fact in the docker image. It is better to use the host name, BUT the host name is not known when we create domain1; it is known when we start for the first time, and it may change in the future. Machine can have also alternative host names, certificates have extensions for that, but that is rather for production use ...

The embedded should be ready for production, so it should avoid any automatically generated certificates, it should be provided by the user if he wants to use any encryption, or should use JDK defaults (but as GF always uses the s1as alias, this is not the way to go).

The server ... the same. Domain1 however is precreated "for playing and testing" purposes, so it has a bigger freedom. But right now I am rewriting the docker image, because its encryption is compromised by default.

What we could do is print warning that user should use keytool and create something, link the documentation of keytool or/and our documentation with explanation and howto even for more complex situations.

Btw, SystemProperties.setProperty should be used to work with system properties as it can log changes and especially warnings if some property is overridden, which often causes undesired surprises.

dmatej: However I still have doubts ... the generated backup with localhost and a lot of warnings might be enough for playing and testing too. Or use NetUtils.getHost()? getCanonicalHost()?

nucleus/core/bootstrap-osgi/src/main/java/org/glassfish/main/boot/embedded/EmbeddedGlassFishRuntime.java
dmatej: It is not just about HTTPS, JMX uses it too, for example.

dmatej:
Perhaps we could write few tests with TestContainers. I can adopt the PR, it would help me too with several use cases I have on my mind ...

1. There should be a way to avoid any pregenerated keys and certificates
2. CN localhost is worthless but is the only option available everywhere
3. CN $(hostname) can backfire. It can change, it may not be known, etc. But it works in docker containers, while localhost cannot be used from the outside.
4. WHEN to generate the key and the certificate and WHERE, that is critical. It has a lot of consequences. Your solution is probably the best timing of all possible. To generate some implicit minimum with a lot of warnings. Good enough for local use, not suitable for production domain.

BalusC:
> There should be a way to avoid any pregenerated keys and certificates

Point taken, we'll have to take a step back and fix the failing test itself.

PR #26033, opened 2026-05-15
Several optimizations of Embedded GlassFish startup

OndroMih:
Reduces startup time by about 300ms, from 2.2s to 2.0s, about 10%

Things done:
- set multiple properties in a single command (not so significant but might help)
- preload some work in background ASAP
- load some non-essential things in background (unpacking RARs to the domain)
- do not initialize alias store if it's not needed

### DIFF
--- a/nucleus/core/bootstrap-osgi/src/main/java/org/glassfish/main/boot/embedded/EmbeddedGlassFishRuntime.java
+++ b/nucleus/core/bootstrap-osgi/src/main/java/org/glassfish/main/boot/embedded/EmbeddedGlassFishRuntime.java
@@ -35,15 +35,20 @@ import java.nio.channels.ReadableByteChannel;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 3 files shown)


nucleus/core/bootstrap-osgi/src/main/java/org/glassfish/main/boot/embedded/EmbeddedGlassFishRuntime.java
```diff
@@ -130,7 +172,15 @@ public synchronized void shutdown() throws GlassFishException {
         }
     }
 
-    private void setEnv(GlassFishProperties gfProps) throws Exception {
+    /*
+       Unpacks the domain directory and sets up the properties related to the environment
+
+       Asynchronously completes tasks that aren't essential during startup to move them from the main thread.
```
pzygielo, line 178: At first glance I fail to see where is this async hidden here. Isn't fully _completed_ `Future` returned from here? And everything happens in the same `Thread` that calls `setEnv`? I think `thenRun`, unlike `thenRunAsync` also runs in the same `Thread` that given `Future` is being computed in.

🤔

dmatej (approved):
Would be worth to do some cleanup, but lets make in some future reiteration.

PR #26037, opened 2026-05-22
Fixed hostname autodetection - resolvable

### LINKED ISSUE #26024: JMX Startup failed in docker
The endpoint can have several hostnames and several certificates, but this tries just with host's name.
The `ConnectorStarter.hostname()` should probably try to check all local/loopback endpoints and accept first which passes rules.
Seen in a docker image test for 8.0.2.

```
GF: [#|2026-05-07T22:04:46.130978Z|WARNING|GF 8.0.2|jakarta.enterprise.system.jmx|_ThreadID=142;_ThreadName=Thread-20;_LevelValue=900;|
GF:   Cannot start JMX connector GlassFishConfigBean.com.sun.enterprise.config.serverbeans.JmxConnector(address=0.0.0.0,port=8686,name=system,auth-realm-name=admin-realm)
GF: java.io.IOException: Cannot bind to URL [rmi://42af97ed75c7:8686/jmxrmi]: javax.naming.CommunicationException [Root exception is java.rmi.ConnectIOException: error during JRMP connection establishment; nested exception is: 
GF:     javax.net.ssl.SSLHandshakeException: (certificate_unknown) No name matching 42af97ed75c7 found]
GF:     at java.management.rmi/javax.management.remote.rmi.RMIConnectorServer.newIOException(RMIConnectorServer.java:831)
GF:     at java.management.rmi/javax.management.remote.rmi.RMIConnectorServer.start(RMIConnectorServer.java:478)
GF:     at org.glassfish.admin.mbeanserver.RMIConnectorStarter.start(RMIConnectorStarter.java:305)
GF:     at org.glassfish.admin.mbeanserver.JMXStartupService$JMXConnectorsStarterThread.startConnector(JMXStartupService.java:284)
GF:     at org.glassfish.admin.mbeanserver.JMXStartupService$JMXConnectorsStarterThread.run(JMXStartupService.java:317)
GF: Caused by: javax.naming.CommunicationException [Root exception is java.rmi.ConnectIOException: error during JRMP connection establishment; nested exception is: 
GF:     javax.net.ssl.SSLHandshakeException: (certificate_unknown) No name matching 42af97ed75c7 found]
GF:     at jdk.naming.rmi/com.sun.jndi.rmi.registry.RegistryContext.rebind(RegistryContext.java:182)
GF:     at java.naming/com.sun.jndi.toolkit.url.GenericURLContext.rebind(GenericURLContext.java:266)
GF:     at java.naming/javax.naming.InitialContext.rebind(InitialContext.java:436)
GF:     at java.naming/javax.naming.InitialContext.rebind(InitialContext.java:436)
GF:     at java.management.rmi/javax.management.remote.rmi.RMIConnectorServer.bind(RMIConnectorServer.java:688)
GF:     at java.management.rmi/javax.management.remote.rmi.RMIConnectorServer.start(RMIConnectorServer.java:473)
GF:     ... 3 more
GF: Caused by: java.rmi.ConnectIOException: error during JRMP connection establishment; nested exception is: 
GF:     javax.net.ssl.SSLHandshakeException: (certificate_unknown) No name matching 42af97ed75c7 found
GF:     at java.rmi/sun.rmi.transport.tcp.TCPChannel.createConnection(TCPChannel.java:308)
GF:     at java.rmi/sun.rmi.transport.tcp.TCPChannel.newConnection(TCPChannel.java:204)
GF:     at java.rmi/sun.rmi.server.UnicastRef.newCall(UnicastRef.java:345)
GF:     at java.rmi/sun.rmi.registry.RegistryImpl_Stub.rebind(RegistryImpl_Stub.java:150)
GF:     at jdk.naming.rmi/com.sun.jndi.rmi.registry.RegistryContext.rebind(RegistryContext.java:180)
GF:     ... 8 more
GF: Caused by: javax.net.ssl.SSLHandshakeException: (certificate_unknown) No name matching 42af97ed75c7 found
GF:     at java.base/sun.security.ssl.Alert.createSSLException(Alert.java:130)
GF:     at java.base/sun.security.ssl.TransportContext.fatal(TransportContext.java:383)
GF:     at java.base/sun.security.ssl.TransportContext.fatal(TransportContext.java:326)
GF:     at java.base/sun.security.ssl.TransportContext.fatal(TransportContext.java:321)
GF:     at java.base/sun.security.ssl.CertificateMessage$T13CertificateConsumer.checkServerCerts(CertificateMessage.java:1287)
GF:     at java.base/sun.security.ssl.CertificateMessage$T13CertificateConsumer.onConsumeCertificate(CertificateMessage.java:1164)
GF:     at java.base/sun.security.ssl.CertificateMessage$T13CertificateConsumer.consume(CertificateMessage.java:1107)
GF:     at java.base/sun.security.ssl.SSLHandshake.consume(SSLHandshake.java:393)
GF:     at java.base/sun.security.ssl.HandshakeContext.dispatch(HandshakeContext.java:477)
GF:     at java.base/sun.security.ssl.HandshakeContext.dispatch(HandshakeContext.java:448)
GF:     at java.base/sun.security.ssl.TransportContext.dispatch(TransportContext.java:206)
GF:     at java.base/sun.security.ssl.SSLTransport.decode(SSLTransport.java:172)
GF:     at java.base/sun.security.ssl.SSLSocketImpl.decode(SSLSocketImpl.java:1506)
GF:     at java.base/sun.security.ssl.SSLSocketImpl.readHandshakeRecord(SSLSocketImpl.java:1421)
GF:     at java.base/sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:455)
GF:     at java.base/sun.security.ssl.SSLSocketImpl.ensureNegotiated(SSLSocketImpl.java:922)
GF:     at java.base/sun.security.ssl.SSLSocketImpl$AppOutputStream.write(SSLSocketImpl.java:1291)
GF:     at java.base/java.io.BufferedOutputStream.flushBuffer(BufferedOutputStream.java:125)
GF:     at java.base/java.io.BufferedOutputStream.implFlush(BufferedOutputStream.java:252)
GF:     at java.base/java.io.BufferedOutputStream.flush(BufferedOutputStream.java:240)
GF:     at java.base/java.io.DataOutputStream.flush(DataOutputStream.java:131)
GF:     at java.rmi/sun.rmi.transport.tcp.TCPChannel.createConnection(TCPChannel.java:243)
GF:     ... 12 more
GF: Caused by: java.security.cert.CertificateException: No name matching 42af97ed75c7 found
GF:     at java.base/sun.security.util.HostnameChecker.matchDNS(HostnameChecker.java:229)
GF:     at java.base/sun.security.util.HostnameChecker.match(HostnameChecker.java:103)
GF:     at java.base/sun.security.ssl.X509TrustManagerImpl.checkIdentity(X509TrustManagerImpl.java:466)
GF:     at java.base/sun.security.ssl.X509TrustManagerImpl.checkIdentity(X509TrustManagerImpl.java:432)
GF:     at java.base/sun.security.ssl.X509TrustManagerImpl.checkTrusted(X509TrustManagerImpl.java:237)
GF:     at java.base/sun.security.ssl.X509TrustManagerImpl.checkServerTrusted(X509TrustManagerImpl.java:132)
GF:     at java.base/sun.security.ssl.CertificateMessage$T13CertificateConsumer.checkServerCerts(CertificateMessage.java:1271)
GF:     ... 29 more
GF: |#]

```

dmatej:
### What Is This?

* Fixes #26024 
* Allows to override the hostname autodetection with `AS_HOSTNAME` env variable
  * Then GF doesn't validate, doesn't ask, just uses it everywhere, expecting that the user properly configured DNS and networking so this hostname is easily and reliably resolvable, and will work in all expected cases.
* Splits detected hostnames to three variables:
  * Host's loopback hostname - the simplest case, `InetAddress.getLoopbackAddress().getHostName()`, which is usually `localhost`.
  * Host's name - aka computer name on Windows, should be same as the hostname command on Linux, etc. It often works just on the same computer while it is unknown to the external world. My current laptop is `dmatej-tux`, for example. `NetUtils` tries to use the `AS_HOSTNAME` first, then if it is not set, tries `InetAddress.getLocalHost().getHostName()`, if it fails (can be unresolvable), tries `COMPUTERNAME` (Windows) or `HOSTNAME` (some Linuxes) env variables, if they are not set, uses the `HOST_NAME_LOOPBACK` constant.
  * Canonical hostname - aka computer name in a network. Snakes here. Every network interface can have different set of names, even multiple names. Autodetection tries to evaluate the best host name, however as it has no information about firewall rules, it maz choose a hostname which would not work.

### Past Troubles

All previous versions were good in some scenarios and wrong in others. In https://github.com/eclipse-ee4j/glassfish/pull/25972 I assumed that computer's host name is usually resolvable by default, localhost is resolvable always to loopback ip address, however all these assumptions may be wrong.

Unfortunately Java doesn't have any tool to just get the name. On Linux based systems there's the `hostname` program - in the past were used `HOST` and `HOSTNAME` environment options set by shell, however modern Linux versions don't set them any more. On Windows there's the `COMPUTERNAME` environment option. (EDIT: it all makes sense now, utfg)

If the host name is resolvable on this machine - it doesn't mean it is resolvable from the outside. That means there should be a way user could explicitly set the host name as any internal autodetection doesn't have all the information and despite Java can resolve all known hostnames and reach all available networks including virtual networks, it still cannot choose the one which is not allowed to use for our purposes by a firewall, so it would not be usable to access GlassFish ports.

### Special Situations

I have found the GH Actions is a good reproducer of several scenarios which are OS dependent, shell dependent, DNS dependent, and even dependent on order of actions. Few problems I have found:

* hostname is often just the name of the computer, short without dots. When you list network interfaces, list their host names, there may be full host name (with dots), for example ...
* Ubuntu based GitHub Actions use 
  * hostname: `runnervm3jyl0`
  * list: `runnervm3jyl0.fwbxogqxw2hu1gnhcu514ce0ub.bx.internal.cloudapp.net, localhost`
* While Windows based GitHub Actions use
  * hostname: `runnervmlu3mh`
  * list: `runnervmlu3mh.jmhn0wqfkjtehmf45kke2dh3lh.bx.internal.cloudapp.net` - no localhost here!

### Why Avoid `getLocalHost().getHostName()`?

Claude's question and answer, I asked how to get the information in Java.

* This does DNS resolution — slow, may fail, may return wrong name:
```
InetAddress.getLocalHost().getHostName();
```

* This also resolves
```
InetAddress.getLocalHost().getCanonicalHostName();
```
It can cause issues when:

* Running in containers (Docker/Kubernetes)
* No DNS is configured
* Hostname isn't registered in DNS
* You need the raw OS hostname, not the DNS name

--- end of Claude's citation.

I have found yet another problem in docker - network services sometimes start bit later, so the DNS resolution may change.

And another problem in GitHub Actions - on MacOS and Windows nodes the DNS is terribly slow, resolving all detected IPs takes around 12 seconds.

### We Don't Use the Host Name just for Networking

* We use it also as an identifier. That may cause problems, because the host name may change. We should probably warn user to avoid host name changes OR to recreate whole domain and its nodes.

### Random Notes Made While Resolving this PR

This chapter is not much important, but I guess when I come back here later, I will read it once again trying to understand what the hell is going on.

Loaded candidate host names: 
```
Mac:     [sjc20-bb661-e9920188-c77e-4f67-81f3-e3b2eed880c9-824BDF8CA402.local, localhost]
Windows: [runnervmuktm0.2b2fowm5wf5e1hblkegia5xywh.yx.internal.cloudapp.net, runnervmuktm0]
Ubuntu:  [runnervm5mmn9.ejzbqw2iu35etlto4ngnz0uiub.xx.internal.cloudapp.net, localhost]
```
Then:

Mac: Chooses `sjc20-bb661-e9920188-c77e-4f67-81f3-e3b2eed880c9-824BDF8CA402.local`, but then GF doesn't allow connections OR firewall forbids plain HTTP OR something times out, the `start-domain` blocks forever.

Windows: Chooses `runnervmuktm0`, while `getCanonicalName` returns `localhost` (loopback). Note: that was true when I was filtering the stream also with a check if I can connect to a created socket. If connection attempt failed, both host names were dropped, so just localhost remained. However that was time consuming and did not work on Mac.

Ubuntu: Chooses `runnervm5mmn9`, passes all tests.

Time is another issue, hostname resolution on GHA is extremely slow:

[INFO]    +-- [OK] getResolvableHostNames - 11.56 s
Mac: 
* `getCanonicalHostName` - 65.16 s
* Every `isResolved` takes about 5 seconds.

When the server is not running, requests to both Mac and Windows time out (asadmin has 20 s timeout in the `GlassFishTestEnvironment`).


nucleus/common/common-util/src/main/java/com/sun/enterprise/util/net/NetUtils.java
```diff
@@ -27,28 +28,84 @@
...
 
     private static final Logger LOG = System.getLogger(NetUtils.class.getName());
+
     /** Maximal allowed port number */
     public static final int MAX_PORT = 65535;
-    private static final String LOCALHOST_IP = "127.0.0.1";
 
-    private static final int IS_RUNNING_DEFAULT_TIMEOUT = 3000;
-    private static final int IS_PORT_FREE_TIMEOUT = 1000;
+    private static final String AS_HOSTNAME = System.getenv("AS_HOSTNAME");
```
OndroMih, line 59: Would it be possible to get this from GlassFish configuration via TranslatedConfigView.getProperty()?

With that, the property could be named as.hostname, and would be resolved both against a system property called "as.hostname" and against env var "AS_HOSTNAME", see: https://docs.omnifish.ee/glassfish/latest/administration-guide/overview.html#variable-references-in-configuration

OndroMih: Same comment for the timeouts below.

nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/ChangeAdminPasswordCommand.java
```diff
@@ -149,16 +150,18 @@ protected int executeCommand() throws CommandException {
         try {
             RemoteRestAdminCommand rac = new RemoteRestAdminCommand(name, programOpts.getHost(), programOpts.getPort(),
                     programOpts.isSecure(), programOpts.getUser(), programOpts.getPassword(), logger, false, false);
+            // If the server is not listening, some environments can be configured to behave as a black hole.
+            // Example: GitHub Actions and Windows+MacOS nodes, firewalls can do that (DROP, while REJECT would be ok).
+            rac.setConnectTimeout(5_000);
+            rac.setReadTimeout(30_000);
```
OndroMih, line 156: Shouldn't these timeouts be configurable, e.g. via an argument or env var?

dmatej: Just one of them was configurable until now and was so long (10 minutes) that nobody (tests) waited for it. I don't think we need to make it configurable at this moment, in fact the whole action should be a matter of 0-3 seconds. When it times out, something went wrong.

I also don't think the command impl is correct, because again it tries to be smarter than the user. I believe these commands should not automatically and silently resolve if they are in remote or local regime, but at this moment I am not sure if results are or are not same then.

Both timeouts are long enough. If not, I want to see the use case and reason. Then we can make it more complicated. However perhaps for 8.0.5 we should make both connection and read timeouts a parameter or even an argument. It would play well with a change of the `--host` argument so it would mean automatically a remote command (with default loopback host).

nucleus/common/common-util/src/main/java/com/sun/enterprise/util/net/NetUtils.java
dmatej: No, see GlassFishVariable. Chicken-egg.
```
    /** Autodetected host name. It is not guaranteed that it is resolvable. */
    HOST_NAME("AS_HOSTNAME", "com.sun.aas.hostName"),
```

OndroMih (approved):

.github/workflows/build-windows.yml
```diff
@@ -19,6 +19,20 @@ jobs:
...
+    - name: Print OS Info
+      shell: pwsh
+      run: |
+        ls env: | sort Name
+        cat C:\Windows\System32\drivers\etc\hosts
+        try {
+            Get-NetIPConfiguration -Detailed
+        } catch {
+            Write-Warning "Get-NetIPConfiguration -Detailed failed (known NetTCPIP bug): $_"
+            Get-NetIPConfiguration
```
pzygielo, line 31: 1. what does it mean _known NetTCPIP bug_? Any ref?
2. https://github.com/eclipse-ee4j/glassfish/issues/26197

PR #26041, opened 2026-05-27
Update embedded Maven plugin applications with GlassFish configuration

### LINKED ISSUE #26036: Test improvement: Convert Embedded GlassFish Maven Plugin Tests to use the latest plugin version
All the tests in https://github.com/eclipse-ee4j/glassfish/tree/6cab4ce3aad9bc65b52bc761a4aa88634a2bd0aa/appserver/tests/embedded/maven-plugin directory use the old maven plugin version 5.

Convert them to use the latest 8.0 version of the plugin: https://github.com/eclipse-ee4j/glassfish-maven-embedded-plugin

1. The new version uses a different artifactId: `embedded-glassfish-maven-plugin` instead of the old `maven-embedded-glassfish-plugin`(the groupId is the same)

2. The new version runs Embedded GlassFish in a separate JVM by default. It's OK to keep the default for most of the tests but it would be good to disable this mode for at least one test so that we test it too. For example, the `simple` test case, or the `applicationDispatcher` test case. To disable it, set in the plugin config:

```
<configuration>
    <fork>false</fork>
</configuration>
```

3. It's possible to remove the dependencies from the plugin definition in the [maven-plugin/pom.xml](https://github.com/eclipse-ee4j/glassfish/tree/6cab4ce3aad9bc65b52bc761a4aa88634a2bd0aa/appserver/tests/embedded/maven-plugin/pom.xml) file and just define the configuration property `<version>${project.version}</version>`. The plugin will know how to use the `glassfish-embedded-all` artifact with that version

SavioFerreira:
Fixes #26036.

## Changes
- Updated all embedded maven plugin test modules with the required configuration changes.

## Validation
- Full embedded maven plugin test suite passes successfully.

## Additional Investigation
I also tested non-forked execution (`mvn embedded-glassfish:run`).

While testing this mode, I encountered initialization/runtime issues related to embedded startup and command resolution.

### Successful test suite execution
Screenshot From 2026-05-26 22-29-01

### Non-forked execution issue
Screenshot From 2026-05-26 22-25-29

### DIFF
--- a/appserver/tests/embedded/maven-plugin/pom.xml
+++ b/appserver/tests/embedded/maven-plugin/pom.xml
@@ -156,7 +156,7 @@
                     <artifactId>embedded-glassfish-maven-plugin</artifactId>
                     <version>8.0</version>
                     <configuration>
-                        <glassfish.version>8.0.1</glassfish.version>
+                        <glassfish.version>${project.version}</glassfish.version>
                         <app>target/${project.artifactId}.war</app>
                         <port>8080</port>
                         <name>${project.artifactId}</name>
--- a/appserver/tests/embedded/maven-plugin/pom.xml
+++ b/appserver/tests/embedded/maven-plugin/pom.xml
@@ -153,9 +153,10 @@
                 </plugin>
                 <plugin>
                     <groupId>org.glassfish.embedded</groupId>
-                    <artifactId>maven-embedded-glassfish-plugin</artifactId>
-                    <version>5.1</version>
+                    <artifactId>embedded-glassfish-maven-plugin</artifactId>
+                    <version>8.0</version>
                     <configuration>
+                        <glassfish.version>8.0.1</glassfish.version>
                         <app>target/${project.artifactId}.war</app>
                         <port>8080</port>
                         <name>${project.artifactId}</name>
@@ -168,18 +169,6 @@
                             <property>com.sun.enterprise.naming.allowJndiLookupFromOSGi=false</property>
                         </systemProperties>
                     </configuration>
-                    <dependencies>
-                        <dependency>
-                            <groupId>org.glassfish.main.common</groupId>
-                            <artifactId>simple-glassfish-api</artifactId>
-                            <version>${project.version}</version>
-                        </dependency>
-                        <dependency>
-                            <groupId>org.glassfish.main.extras</groupId>
-                            <artifactId>glassfish-embedded-all</artifactId>
-                            <version>${project.version}</version>
-                        </dependency>
-                    </dependencies>
                 </plugin>
             </plugins>
         </pluginManagement>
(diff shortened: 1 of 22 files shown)


appserver/tests/embedded/maven-plugin/pom.xml
```diff
@@ -153,9 +153,10 @@
                 </plugin>
                 <plugin>
                     <groupId>org.glassfish.embedded</groupId>
-                    <artifactId>maven-embedded-glassfish-plugin</artifactId>
-                    <version>5.1</version>
+                    <artifactId>embedded-glassfish-maven-plugin</artifactId>
+                    <version>8.0</version>
                     <configuration>
+                        <glassfish.version>8.0.1</glassfish.version>
```
pzygielo: I think the plan is to test current (i.e. `${project.version}`) version...

appserver/tests/embedded/maven-plugin/pom.xml
```diff
@@ -168,18 +169,6 @@
                             <property>com.sun.enterprise.naming.allowJndiLookupFromOSGi=false</property>
                         </systemProperties>
                     </configuration>
-                    <dependencies>
```
dmatej, line 171: ... and the original pom was overriding plugin's dependencies on older GlassFish than current snapshot.

dmatej: @OndroMih Are those dependencies the same as in previous version of the plugin, except versions?

appserver/tests/embedded/maven-plugin/pom.xml
SavioFerreira: Using ${project.version} causes Maven to resolve embedded-glassfish-maven-plugin:8.0.3-SNAPSHOT, but this artifact is not currently available.
image

OndroMih: Version of the maven plugin should be 8.0, that's correct. But the value of the <glassfish.version> property should be `${project.version}`. 

Of course, it requires the Embedded GlassFish JAR artifact to be built before running the test suite. I think that before this change, maven detected the dependency and built it. Now the dependency is not explicit so Maven doesn't know that it should be build.

Therefore maybe it would be better to define plugin dependency on the `glassfish-embedded-all` artifact as before, instead of plain <version> in config, so that maven tries to build it before running these tests.

appserver/tests/embedded/maven-plugin/pom.xml
OndroMih: Yes, the coordinates are the same, they come from the GlassFish repository and they haven't changed in years.

I'm not sure why `simple-glassfish-api` was added as a dependency and whether it was needed. I doubt it. It seems OK to me to remove it from dependencies.

The `glassfish-embedded-all` dependency is not needed anymore with the latest version of the maven plugin - it's enough to define `version` config element. But, as discussed in another thread, we should probably keep it so that maven knows that it's a dependency and it should build it.

OndroMih (changes requested):

SavioFerreira:
Looks like the undeploy goal is having some issues with the expected response.

When running the test suite, the `jmxUndeployEvent` test fails because it expects a `404` response after undeployment, but the application still returns `200`.

I've attached the test output and failure details for reference.
Screenshot From 2026-06-04 18-30-36
Screenshot From 2026-06-04 18-28-02

PR #26043, opened 2026-05-27
Fix: Add graceful validation handling for invalid MDB ActivationConfigProperty types

### LINKED ISSUE #26040: [Bug] EJB Container crashes during deployment when using @ActivationConfigProperty for resourceAdapter

7.0.23 (Docker Image: ghcr.io/eclipse-ee4j/glassfish:7.0.23)

### JDK version

JDK 17

### OS

Docker Desktop (Windows host)

### Database

MySQL 8.0.33

### Problem Description

**Current Behavior:**
When attempting to deploy a Jakarta EE application containing a Message-Driven Bean (MDB) to the `ghcr.io/eclipse-ee4j/glassfish:7.0.23` Docker image, the EJB container initialization fails entirely if the `@ActivationConfigProperty` for `resourceAdapter` is specified in the Java code. 

The deployment is aborted, throwing an `IllegalArgumentException` stating that the type for the environment property is illegal.

**Expected Behavior:**
The MDB should successfully parse the string `"jmsra"`, bind to the internal JMS resource adapter, and allow the application to deploy, as it did in older `javax` versions of GlassFish.

**Stack Trace Snippet:**
```text
[#|2026-05-24T18:58:36.844067Z|WARNING|GF 7.0.23|jakarta.enterprise.resource.resourceadapter.com.sun.enterprise.connectors.inbound|_ThreadID=97;_ThreadName=AutoDeployer;_LevelValue=900;_MessageID=endpoint.activation.failure;|
  RAR8501: Exception during endpoint activation for ra [ jmsra ], activationSpecClass [ com.sun.messaging.jms.ra.ActivationSpec ] : java.lang.IllegalArgumentException: Illegal type for environment properties: [interface jakarta.resource.spi.ResourceAdapter]|#]

Caused by: java.lang.IllegalArgumentException: Illegal type for environment properties: [interface jakarta.resource.spi.ResourceAdapter]
        at com.sun.enterprise.deployment.EnvironmentProperty.getObjectFromString(EnvironmentProperty.java:376)
        at com.sun.enterprise.deployment.EnvironmentProperty.getResolvedValueObject(EnvironmentProperty.java:127)
        at com.sun.enterprise.connectors.util.SetMethodAction.run(SetMethodAction.java:77)
```

### Steps to reproduce

1. Create an MDB annotated as follows:

```java
@MessageDriven(
        activationConfig = {
                @ActivationConfigProperty(propertyName = "destinationLookup", propertyValue = "jms/NewUserTopic"),
                @ActivationConfigProperty(propertyName = "destinationType", propertyValue = "jakarta.jms.Topic"),
                @ActivationConfigProperty(propertyName = "resourceAdapter", propertyValue = "jmsra") // <-- Triggers the crash
        }
)
public class MyListener implements MessageListener { 
    @Override
    public void onMessage(Message message) {}
}

```

2. Build the application into a `.war` file.
3. Deploy the application via autodeploy in a fresh GlassFish 7.0.23 Docker container.
4. Observe the `server.log` crash.

**Workaround:**
Removing the `resourceAdapter` property from the Java annotation and instead defining it globally in `glassfish-ejb-jar.xml` allows the application to deploy successfully.

### Impact of Issue

This issue severely impacts the developer experience when migrating legacy applications to Jakarta EE 10 and GlassFish 7. It prevents developers from utilizing standard annotation-based MDB configuration for custom resource adapters, forcing a fallback to XML deployment descriptors. It completely breaks containerized CI/CD pipelines that rely on autodeploy.

dilansachcha:
This PR addresses issue #26040 by adding graceful error handling inside `ConnectorMessageBeanClient.setup()`. 

Previously, if an invalid property (such as `resourceAdapter`) was passed via `@ActivationConfigProperty`, the `EnvironmentProperty.getObjectFromString()` method would throw an `IllegalArgumentException` during reflection. Because this was only caught by a generic `Exception` block, it propagated the root cause up, printing a raw reflection stack trace and a generic "Setup failed" error.

This patch intercepts the `IllegalArgumentException` specifically, outputting a clean `SEVERE` validation message indicating which MDB failed and why the property type was invalid. By not passing the root exception up to the `ConnectorRuntimeException`, it prevents the raw reflection stack trace from bleeding into `server.log`, drastically improving the Developer Experience (DX) for annotation misconfigurations.

Fixes #26040

### DIFF
--- a/appserver/connectors/connectors-inbound-runtime/src/main/java/com/sun/enterprise/connectors/inbound/ConnectorMessageBeanClient.java
+++ b/appserver/connectors/connectors-inbound-runtime/src/main/java/com/sun/enterprise/connectors/inbound/ConnectorMessageBeanClient.java
@@ -137,7 +137,6 @@ public final class ConnectorMessageBeanClient implements MessageBeanClient, Mess
         ActiveInboundResourceAdapter aira = getActiveResourceAdapter(resourceAdapterMid);
         aira.updateMDBRuntimeInfo(descriptor, this.messageBeanPM.getPoolDescriptor());
 
-        // the resource adapter this MDB client is deployed to
         ResourceAdapter ra = aira.getResourceAdapter();
         if (ra == null) {
             String i18nMsg = localStrings.getString("msg-bean-client.ra.class.not.specified", resourceAdapterMid);
@@ -155,11 +154,16 @@ public final class ConnectorMessageBeanClient implements MessageBeanClient, Mess
             throw new IllegalStateException("Unsupported message listener type");
         }
 
-        if (logger.isLoggable(FINEST)) {
-            logger.log(FINEST, "ActivationSpecClassName = " + activationSpecClassName);
-        }
         try {
-            ActivationSpec activationSpec = getActivationSpec(aira, activationSpecClassName);
+            ActivationSpec activationSpec;
+            try {
+                activationSpec = getActivationSpec(aira, activationSpecClassName);
+            } catch (IllegalArgumentException ex) {
+                String errorMsg = localStrings.getString("msg-bean-client.activation.invalid.property", descriptor.getName());
+                logger.log(Level.SEVERE, errorMsg, ex);
+                throw new ConnectorRuntimeException(errorMsg, ex);
+            }
+
             activationSpec.setResourceAdapter(ra);
 
             // at this stage, activation-spec is created, config properties merged with ejb-descriptor.
@@ -171,9 +175,12 @@ public final class ConnectorMessageBeanClient implements MessageBeanClient, Mess
             this.myState = BLOCKED;
             ra.endpointActivation(this, activationSpec);
             aira.addEndpointFactoryInfo(beanID_, new MessageEndpointFactoryInfo(this, activationSpec));
+
+        } catch (ConnectorRuntimeException cre) {
+            throw cre;
         } catch (Exception ex) {
             logger.log(Level.WARNING, "endpoint.activation.failure",
-                new Object[] {resourceAdapterMid, activationSpecClassName, ex});
+            new Object[] {resourceAdapterMid, activationSpecClassName, ex});
             throw new ConnectorRuntimeException("Setup failed.", ex);
         }
     }
--- a/appserver/connectors/connectors-inbound-runtime/src/main/resources/com/sun/enterprise/connectors/inbound/LocalStrings.properties
+++ b/appserver/connectors/connectors-inbound-runtime/src/main/resources/com/sun/enterprise/connectors/inbound/LocalStrings.properties
@@ -19,3 +19,4 @@ msg-bean-client.multiple-ras-supporting-message-listener=The 'resource-adapter-m
 msg-bean-client.defaulting.message-listener.supporting.rar=No 'resource-adapter-mid' specified, defaulting to the resource-adapter [ {0} ] that supports the requested 'message-listener' [ {1} ]
 msg-bean-client.ra.class.not.specified=<resourceadapter-class> must be specified for the resource-adapter [ {0} ] having inbound (<inbound-resourceadapter> or its equivalent annotation) artifacts
 msg-bean-client.could-not-derive-ra-mid=Unable to determine resourceAdapterMid for descriptor [ {0} ]
+msg-bean-client.activation.invalid.property=Deployment failed for MDB [ {0} ]: Invalid type provided for ActivationConfigProperty.


appserver/connectors/connectors-inbound-runtime/src/main/java/com/sun/enterprise/connectors/inbound/ConnectorMessageBeanClient.java
```diff
@@ -171,6 +171,10 @@ public void setup(MessageBeanProtocolManager messageBeanPM)
             this.myState = BLOCKED;
             ra.endpointActivation(this, activationSpec);
             aira.addEndpointFactoryInfo(beanID_, new MessageEndpointFactoryInfo(this, activationSpec));
+        } catch (IllegalArgumentException ex) {
+            String errorMsg = "Deployment failed for MDB [" + descriptor.getName() + "]: Invalid type provided for ActivationConfigProperty. " + ex.getMessage();
```
pzygielo: This should perhaps be externalized like `endpoint.activation.failure` below.

dilansachcha: Makes perfect sense, just externalized it and pushed the update.

appserver/connectors/connectors-inbound-runtime/src/main/java/com/sun/enterprise/connectors/inbound/ConnectorMessageBeanClient.java
```diff
@@ -171,6 +171,10 @@ public void setup(MessageBeanProtocolManager messageBeanPM)
             this.myState = BLOCKED;
             ra.endpointActivation(this, activationSpec);
             aira.addEndpointFactoryInfo(beanID_, new MessageEndpointFactoryInfo(this, activationSpec));
+        } catch (IllegalArgumentException ex) {
+            String errorMsg = localStrings.getString("msg-bean-client.activation.invalid.property", descriptor.getName(), ex.getMessage());
```
pzygielo: Could we - instead of including `ex.getMessage` in `errorMsg` - pass current `ex` to newly created `CRtEx`, like it's done in block below?

pzygielo: I think ec820bc098bafc8c2febbdaf1da280b42035c130 is missing changes in `ConnectorMessageBeanClient`.

dilansachcha: Good point. I originally left it out to keep the server logs clean from the reflection stack trace, but keeping the exception chain makes much more sense for debugging. Removed the `{1}` placeholder and passed the `ex` through. Just pushed!

dilansachcha: > I think [ec820bc](https://github.com/eclipse-ee4j/glassfish/commit/ec820bc098bafc8c2febbdaf1da280b42035c130) is missing changes in `ConnectorMessageBeanClient`.

yep, this was a silly mistake during the push

dilansachcha:
Hey @pzygielo, the Jenkins PR merge passed successfully, but it looks like the Ubuntu GitHub Action tripped over a flaky test in `ClusterITest.java` (random port collision on `JMX_SYSTEM_CONNECTOR_PORT`). Since I don't have permission to re-run the action, could you re-trigger it when you have a moment?

OndroMih:
@dilansachcha , ubuntu job restarted

pzygielo:
I'm not 100% what **I** think about that.

1. The existing `catch (Exception)` - with its message `endpoint.activation.failure` suggests a bit that the call at 172 (`ra.endpointActivation(this, activationSpec);`) was unsuccessful. Which we now know - is not the case for the reproducer.
1. In the case of original issue - it's `IAEx` being thrown from 162 (` getActivationSpec(aira, activationSpecClassName)`).
So the catch block responds as this is the actual case. But there are few other calls, that I have no guarantee to not throw `IAEx` as well.
:thinking: 

Maybe just the **existing** block should log a `SEVERE`, modified message (updated with MDB name) instead?

dilansachcha:
> 1. The existing `catch (Exception)` - with its message `endpoint.activation.failure` suggests a bit that the call at 172 (`ra.endpointActivation(this, activationSpec);`) was unsuccessful. Which we now know - is not the case for the reproducer.
> 2. In the case of original issue - it's `IAEx` being thrown from 162 (` getActivationSpec(aira, activationSpecClassName)`).
> So  the catch block responds as this is the actual case. But there are few other calls, that I have no guarantee to not throw `IAEx` as well.

You make a great point about the 'blast radius' of that try block. Catching `IllegalArgumentException` over all of those calls is definitely too broad, as `endpointActivation` could throw it for entirely different reasons.

However, if we just fall back to the generic `catch (Exception)` block for everything, we lose the primary goal of issue #26040, which was to intercept the property type mismatch specifically and provide a clean validation message instead of a generic setup failure.

What if we narrow the scope to get the best of both worlds? We can pull `getActivationSpec(...)` out into its own specific `try-catch` block to handle the `IllegalArgumentException` for property mismatches, and leave the rest of the activation logic in the existing generic `try-catch` block. This keeps the specific validation message intact without risking false positives on `endpointActivation`.

I'll push a commit with this narrowed scope so you can see what it looks like!

dilansachcha:
Hey @pzygielo and @OndroMih, it looks like the Ubuntu runner tripped over `admin-tests` again. I checked the logs and it is the `ClusterITest` port collision flakiness (the OS handed out port 45329 twice for JMX and ASADMIN). I actually opened PR #26049 to address this.

Since Jenkins, MacOS, and Windows all passed cleanly with my MDB code, could one of you re-trigger the Ubuntu job when you have a second? Thank you!

pzygielo:
@dilansachcha could you rebase on current main with squashing to single commit, please?

dilansachcha:
@pzygielo Done! I rebased onto the latest `main`, squashed down into a single clean commit, and force-pushed!

OndroMih (approved):

OndroMih:
It seems to me that all of the review comments were addressed and it all looks good to me, I'll merge it.

PR #26044, opened 2026-05-27
Update man page for set-log-attributes #24267 to reflect current JUL-extension logging handlers

### LINKED ISSUE #24267: Update man page for set-log-attributes
In 7.0.0 it still contains old handler names etc.

renatsaf:
Fixes #24267

## Summary

- The `set-log-attributes` man page (`.1`) still documented the long-removed `com.sun.enterprise.server.logging.GFFileHandler.*` / `java.util.logging.*` handlers. This PR replaces the attribute catalog with the current `org.glassfish.main.jul.handler.*` properties (`GlassFishLogHandler`, `SimpleLogHandler`, `SyslogHandler`), with defaults sourced from `nucleus/admin/template/src/main/resources/config/logging.properties` and property definitions from `GlassFishLogHandlerProperty` / `SyslogHandlerProperty`.
- Fixed the NAME/DESCRIPTION wording: this subcommand sets handler/formatter attributes, not logger levels, so "for one or more loggers" becomes "for one or more handlers or formatters" in both the `.1` and the `.adoc`.
- Dropped the stale "Depending on the attributes you set, a server restart may be necessary." sentence in the man page — no longer true for the JUL-extension handlers.
- Added a second example demonstrating the colon-separated multi-attribute syntax that the Synopsis advertises but neither file previously illustrated. Added to both `.1` and `.adoc`.
- The `.adoc` continues to defer the full attribute catalog to the Administration Guide; only the man page carries the inline list (kept that split intentionally to avoid drift between the two documents).

## Test plan

- [ ] Verify the man page renders sensibly (line-length, indentation matches sibling man pages in `nucleus/core/logging/src/main/manpages/.../commands/`).
- [ ] Verify the `.adoc` builds as part of the reference manual without warnings.
- [ ] Spot-check that each documented property in the man page exists in either `GlassFishLogHandlerProperty.java`, `SyslogHandlerProperty.java`, or the template `logging.properties`.

### DIFF
--- a/docs/reference-manual/src/main/asciidoc/set-log-attributes.adoc
+++ b/docs/reference-manual/src/main/asciidoc/set-log-attributes.adoc
@@ -11,7 +11,7 @@ prev=set-batch-runtime-configuration.html
 
 == set-log-attributes
 
-Sets the logging attributes for one or more loggers
+Sets the logging attributes for one or more handlers or formatters
 
 === Synopsis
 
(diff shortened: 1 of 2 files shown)


OndroMih:
Looks good. No code changes so it doesn't impact tests, the failed Jeknins job can be ignored. I checked the PDF in the artifacts and it looks OK.

Merging.

OndroMih (approved):

PR #26045, opened 2026-05-28
Cache WeldUtils.isValidAnnotation results and share excludedScopes across the recursion

BalusC:
## Summary

`WeldUtils.isValidAnnotation` is on the request hot path (called by Weld every time it needs to decide whether some annotation is a CDI scope / interceptor binding / etc.) but its current implementation has two compounding inefficiencies that, combined, dominate request CPU for non-trivial Faces / CDI workloads:

1. **Per-call HashSet allocation.** Every invocation — including every recursive call walking the meta-annotation chain — allocates a fresh `HashSet<String>(excludedTypeNames)` and `addAll`'s the parent's excluded set into it. For an annotation chain of depth N that's N+1 HashSet allocations plus copy work, on every request, for every annotation Weld checks.

2. **No memoization.** The result for a given `(annotationType, validTypeNames)` pair is invariant — `@Dependent` is always a CDI scope, regardless of when you ask. But nothing caches that decision; Weld re-walks `@Dependent`'s meta-annotation chain on every CDI lookup.

This PR fixes both:

- **Shared mutable excluded-scopes set with backtracking.** Splits out a private recursive helper that takes the `HashSet<String> excludedScopes` by reference, adds the current `annotationType.getName()` on entry, and removes it in a `finally` block on exit. A single HashSet allocation per top-level call, instead of N+1 per chain walk.

- **Static result cache** keyed by the caller-supplied `validTypeNames` Collection (Weld reuses the same Set instance across scanning passes) and the annotation class being checked. First call walks the tree; every call after is `O(1)`.

- A `clearIsValidAnnotationCache()` static is exposed for any deployment lifecycle hook that wants to invalidate it, although annotation hierarchies don't change at runtime so the cache is safe to retain for the JVM lifetime.

The patch is binary-compatible — same `protected` signature for `isValidAnnotation(Class, Collection, Collection)`, same return semantics; the new private overload and the cache fields are additions.

## How the hot spot was found

JFR profile of a Mojarra perf bench against stock GlassFish 7.0.25:

- Mojarra perf-bench module: [eclipse-ee4j/mojarra#5756](https://github.com/eclipse-ee4j/mojarra/pull/5756) (WAR + JUnit/Arquillian driver, 18 scenarios spanning GET-only, full postback, ajax-postback, and `f:viewParam`-GET-runs-full-lifecycle).
- Config: `-Dperf=true -Dperf.warmup=20 -Dperf.runs=200`, GlassFish 7.0.25, JDK 21.0.1, JFR `settings=profile`, `stackdepth=128`.

Findings on stock 7.0.25:

```
WeldUtils.isValidAnnotation  ≈ 52.9 %  of http-listener CPU
```

…uniform across all five `http-listener-1(*)` threads (51–54 % each). The CPU samples showed depth-5 stacks (`HashMap` leaf → 4 self-recursive `isValidAnnotation` frames, `truncated=false`) — the JIT was folding the caller chain into a tight inlined loop. The allocation profile confirmed the picture: `isValidAnnotation` was the #5 allocation site, at 3.5 % of total allocations across the JVM.

Source-side check: the method is byte-for-byte identical across the `7.0`, `9.0`, and `main` branches of this repo. The two recent perf commits on `9.0` only (`62596cb` "Faster detection of implicit CDI bean archives" and `7840374` "Faster sniffer validation") fix *different* methods in the same file and are deployment-time concerns, not request-path — so the same hot spot exists on every supported GlassFish version, and upgrading does not help.

## Hard numbers — same Mojarra workload, same JVM args, same bits except this patch

Stock 7.0.25 vs locally-built 7.0.26-SNAPSHOT with this PR applied:

| | stock 7.0.25 | patched 7.0.26-SNAPSHOT | Δ |
|---|---:|---:|---:|
| **Bench wall clock** (200 measurement iterations, 18 scenarios) | **154 s** | **80 s** | **−48 %** |
| Total JFR CPU samples (~20 ms each) | 7,447 | 2,686 | −64 % |
| `WeldUtils.isValidAnnotation` % of http-listener CPU | **52.9 %** | **0.76 %** | **−98.6 %** |
| http-listener samples | 7,258 | 2,492 | |
| `WeldUtils.isValidAnnotation` samples | 3,841 | 19 | |

Roughly half the request CPU on this workload disappears.

The bench is a managed-GlassFish HTTP-driven loop over a representative spread of Faces scenarios (forms with `@FacesConverter(managed=true)` / `@FacesValidator(managed=true)` chains, postback + ajax-partial flows, GETs with `f:viewParam` running the full lifecycle, etc.). It's deliberately CDI-discovery-mode=`annotated` (the CDI 4.0 default) so the managed-converter / managed-validator lookup path is exercised on every request — which is precisely the path that calls into `isValidAnnotation`.

## Mojarra-side hot list (informational — for cross-reference with PR 5756)

With the Weld noise gone the underlying Mojarra hot list becomes visible cleanly:

| % (post-patch) | Method |
|---:|---|
| 3.76 % | `com.sun.faces.el.DemuxCompositeELResolver._getValue` |
| 3.48 % | `jakarta.faces.component.ComponentStateHelper.get` |
| 2.55 % | `jakarta.faces.component.ComponentStateHelper.put` |
| 2.22 % | `UIComponentBase$AttributesMap.getPropertyDescriptor` |
| 2.18 % | `com.sun.faces.util.HtmlUtils.writeText` |
| 1.86 % | `com.sun.faces.util.Cache.get` |
| 1.50 % | `UIComponent.getValueExpression` |
| 1.42 % | `UIData.restoreDescendantState` |
| 1.17 % | `UIComponentBase$AttributesMap.get` |
| 1.05 % | `UIRepeat.restoreChildState` / `UIData.saveDescendantState` |

Same shortlist as the pre-patch projection in PR 5756 — confirming the analysis and giving the Mojarra side a clean baseline to optimize against. None of those are GlassFish concerns; included only to demonstrate that the post-patch profile is dominated by *application-server* and *Faces-implementation* work, not by Weld housekeeping.

## Caveats

- Numbers above are from a single back-to-back run pair. Direction is overwhelmingly clear (−48 % wall clock isn't run-to-run noise) but exact percentages will shift slightly between runs.
- The bench targets request-path CPU. Deployment-time costs (where most of `WeldUtils`' other methods live) are not measured here; this PR doesn't touch those.
- The cache is keyed by the `Collection<String>` instance Weld passes for `validTypeNames`. The grep-and-trace I did suggests Weld reuses the same Set instance across scanning passes, which makes the cache effective. If a caller ever constructs a fresh Collection per call, the cache degrades gracefully — first call walks the tree, second call (same identity, different identity) walks again, no incorrectness.

## Reproducibility

Both JFR files are preserved on the Mojarra perf branch from PR 5756:

```
test/perf/target/perf-bench-7025.jfr           # stock 7.0.25
test/perf/target/perf-bench-7026-patched.jfr   # 7.0.26-SNAPSHOT with this PR
```

To re-run end-to-end:

```bash
# 1. Build this GlassFish branch
cd glassfish && git checkout fix_weldutils_isvalidannotation_performance
. java21 && mvn install -Pfastest -T4C            # ~1.5 min

# 2. Run the Mojarra perf bench against it (per the procedure on mojarra PR 5756)
cd mojarra && git checkout profile_all_prs        # mojarra branch with all in-flight perf PRs merged
mvn -pl impl install -DskipTests
mvn -pl test/perf -am clean package -DskipTests -Dglassfish.version=7.0.26-SNAPSHOT
# add the two <jvm-options> for JFR to test/perf/target/glassfish7/glassfish/domains/domain1/config/domain.xml
mvn -pl test/perf failsafe:integration-test failsafe:verify -Dperf=true -Dperf.warmup=20 -Dperf.runs=200
jfr view --width 200 hot-methods /tmp/perf-bench.jfr
```

### DIFF
--- a/appserver/web/gf-weld-connector/src/main/java/org/glassfish/weld/connector/WeldUtils.java
+++ b/appserver/web/gf-weld-connector/src/main/java/org/glassfish/weld/connector/WeldUtils.java
@@ -41,7 +41,10 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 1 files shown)


dmatej (approved):

appserver/web/gf-weld-connector/src/main/java/org/glassfish/weld/connector/WeldUtils.java
```diff
@@ -142,6 +145,13 @@ public static enum BDAType {
         cdiEnablingAnnotations.add(jakarta.ejb.Singleton.class.getName());
     }
 
+    /**
+     * Memoization of {@link #isValidAnnotation(Class, Collection, Collection)} keyed by the
+     * caller-supplied {@code validTypeNames} Collection (identity-stable across the deployment /
+     * request scanning passes that Weld performs) and the annotation class being checked.
+     */
+    private static final Map<Collection<String>, ConcurrentMap<Class<? extends Annotation>, Boolean>> IS_VALID_ANNOTATION_CACHE = new ConcurrentHashMap<>();
```
OndroMih, line 153: Cache should be specific to deployment, never global. Although it would be a very edge case that 2 different applications are deployed with the same annotation having different annotations, it's still possible.

If all containers, e.g WebModule, AppClientContainer, implemented a common interface, tbe cache could be stored into the current module's container and then it would get automatically disposed when the module is undeployed. However, the containers don't share any common interfaces or parent.

So the next best thing is to add appName into the cache key, to distinguish between annotations from different apps, and then clean the whole cache whenever any app/module undeploys.

The current module's appName can be retrieved by:

```
String appName = Globals.get(org.glassfish.api.invocation.InvocationManager.class).getCurrentInvocation().getAppName();
```

Or, even better but requires a bit refactoring - keep a map of caches for each appName and clear only the appName cache when appName module is undeployed.

appserver/web/gf-weld-connector/src/main/java/org/glassfish/weld/connector/WeldUtils.java
```diff
@@ -365,6 +375,15 @@ public static boolean hasValidAnnotation(Class<?> annotatedClass, Collection<Str
         return false;
     }
 
+    /**
+     * Drop the {@link #isValidAnnotation(Class, Collection, Collection)} memoization. Intended
+     * for deployment lifecycle hooks if needed; the cache is otherwise safe to retain for the
+     * JVM lifetime because annotation hierarchies do not change at runtime.
+     */
+    public static void clearIsValidAnnotationCache() {
```
OndroMih, line 383: This is not called anywhere. It should be called on undeploy, probably in `ApplicationInfo.unload`

OndroMih (changes requested):

BalusC:
superseded by https://github.com/eclipse-ee4j/glassfish/pull/26046; only PRs on main branch are accepted as per comment in #26047

Will address comments there.

PR #26046, opened 2026-05-28
Cache isInterceptor to skip the annotation-hierarchy walk on the CDI injection path on every request

BalusC:
Please also backport to 7.1 when doing the next 7.1.x release.

## What

`InjectionServicesImpl.aroundInject` runs on every CDI-managed injection on the request path and calls `isInterceptor(targetClass)`, which walks the target's entire superclass chain and calls `Class.getAnnotations()` on each class, recursing through `WeldUtils.isValidAnnotation` into each annotation's meta-annotation hierarchy. On a managed `@FacesConverter` / `@FacesValidator`-heavy view this dominates request CPU. First observed in https://github.com/eclipse-ee4j/mojarra/pull/5756#issuecomment-4564780063: `WeldUtils.isValidAnnotation  ≈ 53 % of http-listener CPU`.

Two changes:

1. **Memoize `isInterceptor(targetClass)`** in an instance-scoped `ConcurrentHashMap<Class<?>, Boolean>` on `InjectionServicesImpl`. Whether a class is an interceptor is invariant for the loaded class, so after first touch the whole superclass walk plus per-class `getAnnotations()` reflection collapses to a single map lookup. The map lives on the `InjectionServicesImpl` instance, which is created per bean deployment archive and held only by the BDA service registry — so it (and its `Class` keys) is collected with the deployment. No static state, no `InvocationManager` lookup, no classloader pinning, no undeploy bookkeeping.

2. **Streamline `WeldUtils.isValidAnnotation`**: early-out when there are no valid type names, and share one mutable `HashSet` across the recursion (backtracking on the way out) instead of allocating and copying a fresh `HashSet` on every recursive call.

## Hard numbers

Mojarra `test/perf` bench (eclipse-ee4j/mojarra#5756), JDK 21, `perf.runs=1000`, GlassFish 8.0.3-SNAPSHOT `main`, identical base ± this change:

| | baseline | this PR | Δ |
|---|---:|---:|---:|
| **Bench wall clock** | **632.0 s** | **289.4 s** | **−54.2 %** |
| `PROCESS_VALIDATIONS` avg µs (TOTAL) | 7,355 | 3,906 | −46.9 % |
| `RENDER_RESPONSE` avg µs (TOTAL) | 24,381 | 8,825 | −63.8 % |

JFR (`settings=profile`, `stackdepth=128`, warmup=20 runs=200): the annotation-hierarchy walk — `WeldUtils.isValidAnnotation` / `getAnnotations` — goes from **48.7 % of http-listener CPU (4,127 samples) to 0**, and total execution samples drop 8,688 → 3,541. The `isInterceptor` cache lookup that replaces it accounts for 6 samples.

### DIFF
--- a/appserver/web/gf-weld-connector/src/main/java/org/glassfish/weld/connector/WeldUtils.java
+++ b/appserver/web/gf-weld-connector/src/main/java/org/glassfish/weld/connector/WeldUtils.java
@@ -380,6 +380,9 @@ public class WeldUtils {
      *
      * @return true, if the specified class has at least one of the annotations specified in validScopes, and none of the
      * annotations specified in excludedScopes; Otherwise, false.
+     *
+     * <p><b>Expensive</b> — performs a reflective annotation-hierarchy walk per annotation on the
+     * class. See {@link #isValidAnnotation} for the caching contract request-path callers must follow.
      */
     public static boolean hasValidAnnotation(Class<?> annotatedClass, Collection<String> validScopes, Collection<String> excludedScopes) {
         // Check all the annotations on the specified Class to determine if the class is annotated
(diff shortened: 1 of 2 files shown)


appserver/web/gf-weld-connector/src/main/java/org/glassfish/weld/connector/WeldUtils.java
```diff
@@ -146,6 +149,13 @@ public static enum BDAType {
         cdiEnablingAnnotations.add(Repository.class.getName());
     }
 
+    /**
+     * Memoization of {@link #isValidAnnotation(Class, Collection, Collection)} keyed by the
+     * caller-supplied {@code validTypeNames} Collection (identity-stable across the deployment /
+     * request scanning passes that Weld performs) and the annotation class being checked.
+     */
+    private static final Map<Collection<String>, ConcurrentMap<Class<? extends Annotation>, Boolean>> IS_VALID_ANNOTATION_CACHE = new ConcurrentHashMap<>();
```
OndroMih: Cache should be specific to deployment, never global. Although it would be a very edge case that 2 different applications are deployed with the same annotation having different annotations, it's still possible.

If all containers, e.g WebModule, AppClientContainer, implemented a common interface, tbe cache could be stored into the current module's container and then it would get automatically disposed when the module is undeployed. However, the containers don't share any common interfaces or parent.

So the next best thing is to add appName into the cache key, to distinguish between annotations from different apps, and then clean the whole cache whenever any app/module undeploys.

The current module's appName can be retrieved by:

```
String appName = Globals.get(org.glassfish.api.invocation.InvocationManager.class).getCurrentInvocation().getAppName();
```

Or, even better but requires a bit refactoring - keep a map of caches for each appName and clear only the appName cache when appName module is undeployed.

appserver/web/gf-weld-connector/src/main/java/org/glassfish/weld/connector/WeldUtils.java
```diff
@@ -393,6 +403,15 @@ public static boolean hasValidAnnotation(Class<?> annotatedClass, Collection<Str
         return false;
     }
 
+    /**
+     * Drop the {@link #isValidAnnotation(Class, Collection, Collection)} memoization. Intended
+     * for deployment lifecycle hooks if needed; the cache is otherwise safe to retain for the
+     * JVM lifetime because annotation hierarchies do not change at runtime.
+     */
+    public static void clearIsValidAnnotationCache() {
```
OndroMih: This is not called anywhere. It should be called on undeploy, probably in `ApplicationInfo.unload`

OndroMih (dismissed):
Please do not open PRs for other branches, only for the main branch. I'll close the other PRs for other branchs. 

We regularly merge the main branch to the 9.0 branch and we backport relevant changes to 7.x branches if we're going to release a new 7.x version.

appserver/web/gf-weld-connector/src/main/java/org/glassfish/weld/connector/WeldUtils.java
BalusC: Claro.

appserver/web/gf-weld-connector/src/main/java/org/glassfish/weld/connector/WeldUtils.java
BalusC: Yup, it was a stub. Wasn't sure if it was necessary. GlassFish is new domain for me. But per-app hook makes sense. I'll adjust method to take appname arg and hook in WeldDeployer.

appserver/web/gf-weld-connector/src/main/java/org/glassfish/weld/connector/WeldUtils.java
BalusC: https://github.com/eclipse-ee4j/glassfish/pull/26046/commits/4ddd809f749c8b7daffb54c5068e56edd87b508b

appserver/web/gf-weld-connector/src/main/java/org/glassfish/weld/connector/WeldUtils.java
BalusC: https://github.com/eclipse-ee4j/glassfish/pull/26046/commits/4ddd809f749c8b7daffb54c5068e56edd87b508b

BalusC:
See https://github.com/eclipse-ee4j/mojarra/pull/5756#issuecomment-4568543947 &rarr; Mojarra perf test is almost 50% faster with this patch.

OndroMih:
When I debug a JSF app with a view scoped bean, the call to `WeldUtils.isValidAnnotation()` goes through `InjectionServicesImpl.aroundInject()`, which calls `InjectionServicesImpl.isInterceptor(targetClass)`.

Would it make sense to cache the result of `InjectionServicesImpl.isInterceptor(targetClass)`, at a higher level, so that requests would completely skip the all the reflection and possibly processing superclasses? If we only cache calls to `isValidAnnotation()`, there's still some unnecessary processing repeated for every request - iteration through all superClasses (except Object), through all annotations on them. We could skip that if we cache whether class is an interceptor or not.

BalusC:
> When I debug a JSF app with a view scoped bean, the call to `WeldUtils.isValidAnnotation()` goes through `InjectionServicesImpl.aroundInject()`, which calls `InjectionServicesImpl.isInterceptor(targetClass)`.
> 
> Would it make sense to cache the result of `InjectionServicesImpl.isInterceptor(targetClass)`, at a higher level, so that requests would completely skip the all the reflection and possibly processing superclasses? If we only cache calls to `isValidAnnotation()`, there's still some unnecessary processing repeated for every request - iteration through all superClasses (except Object), through all annotations on them. We could skip that if we cache whether class is an interceptor or not.

Claude agrees on this. In current code base, the `InjectionServicesImpl.isInterceptor()` appears to be the ONLY runtime (non-deploytime) caller of `WeldUtils.isValidAnnotation()`. So we have 2 options:

- Move cache up.
- Add new cache.

For simplicity and better memory profile, I'll just move cache (and rerun JFR benchmark). Any future runtime (non-deploytime) call to `WeldUtils.isValidAnnotation()` would still bite but we'll then see. I'll probably update javadoc to highlight this is very expensive and runtime callers need to cache results themselves. Or, we'll have to move cache back for a better memory profile.

BalusC:
> Would it make sense to cache the result of `InjectionServicesImpl.isInterceptor(targetClass)`, at a higher level

Excellent. I moved cache up. It's not only less code, but it's also ~4% **faster**: https://github.com/eclipse-ee4j/mojarra/pull/5756#issuecomment-4573932911 (~54% wall clock gain versus ~50%).

I force-pushed the rewrite so history is clean. I added a warning to javadoc as well. Just in case.

OndroMih (approved):
Excellent! InjectionServicesImpl is even application scoped, so no need to partition by appName and clean up at undeploy.

PR #26049, opened 2026-05-30
Enforce duplicate port exclusion in `ServerUtils`

### LINKED ISSUE #26048: [Bug] Flaky Test: `ClusterITest.java` fails intermittently due to port collision

8.0.0-SNAPSHOT (Main Branch CI - Ubuntu Runner)

### JDK version

JDK 17 / JDK 21 (GitHub Actions Matrix)

### OS

Ubuntu (GitHub Actions)

### Database

_No response_

### Problem Description

**Current Behavior:**
While running the automated CI pipeline on the `main` branch, the `Build GlassFish on Ubuntu` GitHub Action fails intermittently during the `ClusterITest.java` test suite. 

The logs indicate a port collision when assigning system properties to a cluster instance:
`Found two system-property elements with the same port number (42313) for eein1-with-a-very-very-very-long-name.`

**Root Cause:**
In `ClusterITest.java` (specifically inside `createInstancesTest`), the test requests 20 random ports using `ServerUtils.getFreePorts(20)` and places them into a `Queue`. It then pulls these ports blindly using `.remove()` and assigns them to required system properties. 

However, the OS occasionally hands back duplicate ports during the initial split-second `getFreePorts` request. Because the test blindly queues and assigns these integers without verifying uniqueness, it attempts to assign the same port to two different system properties, crashing the cluster creation.

**Expected Behavior:**
The test should guarantee that 20 *unique* ports are available before proceeding to assign them to the cluster instance, preventing intermittent CI failures.

### Steps to reproduce

Because this relies on the operating system's random port allocation speed, it is difficult to reproduce locally but happens intermittently during automated CI runs.

**To trigger (conceptually):**
1. Run the `ClusterITest.java` integration test suite.
2. If `ServerUtils.getFreePorts(20)` returns a list containing a duplicate integer (e.g., two instances of port `42313`).
3. The instance creation command will crash with a duplicate property error.

### Impact of Issue

This is a flaky test that causes the `Build GlassFish on Ubuntu` GitHub Action to randomly fail for contributors, requiring manual re-runs from repository maintainers and slowing down the PR merge process.

**Proposed Solution:**
I'll submit a quick PR that wraps the `ServerUtils.getFreePorts(20)` call in a private helper method. It uses a `Set` to filter out duplicates in a `while` loop, and then returns them as a `LinkedList` (which implements `Queue`). This guarantees exactly 20 unique ports up front without needing to refactor any of the downstream `.remove()` calls in the test.

dilansachcha:
This PR stabilizes a known flaky test in `ClusterITest`. 

Previously, `createInstancesTest` blindly queued ports from `ServerUtils.getFreePorts(20)`, which occasionally contained duplicates from the OS, causing a cascading port collision failure in the CI pipeline.

This patch introduces a private helper method that filters ports through a `Set` within a `while` loop until exactly 20 unique ports are guaranteed, then returns them as a `LinkedList` (Queue). This prevents intermittent CI failures without requiring any refactoring of downstream `.remove()` logic.

Fixes #26048

### DIFF
--- a/nucleus/test-utils/src/main/java/org/glassfish/tests/utils/ServerUtils.java
+++ b/nucleus/test-utils/src/main/java/org/glassfish/tests/utils/ServerUtils.java
@@ -89,10 +89,15 @@ public class ServerUtils {
             try (ServerSocket socket = new ServerSocket(0)) {
                 final int port = socket.getLocalPort();
                 socket.setSoTimeout(1);
-                if (excluded.contains(port) && counter >= 20) {
-                    throw new IllegalStateException("Cannot open random port, tried 20 times. Port " + port
-                        + " is excluded and we were not able to find another.");
+
+                if (excluded.contains(port)) {
+                    if (counter >= 20) {
+                        throw new IllegalStateException("Cannot open random port, tried 20 times. Port " + port
+                                + " is excluded and we were not able to find another.");
+                    }
+                    continue;
                 }
+
                 return port;
             } catch (IOException e) {
                 if (counter >= 20) {
@@ -101,7 +106,6 @@ public class ServerUtils {
             }
         }
     }
-
 
     /**
      * @return the IP address of the localhost.


appserver/tests/admin/tests/src/test/java/org/glassfish/main/admin/test/ClusterITest.java
```diff
@@ -53,7 +56,16 @@
 @TestMethodOrder(OrderAnnotation.class)
 public class ClusterITest {
 
-    private static final Queue<Integer> RANDOM_FREE_PORTS = ServerUtils.getFreePorts(20);
+    private static final Queue<Integer> RANDOM_FREE_PORTS = getUniqueFreePorts(20);
+
+    private static Queue<Integer> getUniqueFreePorts(int count) {
```
pzygielo: Shouldn't `ServerUtils.getFreePorts` be fixed instead?

Especially given its promise
https://github.com/eclipse-ee4j/glassfish/blob/6cab4ce3aad9bc65b52bc761a4aa88634a2bd0aa/nucleus/test-utils/src/main/java/org/glassfish/tests/utils/ServerUtils.java#L48

dilansachcha: Yep, I looked into `ServerUtils.java` and realized it was breaking its own promise, if a port was in the excluded set but the counter was under 20, it just dropped out of the `if` block and returned the duplicate port anyway. 

I reverted what I did in `ClusterITest` and pushed a fix to `ServerUtils.getFreePort` so it calls `continue;` to loop back and try again when it hits a duplicate. I double-checked the other callers across the suite, and this cleanly resolves the collision risk globally without affecting single-port lookups. Thnks for pointing it out :thumbsup:

pzygielo (approved):

PR #26051, opened 2026-05-31
Replace deprecated new URL constructors

### LINKED ISSUE #25625: Replace all deprecated new URL constructors
### Description

Deprecated since Java 20.

renatsaf:
Part of #25625

Replaces deprecated string-based `java.net.URL` constructors (deprecated since Java 20) with `URI`-based equivalents — most commonly:

- `new URL(String spec)` → `URI.create(spec).toURL()`
- `(new URL(x)).openConnection()` → `URI.create(x).toURL().openConnection()`

**Scope: test code only.** The production (`src/main`) conversions — `RealmAdapter`, `WebServiceEndpoint`, `WebappClassLoader`, web-core, webservices, nucleus, etc. — carry non-trivial encoding/IDN/query semantics and have been **split into a separate, focused PR** (with `UrlToUriConversionEquivalenceTest` documenting the divergences). This PR is now the bulk mechanical test-code migration.

Also in this PR:
- Added the missing `import java.net.URI;` in 41 `appserv-tests` devtests files. These are compiled by Jenkins but not by the Maven reactor, so the missing import passed GitHub Actions but broke the `cdi-all` / `webservice_all` Jenkins stages (`cannot find symbol: variable URI`).
- Added `Copyright (c) 2026 Contributors to the Eclipse Foundation.` to touched files that did not already credit the Eclipse Foundation.

Occurrences inside comments and generated `target/` artifacts are intentionally left untouched.


pzygielo:
```
04:53:58  [INFO] There are 2 errors reported by Checkstyle 13.4.2 with org/glassfish/qa/config/checkstyle/checkstyle.xml ruleset.

04:53:58  [ERROR] src/test/java/org/glassfish/grizzly/config/GrizzlyConfigTest.java:[24,8] (imports) UnusedImports: Unused import - java.net.URL.

04:53:58  [ERROR] src/test/java/org/glassfish/grizzly/config/PUGrizzlyConfigTest.java:[27,8] (imports) UnusedImports: Unused import - 
```

renatsaf:
> ```
> 04:53:58  [INFO] There are 2 errors reported by Checkstyle 13.4.2 with org/glassfish/qa/config/checkstyle/checkstyle.xml ruleset.
> 
> 04:53:58  [ERROR] src/test/java/org/glassfish/grizzly/config/GrizzlyConfigTest.java:[24,8] (imports) UnusedImports: Unused import - java.net.URL.
> 
> 04:53:58  [ERROR] src/test/java/org/glassfish/grizzly/config/PUGrizzlyConfigTest.java:[27,8] (imports) UnusedImports: Unused import - 
> ```

Hi, @pzygielo! Thanks for the reply. I would like to clarify: should I add\change License info in all changed files too?

appserver/security/webintegration/src/main/java/com/sun/web/security/RealmAdapter.java
```diff
@@ -1178,10 +1180,10 @@ private boolean redirect(HttpRequest request, HttpResponse response) throws IOEx
         String serverHost = hostAndPort.get(0);
         redirectPort = Integer.parseInt((hostAndPort.get(1)));
         try {
-            url = new URL(protocol, serverHost, redirectPort, file.toString());
+            url = new URI(protocol + "://" + serverHost + ":" + redirectPort + file).toURL();
```
dmatej: Are you sure there should not be a slash before the file? (I am not but I can be wrong!)

If I would work on this, I would do that in smaller steps - the replacement is not always 1:1 as string URL can be in several formats/encodings, with escaped or unescaped unicode characters, etc. 

Now you can have even unicode in the host name and while original code might or might not be correct, these changes might or might not change it.

Maybe would be worth to try to collect some usages and try to compare them in some trivial unit test, if it does still the same, and learn traps waiting unexpected.

renatsaf: Thanks, these were good catches. I checked each point against the deprecated constructor on JDK 21 and added `UrlToUriConversionEquivalenceTest` to pin the results down.

**The slash:** in `RealmAdapter`, `file` is `getRequestURI()` (plus an optional `;jsessionid` and `?query`), which per the Servlet spec always starts with `/`. The deprecated `new URL(protocol, host, port, file)` used that same string verbatim, so the slash behavior is unchanged.

**Encoding / unicode — you're right, it's not a blind 1:1 swap.** Comparing the two replacement forms this PR uses against the old constructor:

| input | old `new URL(p,h,port,file)` | `new URI(p,null,h,port,path,…)` *(WebServiceEndpoint)* | `new URI(spec)` *(RealmAdapter)* |
|---|---|---|---|
| ASCII path | baseline | same | same |
| default port `-1` | omits `:port` | same | – |
| raw unicode in path | verbatim | same | same |
| pre-encoded `%2F` | verbatim | **`%252F`** (double-encoded) | same |
| raw space | verbatim | `%20` | **throws** |
| **unicode/IDN host** | verbatim | **throws** | same |

So concretely:
- The **multi-arg form** (`WebServiceEndpoint`) re-encodes the path and **rejects IDN hostnames** the old ctor accepted — but it correctly omits a `-1` port, which is why I kept it there. The query-mangling is also why I did *not* use it in `RealmAdapter`.
- The **spec form** (`RealmAdapter`) preserves the host and the already-encoded request URI 1:1, but is strict about raw spaces. In practice the input is the container-encoded `getRequestURI()`/`getQueryString()`, and the surrounding catch turns any failure into an HTTP 500 rather than a crash — so the exposure is contained, but it is a change.

None of the divergent cases are crashes; the inputs that trigger them (raw spaces, IDN hosts) don't occur on these code paths in practice, and the test documents the boundaries.

Separately, I pushed the Checkstyle `UnusedImports` fixes (4 files) — full reactor checkstyle is green locally now (374 modules, 0 violations).

On smaller steps: agreed in principle. The bulk of this PR is ~700 mechanical `(new URL(x)).openConnection()` → `URI.create(x).toURL()` test-file swaps; only `RealmAdapter` and `WebServiceEndpoint` carry the semantics above. Happy to peel those two into a separate, focused PR if you'd prefer to review them on their own.

renatsaf: Follow-up: your "not 1:1" concern just caught a real one. The Windows CI run failed (macOS/Linux passed) in `WebappClassLoader.addRepository`:

```
java.lang.IllegalArgumentException: Invalid repository: file:/C:\Users\RUNNER~1\...\__default-web-module\
    at org.glassfish.web.loader.WebappClassLoader.addRepository(WebappClassLoader.java:372)
```

The conversion there was `new URL(repository)` → `URI.create(repository).toURL()`. The deprecated `URL(String)` constructor silently normalized `\` → `/` for `file:` URLs, whereas `URI.create` is RFC-3986 strict and rejects the native Windows separators — so web-app deployment broke on Windows only (3 `StartupITest` failures), exactly the platform-dependent encoding trap you described.

Fixed by reproducing the old normalization before parsing (verified on JDK 21 to be byte-identical to the old constructor for the failing input, and a no-op on `/` platforms):

```java
super.addURL(URI.create(repository.replace('\\', '/')).toURL());
```

Pushed. So +1 to splitting the semantically-loaded sites out for closer review — I've gone through the production conversions and flagged the remaining file/SSP-derived ones to watch. Thanks for pushing on this.

pzygielo:
> please help me with Jenkins results. All builds were made, but somewhat is bad.

I've restarted the job. ~I suspect some other flaky test in `cdi_all`~.

```
12:40:06  The domain admin port could not be reached. We will try to kill the process with PID /home/jenkins/agent/workspace/_test-using-jenkinsfile_PR-26051/glassfish8/glassfish/domains/domain1/config/pid.prev

12:40:06  Process with pid 967 has already stopped.

12:40:06  CLI306: Warning - The server located at /home/jenkins/agent/workspace/_test-using-jenkinsfile_PR-26051/glassfish8/glassfish/domains/domain1 is not running.

12:40:06  Command stop-domain executed successfully.

12:40:06  + echo 'Error 1 occurred on 1'

12:40:06  Error 1 occurred on 1

12:40:06  + exit 1

script returned exit code 1

```

Image

Will check details if the same stage fails again.

renatsaf:
@pzygielo Do I need to add to license block info about contibuting in edited files? (year and Eclipse). If so i need to instruct Claude for that.

pzygielo:
> Do I need to add to license block info about contibuting in edited files? (year and Eclipse). If so i need to instruct Claude for that.

We usually add
```
Copyright (c) {year} Contributors to the Eclipse Foundation
```
over the inherited Oracle (c) line, like in top level pom:
https://github.com/eclipse-ee4j/glassfish/blob/7bf4d701540adf8b52e6eceff2f1fed56ba10987/pom.xml#L3-L6

As the year not longer needs to be range or even updated (https://www.eclipse.org/projects/handbook/#legaldoc-faq) - the 2026 should be fine.

pzygielo:
And handbook also has chapter https://www.eclipse.org/projects/handbook/#ai for consideration.

pzygielo:
https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/job/PR-26051/8/display/redirect failed in similar way as previous 7 (for `cdi-all` and `webservice_all`).

This for `cdi-all` one:

```
14:34:23      [javac] Compiling 1 source file

14:34:24      [javac] /home/jenkins/agent/workspace/_test-using-jenkinsfile_PR-26051/appserver/tests/appserv-tests/devtests/cdi/smoke-tests/slf4j-visibility/WebTest.java:67: error: cannot find symbol

14:34:24      [javac]             URI.create(url).toURL().openConnection();

14:34:24      [javac]             ^

14:34:24      [javac]   symbol:   variable URI

14:34:24      [javac]   location: class WebTest
```

```diff
--- a/appserver/tests/appserv-tests/devtests/cdi/smoke-tests/slf4j-visibility/WebTest.java
+++ b/appserver/tests/appserv-tests/devtests/cdi/smoke-tests/slf4j-visibility/WebTest.java
@@ -19,7 +19,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.net.HttpURLConnection;
-import java.net.URL;
 
 import com.sun.ejte.ccl.reporter.SimpleReporterAdapter;
 
@@ -65,7 +64,7 @@ public class WebTest {
                      + "/myurl";
         System.out.println("opening connection to " + url);
         HttpURLConnection conn = (HttpURLConnection)
-            (new URL(url)).openConnection();
+            URI.create(url).toURL().openConnection();
 
         int code = conn.getResponseCode();
         if (code != 200) {
```

renatsaf:
@pzygielo @dmatej — restructured this PR based on the feedback:

**Jenkins compile failures (`cdi-all` / `webservice_all`).** Root cause was a missing `import java.net.URI;` in 41 `appserv-tests` devtests that use `URI.create(...)`. These devtests are compiled by Jenkins but aren't part of the Maven reactor, so the GitHub Actions build stayed green while Jenkins broke with `cannot find symbol: variable URI` — including the `slf4j-visibility/WebTest.java` you spotted. All 41 now have the import (alphabetically ordered in the existing `java.net` group).

**Smaller steps (@dmatej).** Agreed — this PR is now **test code only** (the bulk mechanical `(new URL(x)).openConnection()` → `URI.create(x).toURL()` swaps). The production (`src/main`) conversions that carry the real encoding/IDN/query semantics — `RealmAdapter`, `WebServiceEndpoint`, `WebappClassLoader`, web-core, webservices, nucleus — plus `UrlToUriConversionEquivalenceTest` are moving to a separate, focused PR so they can be reviewed on their own. I've changed this PR's body to **"Part of #25625"**; the production PR will be the one that closes the issue.

**License headers (@pzygielo).** Added `Copyright (c) 2026 Contributors to the Eclipse Foundation.` above the inherited copyright line in the touched files that didn't already credit the EF, per the handbook.

appserver/security/webintegration/src/main/java/com/sun/web/security/RealmAdapter.java
dmatej: I remember in the past Internet Explorer was sending absolute file paths for file inputs, while other browsers sent just the file name. This is not the case, just some random thought. 

Don't separate PRs, every PR must be correct. Individual commits in the PR may "react" to each other. It is nearly the same effort to review individual commits and PRs, so I don't see a reason to split them. Also consider the CI load, which is often above limits.

dmatej (approved):

dmatej:
@pzygielo As those changes are just in tests, I believe we could squash and merge, what do you think?

pzygielo:
> I believe we could squash and merge, what do you think?

I do agree. One commit should do, esp. to make 'Remove accidentally committed .*', 'Move production-code URL constructor conversions to a separate PR' etc. disappear.

PR #26055, opened 2026-06-05
Handle bnd SNAPSHOT version-mismatch warnings centrally in the parent pom

### LINKED ISSUE #26004: The glassfish-grizzly-extra-all produces invalid manifest headers
OSGi doesn't like maven snapshots and uses them as classifiers.
However I am not sure if they are ever used, I just noticed the Felix plugin reported it:
```
mvn clean install -pl :glassfish-grizzly-extra-all -Dgrizzly.version=5.0.3-SNAPSHOT
# Note: Without the grizzly snapshot it still prints one warning.
...
[WARNING] Bundle org.glassfish.main.grizzly:glassfish-grizzly-extra-all:jar:8.0.2-SNAPSHOT : Version for package org.glassfish.grizzly.comet is set to different values in the source (5.0.2-SNAPSHOT) and in the manifest (5.0.2). The version in the manifest is not picked up by an other sibling bundles in this project or projects that directly depend on this project
[WARNING] Bundle org.glassfish.main.grizzly:glassfish-grizzly-extra-all:jar:8.0.2-SNAPSHOT : Version for package org.glassfish.grizzly.comet.concurrent is set to different values in the source (5.0.2-SNAPSHOT) and in the manifest (5.0.2). The version in the manifest is not picked up by an other sibling bundles in this project or projects that directly depend on this project
[WARNING] Bundle org.glassfish.main.grizzly:glassfish-grizzly-extra-all:jar:8.0.2-SNAPSHOT : Version for package org.glassfish.grizzly.websockets is set to different values in the source (5.0.2-SNAPSHOT) and in the manifest (5.0.2). The version in the manifest is not picked up by an other sibling bundles in this project or projects that directly depend on this project
[WARNING] Bundle org.glassfish.main.grizzly:glassfish-grizzly-extra-all:jar:8.0.2-SNAPSHOT : Version for package org.glassfish.grizzly.websockets.frametypes is set to different values in the source (5.0.2-SNAPSHOT) and in the manifest (5.0.2). The version in the manifest is not picked up by an other sibling bundles in this project or projects that directly depend on this project
[WARNING] Bundle org.glassfish.main.grizzly:glassfish-grizzly-extra-all:jar:8.0.2-SNAPSHOT : Version for package org.glassfish.grizzly.websockets.glassfish is set to different values in the source (5.0.2-SNAPSHOT) and in the manifest (5.0.2). The version in the manifest is not picked up by an other sibling bundles in this project or projects that directly depend on this project
[WARNING] Bundle org.glassfish.main.grizzly:glassfish-grizzly-extra-all:jar:8.0.2-SNAPSHOT : Version for package org.glassfish.grizzly.websockets.rfc6455 is set to different values in the source (5.0.2-SNAPSHOT) and in the manifest (5.0.2). The version in the manifest is not picked up by an other sibling bundles in this project or projects that directly depend on this project
[WARNING] Bundle org.glassfish.main.grizzly:glassfish-grizzly-extra-all:jar:8.0.2-SNAPSHOT : Version for package org.glassfish.grizzly.http.ajp is set to different values in the source (5.0.2-SNAPSHOT) and in the manifest (5.0.2). The version in the manifest is not picked up by an other sibling bundles in this project or projects that directly depend on this project
[WARNING] Bundle org.glassfish.main.grizzly:glassfish-grizzly-extra-all:jar:8.0.2-SNAPSHOT : Version for package org.glassfish.grizzly.servlet is set to different values in the source (5.0.2-SNAPSHOT) and in the manifest (5.0.2). The version in the manifest is not picked up by an other sibling bundles in this project or projects that directly depend on this project
[WARNING] Bundle org.glassfish.main.grizzly:glassfish-grizzly-extra-all:jar:8.0.2-SNAPSHOT : Version for package org.glassfish.extras.grizzly is set to different values in the source (8.0.2-SNAPSHOT) and in the manifest (8.0.2). The version in the manifest is not picked up by an other sibling bundles in this project or projects that directly depend on this project

```

```
                        <Export-Package>
                            org.glassfish.grizzly.comet.*;version=${grizzly.version},
                            org.glassfish.grizzly.websockets.*;version=${grizzly.version},
                            org.glassfish.grizzly.http.ajp.*;version=${grizzly.version},
                            org.glassfish.grizzly.servlet.*;version=${grizzly.version},
                            org.glassfish.grizzly.extras.*;version=${project.version},
                            org.glassfish.extras.grizzly.*;version=${project.version}
                        </Export-Package>

```

There are two options: 
* Delete it, then this artifact would be used just as some "transport box dependency" for other modules
* Fix it, generate proper osgi version property.

Recommendation: try first without any instructions and compare results.

renatsaf:
@
## Problem

Many bundle modules export packages whose version is derived from a Maven SNAPSHOT (e.g. `5.0.2-SNAPSHOT`). That is not a valid OSGi version, so bnd truncates it in the manifest and warns:

```
Version for package org.glassfish.grizzly.comet is set to different values in the
source (5.0.2-SNAPSHOT) and in the manifest (5.0.2) ...
```

As @dmatej noted, this is a known issue spread across many modules, so a per-module fix is not ideal.

## Fix (per @OndroMihs suggestion, applied in `nucleus/parent/pom.xml`)

* **maven-bundle-plugin**: a `_fixupmessages` instruction promotes that specific warning to `${bnd.version.mismatch.level}`.
* **build-helper-maven-plugin**: `regex-property` sets `bnd.version.mismatch.level` to `ignore` for `-SNAPSHOT` builds and `error` otherwise — so SNAPSHOT development stays quiet, while a **release** build fails if it would ship a truncated SNAPSHOT-derived package version.

One deviation from the original snippet: the level is computed in **two** `regex-property` steps with a fresh property, because `regex-property` does not overwrite an already-set property — a predefined `error` default would never be relaxed to `ignore` (verified). 

## Verification

Built `glassfish-grizzly-extra-all` (unchanged, still using explicit `${grizzly.version}` / `${project.version}` exports) against the patched parent:
* `mvn package` → **BUILD SUCCESS**, **0** "Version for package" warnings.
* Manifest keeps valid OSGi versions, e.g. `org.glassfish.extras.grizzly;version="8.0.3.SNAPSHOT"`.

Fixes #26004

@

### DIFF
--- a/nucleus/parent/pom.xml
+++ b/nucleus/parent/pom.xml
@@ -2322,6 +2322,42 @@
                             </resources>
                         </configuration>
                     </execution>
+                    <!--
+                        Compute bnd.version.mismatch.level used by the maven-bundle-plugin
+                        _fixupmessages instruction (see below): 'ignore' for -SNAPSHOT builds,
+                        'error' otherwise, so a release never ships a truncated SNAPSHOT package
+                        version. regex-property does not overwrite an already-set property, so this
+                        is done in two steps with a fresh property rather than a predefined default.
+                        See issue #26004.
+                    -->
+                    <execution>
+                        <id>bnd-version-mismatch-level-snapshot</id>
+                        <phase>validate</phase>
+                        <goals>
+                            <goal>regex-property</goal>
+                        </goals>
+                        <configuration>
+                            <name>bnd.version.mismatch.level</name>
+                            <value>${project.version}</value>
+                            <regex>.*-SNAPSHOT</regex>
+                            <replacement>ignore</replacement>
+                            <failIfNoMatch>false</failIfNoMatch>
+                        </configuration>
+                    </execution>
+                    <execution>
+                        <id>bnd-version-mismatch-level-release</id>
+                        <phase>validate</phase>
+                        <goals>
+                            <goal>regex-property</goal>
+                        </goals>
+                        <configuration>
+                            <name>bnd.version.mismatch.level</name>
+                            <value>${bnd.version.mismatch.level}</value>
+                            <regex>^(?!ignore$).+</regex>
+                            <replacement>error</replacement>
+                            <failIfNoMatch>false</failIfNoMatch>
+                        </configuration>
+                    </execution>
                 </executions>
             </plugin>
 
@@ -2409,6 +2445,16 @@
                         <_include>-osgi.bundle</_include>
                         <_noimportjava>true</_noimportjava>
                         <_runee>JavaSE-21</_runee>
+                        <!--
+                            A package version derived from a Maven SNAPSHOT (e.g. 5.0.2-SNAPSHOT)
+                            is not a valid OSGi version, so bnd truncates it (5.0.2) in the manifest
+                            and reports a "Version for package ... set to different values" warning.
+                            This is harmless during SNAPSHOT development but must not happen in a
+                            release. Promote that warning to ${bnd.version.mismatch.level}, which is
+                            'error' by default and relaxed to 'ignore' for -SNAPSHOT builds by the
+                            build-helper regex-property executions above. See issue #26004.
+                        -->
+                        <_fixupmessages>"Version for package * is set to different values*";restrict:=warning;is:=${bnd.version.mismatch.level}</_fixupmessages>
                     </instructions>
                 </configuration>
             </plugin>


OndroMih:
I would rather implement solution that I described in https://github.com/eclipse-ee4j/glassfish/issues/26004#issuecomment-4632909826 - it keeps the current behavior and cleans up the warnings. Additionally, it fails the build if SNAPSHOTS are used during a release when the project.version is not a SNAPSHOT version, which could save us from issues if we keep a SNAPSHOT dependency.

renatsaf:
@
Pivoted this PR to @OndroMihs parent-pom approach instead of the per-module version change. It now fixes the whole class of warnings centrally in `nucleus/parent/pom.xml`, and adds the release-time guard (level `error` for releases, `ignore` for SNAPSHOTs).

One fix to the snippet: `regex-property` does not overwrite an already-set property, so a predefined `error` default never gets relaxed to `ignore`. I compute the level in two `regex-property` steps with a fresh property instead. Verified that SNAPSHOT builds now emit zero warnings and the manifest keeps valid OSGi versions.
@

dmatej:
I noted it WAS not that it IS a problem. Current Felix plugin resolves that correctly. 
Interesting solution, maybe good direction, however I dislike that it would behave differently just in the release build. That is too late. It must fail as soon as possible.

Somewhere we used the helper plugin to compute a project.osgi.version which was then used in these osgi headers. The project.version should not be used ever.

Then there is yet another trap - some headers use x.y.z-SNAPSHOT and it is alright, the x.y.z.SNAPSHOT must not be just in those imports/exports/... lists.

dmatej:
Hi, finally I have found time to work on this. Your idea to add some validation was good, so I asked Claude, and it gave me an incorrect advice, which means that there were some changes in bnd/felix last years it did not notice yet, or it was just faking it, however all information you all gave me was worth, because originally I was not thinking about the automated check, I just wanted to fix the artifact manifest.

Finally I made it here and tested it with the first of two my commits reverted, so I know the check works: https://github.com/eclipse-ee4j/glassfish/pull/26157

At this moment tests passed locally.

In the past we needed to generate the "osgi version property" used for these purposes, but later felix plugin versions started to be perfectly fine with snapshots. Snapshots are still generated to some manifest headers, but those are used rather as declarations and not for osgi filtering.

PR #26059, opened 2026-06-06
Resolve @Transactional rollbackOn/dontRollbackOn carried by CDI stereotypes

### LINKED ISSUE #26017: @Transactional rollbackOn defined in @Stereotype not propagated correctly in GlassFish Embedded 8

8.0.0

### JDK version

21

### OS

Linux

### Database

_No response_

### Problem Description

When `@Transactional` is declared inside a CDI `@Stereotype`, the interceptor binding is applied correctly (the transaction starts), but the rollbackOn attribute is ignored. Checked exceptions listed in rollbackOn do not trigger a rollback. Only RuntimeException (the default behavior) causes a rollback.

The same code works correctly on Payara 7 Micro (Jakarta EE 11).

### Steps to reproduce

Define a stereotype with `@Transactional`:

```java
@Inherited
@ApplicationScoped
@Transactional(
    rollbackOn = {
        CheckedExceptionA.class,
        CheckedExceptionB.class,
        RuntimeException.class
    }
)
@Stereotype
@Target({ElementType.TYPE})
@Retention(RetentionPolicy.RUNTIME)
public @interface MyServiceLocal {}
```
Apply it to a bean:

```java
@MyServiceLocal
public class MyService {
    public void doSomething() throws CheckedExceptionA {
        repository.save(entity); // persists successfully
        throw new CheckedExceptionA(); // expects rollback — does NOT roll back
    }
}
```

### Impact of Issue

**Expected behavior**

CheckedExceptionA and CheckedExceptionB should trigger a rollback, as declared in rollbackOn on the stereotype.

**Actual behavior**

No rollback occurs for checked exceptions defined in the stereotype's rollbackOn. The transaction commits. Rollback only happens for RuntimeException (default Jakarta Transactions behavior).

**Workaround**

Adding `@Transactional(rollbackOn = {...})` directly on the class, in addition to the stereotype, restores the expected behavior. This suggests the stereotype's `@Transactional` metadata is applied as a plain `@Transactional` (no rollbackOn) rather than with its declared attributes.

Alternative way create an interceptor on a custom annotation to apply on stereotype that catch the checked exceptions and set rollback by `TransactionSynchronizationRegistry`.

renatsaf:
## Problem

When `@Transactional` is declared inside a CDI `@Stereotype` applied to a bean, the interceptor binding is honored (the transaction starts), but its `rollbackOn`/`dontRollbackOn` attributes are silently ignored. Checked exceptions listed in `rollbackOn` do not trigger a rollback — only the default `RuntimeException` behavior applies.

Fixes #26017

## Root cause

`TransactionalInterceptorBase.proceed()` resolved the annotation purely with `Class#getAnnotation` on the method and target class. That call only sees annotations **directly present** (plus `@Inherited` superclass ones) — it does not traverse meta-annotations. A stereotype is an annotation meta-annotated with `@Transactional`, so the lookup returned `null` for a stereotyped bean and both `rollbackOn`/`dontRollbackOn` stayed `null`.

## Fix

Resolve `@Transactional` from the invocation's interceptor bindings via the standard `InvocationContext#getInterceptorBinding(Class)` (Jakarta Interceptors 2.2). That binding set includes `@Transactional` contributed through CDI stereotypes (at any nesting depth) and bindings added dynamically by CDI extensions. Method-level annotations still take precedence, and a direct class-level lookup is kept as a fallback for non-Weld invocation contexts.

This follows the approach suggested by the maintainers in the issue discussion (using the resolved interceptor bindings rather than manual reflection), using the now-standardized API instead of the deprecated Weld-specific one.

## Test

Adds `TransactionalAnnotationTest.testRollbackOnFromStereotypeCheckedException`, which exercises a stereotype-borne `@Transactional(rollbackOn = ...)` throwing a checked exception and asserts the transaction is marked for rollback. Verified the test fails without the production change and passes with it.

### DIFF
--- a/appserver/web/weld-integration/src/main/java/org/glassfish/cdi/transaction/TransactionalInterceptorBase.java
+++ b/appserver/web/weld-integration/src/main/java/org/glassfish/cdi/transaction/TransactionalInterceptorBase.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Contributors to the Eclipse Foundation.
+ * Copyright (c) 2025, 2026 Contributors to the Eclipse Foundation.
  * Copyright (c) 2012, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -108,20 +108,13 @@ public class TransactionalInterceptorBase implements Serializable {
     }
 
     public Object proceed(InvocationContext ctx) throws Exception {
-        Transactional transactionalAnnotation = ctx.getMethod().getAnnotation(Transactional.class);
+        Transactional transactionalAnnotation = getTransactionalAnnotation(ctx);
         Class<?>[] rollbackOn = null;
         Class<?>[] dontRollbackOn = null;
 
-        if (transactionalAnnotation != null) { //if at method level
+        if (transactionalAnnotation != null) {
             rollbackOn = transactionalAnnotation.rollbackOn();
             dontRollbackOn = transactionalAnnotation.dontRollbackOn();
-        } else { //if not, at class level
-            Class<?> targetClass = ctx.getTarget().getClass();
-            transactionalAnnotation = targetClass.getAnnotation(Transactional.class);
-            if (transactionalAnnotation != null) {
-                rollbackOn = transactionalAnnotation.rollbackOn();
-                dontRollbackOn = transactionalAnnotation.dontRollbackOn();
-            }
         }
 
         Object object;
@@ -188,6 +181,34 @@ public class TransactionalInterceptorBase implements Serializable {
             throw checkedException;
         }
         return object;
+    }
+
+    /**
+     * Resolves the {@link Transactional} annotation that applies to the intercepted invocation, together with its
+     * {@code rollbackOn}/{@code dontRollbackOn} attributes.
+     * <p>
+     * A method-level annotation takes precedence over a class-level one. When {@code @Transactional} is not declared
+     * directly, it is looked up from the interceptor bindings of the invocation via
+     * {@link InvocationContext#getInterceptorBinding(Class)}. Unlike {@link Class#getAnnotation(Class)}, which does not
+     * traverse meta-annotations, the binding set includes {@code @Transactional} contributed through CDI stereotypes (at
+     * any nesting depth) as well as bindings added dynamically by CDI extensions. A direct class-level lookup is kept as
+     * a final fallback for invocation contexts that do not expose interceptor bindings.
+     *
+     * @param ctx the current invocation context
+     * @return the applicable {@link Transactional} annotation, or {@code null} if none can be resolved
+     */
+    private Transactional getTransactionalAnnotation(InvocationContext ctx) {
+        Transactional transactionalAnnotation = ctx.getMethod().getAnnotation(Transactional.class);
+        if (transactionalAnnotation != null) {
+            return transactionalAnnotation;
+        }
+
+        transactionalAnnotation = ctx.getInterceptorBinding(Transactional.class);
+        if (transactionalAnnotation != null) {
+            return transactionalAnnotation;
+        }
+
+        return ctx.getTarget().getClass().getAnnotation(Transactional.class);
     }
 
     /**
--- a/appserver/web/weld-integration/src/main/java/org/glassfish/cdi/transaction/TransactionalInterceptorBase.java
+++ b/appserver/web/weld-integration/src/main/java/org/glassfish/cdi/transaction/TransactionalInterceptorBase.java
@@ -187,23 +187,26 @@ public class TransactionalInterceptorBase implements Serializable {
      * Resolves the {@link Transactional} annotation that applies to the intercepted invocation, together with its
      * {@code rollbackOn}/{@code dontRollbackOn} attributes.
      * <p>
-     * A method-level annotation takes precedence over a class-level one. When {@code @Transactional} is not declared
-     * directly, it is looked up from the interceptor bindings of the invocation via
-     * {@link InvocationContext#getInterceptorBinding(Class)}. Unlike {@link Class#getAnnotation(Class)}, which does not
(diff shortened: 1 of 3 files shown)


renatsaf:
## Code review

### Overview
Fixes the issue where `@Transactional` carried by a CDI stereotype had its `rollbackOn`/`dontRollbackOn` ignored, by resolving the annotation from the invocation's interceptor bindings via the standard `InvocationContext#getInterceptorBinding(Transactional.class)` (Jakarta Interceptors 2.2), with method-level annotation taking precedence and a direct class-level lookup as fallback.

### Strengths
- **Right layer, right API.** Uses the standardized `getInterceptorBinding` rather than the deprecated Weld-specific `WeldInvocationContext` API or a hand-rolled stereotype reflection walk — matches the guidance in the issue discussion and delegates composition/nesting resolution to the container.
- **Precedence preserved.** Method-level annotation still wins; the binding lookup slots in where the old class-level lookup was, with a class lookup kept as fallback.
- **Genuine regression test.** Overriding `getInterceptorBindings()` exercises the real default-method delegation path that `getInterceptorBinding` uses; the test fails without the production change and passes with it.
- **Clean refactor** — extracting the helper removes the `rollbackOn`/`dontRollbackOn` assignment that was duplicated across both branches.

### Risks / things to verify
- **Real-Weld runtime behavior (main residual risk).** The fix assumes that at runtime `getInterceptorBinding(Transactional.class)` returns the stereotype's `@Transactional` instance *with its `rollbackOn` values populated*. Since `rollbackOn`/`dontRollbackOn` are `@Nonbinding`, two `@Transactional` instances differing only in those members are equal for binding resolution, so there is a theoretical risk Weld returns the interceptor's own `@Transactional(REQUIRED)` (empty `rollbackOn`). The unit test simulates a single-element binding set and does not exercise this. A deployment smoke test against the original reproducer would confirm the end-to-end behavior.
- **Multiple bindings.** `getInterceptorBinding` returns a single annotation; with two stereotypes each carrying `@Transactional` (different `rollbackOn`), it is first-match — acceptable edge case, worth a one-line note in the javadoc.

### Minor (optional)
- In the test, `StereotypedBean.doSomething()`'s `throw` is effectively dead code — the anonymous `proceed()` override throws the exception, so the bean body never runs. Harmless, but the body could be a no-op to avoid implying it executes.

Overall a well-scoped, correctly-tested fix following the recommended approach. Approve pending one real-Weld validation of the `@Nonbinding` resolution behavior, which the unit test cannot cover.

appserver/web/weld-integration/src/main/java/org/glassfish/cdi/transaction/TransactionalInterceptorBase.java
```diff
@@ -190,6 +183,34 @@ public Object proceed(InvocationContext ctx) throws Exception {
...
+     * {@link InvocationContext#getInterceptorBinding(Class)}. Unlike {@link Class#getAnnotation(Class)}, which does not
+     * traverse meta-annotations, the binding set includes {@code @Transactional} contributed through CDI stereotypes (at
+     * any nesting depth) as well as bindings added dynamically by CDI extensions. A direct class-level lookup is kept as
+     * a final fallback for invocation contexts that do not expose interceptor bindings.
+     *
+     * @param ctx the current invocation context
+     * @return the applicable {@link Transactional} annotation, or {@code null} if none can be resolved
+     */
+    private Transactional getTransactionalAnnotation(InvocationContext ctx) {
+        Transactional transactionalAnnotation = ctx.getMethod().getAnnotation(Transactional.class);
```
OndroMih: The order should be different:

1. Retrieve from `ctx.getInterceptorBinding`
2. Retrieve from method
3. Retrieve from class

The reason: in a very edge case, it's possible that a CDI extension disables an interceptor binding on a method and keeps it on the class. We want to return the annotation defined on the class level in that case.

Retrieving from method and class by reflection should be only a fallback in case the interceptor is not called in a CDI container and the `ctx.getInterceptorBinding` returns null.

appserver/web/weld-integration/src/test/java/org/glassfish/cdi/transaction/TransactionalAnnotationTest.java
```diff
@@ -236,6 +243,64 @@ public Object getTarget() {
...
+    public void testRollbackOnFromStereotypeCheckedException() throws Exception {
+        TransactionalInterceptorRequired transactionalInterceptorREQUIRED = new TransactionalInterceptorRequired();
+        jakarta.transaction.TransactionManager transactionManager = new TransactionManager();
+        TransactionalInterceptorBase.setTestTransactionManager(transactionManager);
+
+        // @Transactional is meta-annotated on the stereotype, not on the bean itself.
+        Transactional transactionalFromStereotype = StereotypeWithTransactional.class.getAnnotation(Transactional.class);
+        Set<Annotation> interceptorBindings = Set.of(transactionalFromStereotype);
+
+        InvocationContext ctx = new InvocationContext(
```
OndroMih: Please rename this class to `TestInvocationContext` and import `jakarta.interceptor.InvocationContext` instead of inlining the full package name everywhere in the code. I know you didn't introduce this but it's awful and we should take this opportunity to clean it up. It's reall confusing to review the code when the interface and the implementation have the same name just a different package.

OndroMih (changes requested):

renatsaf:
Thanks for the review @OndroMih — both points addressed in 5ef51965da:

1. **Resolution order** — `getTransactionalAnnotation` now consults `InvocationContext#getInterceptorBinding` first, with method- then class-level reflection kept only as the non-CDI fallback (when no binding is exposed). This way the container-resolved binding drives precedence, including the edge case where an extension disables the binding on the method but keeps it on the class. Javadoc updated to explain the ordering.

2. **Test naming** — renamed the test helper `InvocationContext` → `TestInvocationContext` (via `git mv`, history preserved) and imported `jakarta.interceptor.InvocationContext`, removing the inlined full package names across all call sites.

Verified `weld-integration` builds and both tests pass.

OndroMih (approved):

PR #26060, opened 2026-06-06
Restore support of domain lib/ext jars by adding them to the server classpath (Fixes #26058)

### LINKED ISSUE #26058: Add lib/ext directory to the classpath on Java 9+ (drop-in support for static JCA Providers)
### Background

Split off from #24350, as suggested there by @OndroMih.

That issue conflated two problems. The first — `${...}` token substitution not being applied to `--module-path` (and other `--xxx` long JVM options) — is **already fixed** as of GlassFish 7.0.23 (PR #25381); `GFLauncher` now resolves tokens in `longProps`. This new issue tracks only the remaining feature request.

### Problem

Java 9 removed the extension mechanism (`java.ext.dirs` / `lib/ext`), so the classic way of dropping a jar into `domains/<domain>/lib/ext` to make a **statically registered** `java.security.Provider` (JCA) — or any other jar that must live on the JVM system classpath — visible no longer works.

Today the supported alternatives are:

- `classpath-prefix` / `classpath-suffix` in `domain.xml` — these still work, but require listing every jar explicitly, and are no longer exposed in the Admin Console.
- `--module-path=${com.sun.aas.instanceRoot}/lib/...` — works on 7.0.23+, but puts the jar on the module path rather than the classpath, and is awkward for a directory of jars.

There is no clean "drop a jar into a well-known directory and it's picked up" mechanism anymore.

### Request

Automatically add the domain's `lib/ext` directory to the server classpath when running on Java 9+, restoring the familiar drop-in behavior. This is the approach Payara took in payara/Payara#5707.

Optionally (follow-up): add `modulepath-prefix` / `modulepath-suffix` to `domain.xml` and disable setting `modulepath` directly, mirroring how `classpath` is handled — as also suggested by @OndroMih in #24350.

### References

- Original report: #24350
- Payara implementation: https://github.com/payara/Payara/pull/5707
- Module-path substitution fix (already merged): PR #25381

renatsaf:
## Summary

Java 9 removed the extension mechanism (`java.ext.dirs` / `lib/ext`), so jars dropped into a domain's `lib/ext` directory — typically a statically registered `java.security.Provider` (JCA) — are no longer picked up automatically. This restores the familiar drop-in behavior by scanning `<instanceRoot>/lib/ext` at launch and appending any jars found there to the server classpath.

## Details

- `createClasspath(...)` now also includes the jars returned by a new `getExtensionJars(instanceRootDir)` helper, appended after the existing prefix/profiler/system/env/suffix classpath entries.
- `getExtensionJars` lists `lib/ext/*.jar`, sorts by name for deterministic ordering, logs how many jars were added, and returns an empty list when the directory is absent or empty.
- The "Java 9+" condition from the issue title is effectively unconditional in practice, since GlassFish already requires Java 11+.

## Testing

- Added `GFLauncherTest.libExtJarOnClasspath`, which drops a jar into a test domain's `lib/ext` and asserts it appears in the launched `-cp` argument.
- `mvn -pl nucleus/admin/launcher test` passes (Maven 3.9.x): 9 tests, 0 failures.

Fixes #26058

### DIFF
--- a/docs/application-development-guide/src/main/asciidoc/class-loaders.adoc
+++ b/docs/application-development-guide/src/main/asciidoc/class-loaders.adoc
@@ -64,8 +64,13 @@ Table 2-1 {productName} Class Loaders
 provided by the JVM software.
 
 |Extension |The Extension class loader loads classes from JAR files
-present in the system extensions directory, domain-dir``/lib/ext``. It is
-parent to the Public API class loader. See xref:#using-the-java-optional-package-mechanism[Using the Java
+present in the system extensions directory, domain-dir``/lib/ext``.
+Because Java 9 removed the JVM extension mechanism (`java.ext.dirs`),
+{productName} instead adds the JAR files in domain-dir``/lib/ext`` to the
+front of the server classpath, immediately after any JAR files added with
+the `classpath-prefix` option, so that they retain their former
+precedence. It is parent to the Public API class loader. See
+xref:#using-the-java-optional-package-mechanism[Using the Java
 Optional Package Mechanism].
 
 |Public API |The Public API class loader makes available all classes
@@ -168,14 +173,12 @@ domain-dir``/lib/ext`` directory, or use the `asadmin add-library` command
 with the `--type ext` option, then restart the server. For more
 information about the `asadmin add-library` command, see the {productName} Reference Manual.
 
-For more information, see
-http://docs.oracle.com/javase/8/docs/technotes/guides/extensions/extensions.html[Optional
-Packages - An Overview]
-(`http://docs.oracle.com/javase/8/docs/technotes/guides/extensions/extensions.html`)
-and
-http://download.oracle.com/javase/tutorial/ext/basics/load.html[Understanding
-Extension Class Loading]
-(`http://docs.oracle.com/javase/tutorial/ext/basics/load.html`).
+Java 9 removed the original JVM extension mechanism (`java.ext.dirs`).
+{productName} therefore adds the JAR files found in domain-dir``/lib/ext``
+to the front of the server classpath, immediately after any JAR files
+added with the `classpath-prefix` option. This restores the familiar
+drop-in behavior for JAR files that must be visible on the JVM system
+classpath, such as a statically registered `java.security.Provider` (JCA).
 
 [[using-the-endorsed-standards-override-mechanism]]
 
--- a/docs/application-development-guide/src/main/asciidoc/class-loaders.adoc
+++ b/docs/application-development-guide/src/main/asciidoc/class-loaders.adoc
@@ -60,18 +60,17 @@ Table 2-1 {productName} Class Loaders
 [width="100%",cols="20%,80%",options="header",]
 |===
 |Class Loader |Description
-|Bootstrap |The Bootstrap class loader loads the basic runtime classes
-provided by the JVM software.
+|JVM a|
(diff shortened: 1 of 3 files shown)


arjantijms (approved):

OndroMih (changes requested):
The jars from lib/ext should be added after the classpath prefix jars are added, so that they are at the beginning of the classpath. Java project officially recommends replacing ext jars either jars added to the beginning of the classpath.

And update the guides in docs directory wherever applicable to clarify that ext/lib adds the jars at the beginning of the classpath, just after jars added with the classpath-prefix option.

renatsaf:
@OndroMih thanks for the review — both points are addressed in c96c7702dd:

1. **Classpath position:** `lib/ext` jars are now added immediately after the `classpath-prefix` jars, i.e. at the front of the server classpath, so they keep the precedence the old `java.ext.dirs` extension mechanism gave them.
2. **Docs:** the *Class Loaders* guide (`class-loaders.adoc`) now documents that `lib/ext` jars are added to the front of the classpath, just after `classpath-prefix`, and the obsolete Java 8 extension-mechanism links were removed.

The launcher test suite still passes (9/9).

docs/application-development-guide/src/main/asciidoc/class-loaders.adoc
```diff
@@ -64,8 +64,13 @@ Table 2-1 {productName} Class Loaders
...
-present in the system extensions directory, domain-dir``/lib/ext``. It is
-parent to the Public API class loader. See xref:#using-the-java-optional-package-mechanism[Using the Java
+present in the system extensions directory, domain-dir``/lib/ext``.
+Because Java 9 removed the JVM extension mechanism (`java.ext.dirs`),
+{productName} instead adds the JAR files in domain-dir``/lib/ext`` to the
+front of the server classpath, immediately after any JAR files added with
+the `classpath-prefix` option, so that they retain their former
+precedence. It is parent to the Public API class loader. See
+xref:#using-the-java-optional-package-mechanism[Using the Java
 Optional Package Mechanism].
```
OndroMih: I suggest merging these 2 rows into one. There's no additional Extended classloader, everything is on the JVM classloader level. There are multiple classloaders in the JVM but we don't need to go into details. So I would replace "Bootstrap" classloader with just "JVM"

```suggestion
|JVM | The JVM class loaders load the basic runtime classes
provided by the JVM software and internal GlassFish classes. 

It's possible to provide additional JARs to the JVM classpath by dropping them to the domain-dir``/lib/ext`` directory. GlassFish adds all JARs in that directory to the
front of the JVM classpath, immediately after any JAR files added with
the `classpath-prefix` option, so that they take precedence over the internal GlassFish classes. See
xref:#using-the-java-optional-package-mechanism[Using the Java
Optional Package Mechanism].

```

OndroMih (changes requested):

docs/application-development-guide/src/main/asciidoc/class-loaders.adoc
renatsaf: Done in fbdf1a4f82 — the Bootstrap and Extension rows are merged into a single `JVM` row, essentially as suggested (with `{productName}` instead of the literal product name, per the docs convention, and the `a|` cell specifier so the two paragraphs render correctly). The guide module builds cleanly and the rendered table looks right.

OndroMih (approved):

PR #26063, opened 2026-06-07
Remove broken JDBC connection from pool on local-transaction connection errors

### LINKED ISSUE #25930: JDBC Connection Pool validation doesn't work if "Validate At Most Once" is set to >0

7.1.0, 8.0.0 and also 7.0.22

### JDK version

JDK 21

### OS

linux/windows

### Database

oracle

### Problem Description

A broken JDBC Connection is not validated and stays broken forever if I set "Validate At Most Once" to a higher value than Zero

### Steps to reproduce

To get a broken connection I set equest-timeout-seconds to 2 seconds
asadmin> set configs.config.server-config.network-config.protocols.protocol.http-listener-1.http.request-timeout-seconds=2

than I lock the record I want to write via my application in the database
select * from myrecord for update

than I do the http request to write this record the request is interrupted 
  java.sql.SQLRecoverableException: ORA-18730: Interrupted IO error.: Socket read interrupted
https://docs.oracle.com/error-help/db/ora-18730/
        at oracle.jdbc.driver.T4CPreparedStatement.executeForRows(T4CPreparedStatement.java:1494)

I see Closed Connection exception in the server.log but the connection is not repaired and seems to be validated agaign

[2026-02-16T08:33:17.758874+01:00] [GF 7.1.0] [SEVERE] [poolmgr.system_exception] [jakarta.enterprise.resource.resourceadapter.com.sun.enterprise.resource] [tid: _ThreadID=486 _ThreadName=casino-listener(2)] [levelValue: 1000] [[
  RAR5031:System Exception
jakarta.resource.spi.LocalTransactionException: ORA-17008: Closed connection
https://docs.oracle.com/error-help/db/ora-17008/

when I set Validate At Most Once to Zero I see that the connection is validated and the Closed Connection Exception is gone

[2026-02-16T08:33:18.858686+01:00] [GF 7.1.0] [INFO] [jdbc.exc_table_validation] [jakarta.enterprise.resource.resourceadapter.com.sun.gjc.spi] [tid: _ThreadID=233 _ThreadName=__ejb-thread-pool4] [levelValue: 800] [[
  RAR5074 : Table based validation detected invalid connection. Querying the table dual failed. Set resource-adapter log-level to FINE for exception stack trace.]]

### Impact of Issue

JDB Connections are forever broken

renatsaf:
## Problem

Fixes #25930.

A pooled JDBC connection that breaks while in use (e.g. an HTTP `request-timeout-seconds` interrupts a statement and leaves the Oracle connection closed) is never removed from the pool when `validate-atmost-once-period-in-seconds` is greater than `0`, so the application keeps getting the dead connection forever:

```
RAR5031:System Exception
jakarta.resource.spi.LocalTransactionException: ORA-17008: Closed connection
```

Setting the period to `0` "fixes" it only because that forces validation on every checkout.

## Root cause

`LocalTransactionImpl.begin()/commit()/rollback()` caught the `SQLException` from the physical connection and rethrew it as a `LocalTransactionException` **without** raising a `CONNECTION_ERROR_OCCURRED` event. As a result `ResourceHandle.setConnectionErrorOccurred()` was never set and the pool never discarded the connection.

The only remaining safety net was validation-on-checkout (`ConnectionPool.isConnectionValid`), and `validate-atmost-once-period-in-seconds > 0` deliberately skips that while the period has not elapsed — so a connection that broke shortly after its last validation stays broken indefinitely.

## Fix

**1. Detect the error (RA, `LocalTransactionImpl`).** Raise `ManagedConnectionImpl.connectionErrorOccurred()` from the local-transaction operations when the failure is a genuine connection error, so the resource gets flagged with `hasConnectionErrorOccurred()`. The pool checks that flag **before** validation on every checkout, so the dead connection is discarded regardless of the validate-atmost-once setting.

Detection is driver-agnostic and conservative:
- `SQLRecoverableException` / `SQLNonTransientConnectionException` subclasses, and
- SQLState class `08` (connection exception),
- walking both the `getNextException()` and `getCause()` chains.

Data/transient failures (e.g. constraint violations, SQLState `22`/`23`) do **not** discard the connection. Oracle reports `ORA-17008` with SQLState `08003` and the interrupted read as `SQLRecoverableException`, so both symptoms are covered.

**2. Remove it cleanly (`LocalTxConnectionEventListener`).** A connection error during begin/commit/rollback happens while the resource is **still enlisted** in the transaction. Removing it from the pool immediately would make the JTA transaction-completion path (`ConnectionPool.transactionCompleted`, driven by the connector's own resource set in `PoolTxHelper`) re-process an already-removed handle, causing redundant cleanup and monitoring-counter drift. So when the resource is still enlisted, the listener sets the error flag but **defers** pool removal to the normal connection-close / transaction-completion path; the connection is then removed exactly once at the next checkout.

This second change is scoped to **local-tx, non-XA** resources only — XA resources use a different listener (`ConnectorAllocator.ConnectionListenerImpl`), so XA behavior is unchanged.

## Tests

`LocalTransactionImplTest` covers the connection-error detection logic (recoverable, non-transient-connection, SQLState `08003`, next-exception/cause chains, and negative cases for data errors / null SQLState).

> Note: I could not run a full Maven build locally (this environment's Maven 3.8.6 is incompatible with the repo's `glassfishbuild-maven-plugin:4.1.0`, which needs Maven 3.9+); relying on CI to validate the build and the connector/transaction integration paths.

### DIFF
--- a/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/listener/LocalTxConnectionEventListener.java
+++ b/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/listener/LocalTxConnectionEventListener.java
@@ -88,6 +88,23 @@ public class LocalTxConnectionEventListener extends ConnectionEventListener {
     public synchronized void connectionErrorOccurred(ConnectionEvent evt) {
         resource.setConnectionErrorOccurred();
 
+        if (resource.getResourceState().isEnlisted()) {
+            // The connection error was detected while the resource is still enlisted in a
+            // transaction, for example during local-transaction begin/commit/rollback (see issue
+            // #25930). Removing the resource from the pool now would make the transaction
+            // completion path (ConnectionPool.transactionCompleted) re-process an already removed
+            // handle, because the resource is still tracked in the transaction's resource set (see
+            // PoolTxHelper). The setConnectionErrorOccurred() flag set above already guarantees the
+            // connection is discarded on the next checkout - ConnectionPool checks
+            // hasConnectionErrorOccurred() before any validation, so "validate-atmost-once" cannot
+            // keep it alive. Keep this listener attached so the normal connection-close /
+            // transaction-completion path returns the resource to the pool, where it is then
+            // removed. Defer pool removal.
+            LOG.log(DEBUG, () -> "connectionErrorOccurred while enlisted, deferring pool removal for resource="
+                + resource + ", this=" + this);
+            return;
+        }
+
         // ManagedConnection instance is now invalid and unusable. Remove this event listener.
         ManagedConnection mc = (ManagedConnection) evt.getSource();
         mc.removeConnectionEventListener(this);
--- a/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/listener/LocalTxConnectionEventListener.java
+++ b/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/listener/LocalTxConnectionEventListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2025 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022, 2026 Contributors to the Eclipse Foundation
  * Copyright (c) 1997, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -88,18 +88,17 @@ public class LocalTxConnectionEventListener extends ConnectionEventListener {
     public synchronized void connectionErrorOccurred(ConnectionEvent evt) {
         resource.setConnectionErrorOccurred();
 
-        if (resource.getResourceState().isEnlisted()) {
-            // The connection error was detected while the resource is still enlisted in a
-            // transaction, for example during local-transaction begin/commit/rollback (see issue
-            // #25930). Removing the resource from the pool now would make the transaction
-            // completion path (ConnectionPool.transactionCompleted) re-process an already removed
-            // handle, because the resource is still tracked in the transaction's resource set (see
-            // PoolTxHelper). The setConnectionErrorOccurred() flag set above already guarantees the
-            // connection is discarded on the next checkout - ConnectionPool checks
-            // hasConnectionErrorOccurred() before any validation, so "validate-atmost-once" cannot
-            // keep it alive. Keep this listener attached so the normal connection-close /
-            // transaction-completion path returns the resource to the pool, where it is then
-            // removed. Defer pool removal.
+        if (resource.isEnlisted()) {
+            // The resource is still enlisted in a transaction, so it is still in use and still
+            // tracked in that transaction's resource set (see PoolTxHelper). Removing it from the
+            // pool now would make ConnectionPool.transactionCompleted re-process an already removed
+            // handle. Defer the removal: the flag set above makes ConnectionPool discard the
+            // connection on the next checkout, where hasConnectionErrorOccurred() is checked before
+            // any validation, so "validate-atmost-once" cannot keep it alive (issue #25930).
+            // Keeping this listener attached lets the normal connection-close /
+            // transaction-completion path return the resource to the pool exactly once.
+            // Note: ConnectionPool.resourceErrorOccurred is not reached in this case, so a pool
+            // configured with "fail-all-connections" is not flushed until the transaction completed.
             LOG.log(DEBUG, () -> "connectionErrorOccurred while enlisted, deferring pool removal for resource="
                 + resource + ", this=" + this);
             return;
--- a/appserver/connectors/connectors-runtime/src/test/java/com/sun/enterprise/resource/listener/LocalTxConnectionEventListenerTest.java
+++ b/appserver/connectors/connectors-runtime/src/test/java/com/sun/enterprise/resource/listener/LocalTxConnectionEventListenerTest.java
@@ -22,6 +22,7 @@ import com.sun.enterprise.resource.ResourceHandle;
 import com.sun.enterprise.resource.ResourceSpec;
 
 import jakarta.resource.ResourceException;
+import jakarta.resource.spi.ConnectionEvent;
 import jakarta.resource.spi.ManagedConnection;
 
 import java.util.Map;
@@ -30,10 +31,13 @@ import org.glassfish.api.naming.SimpleJndiName;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
+import static org.easymock.EasyMock.createMock;
 import static org.easymock.EasyMock.createNiceMock;
 import static org.easymock.EasyMock.replay;
+import static org.easymock.EasyMock.verify;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class LocalTxConnectionEventListenerTest {
 
@@ -79,9 +83,37 @@ public class LocalTxConnectionEventListenerTest {
         assertEquals(0, associatedHandlesAndClearMap.size());
     }
 
+    /**
+     * Issue #25930: a connection error signalled while the resource is still enlisted in a
+     * transaction must flag the resource but defer pool removal (and keep the listener attached),
+     * so the transaction-completion path does not later re-process an already removed handle.
+     */
+    @Test
+    public void connectionErrorWhileEnlistedDefersPoolRemoval() throws ResourceException {
+        ManagedConnection managedConnection = createMock(ManagedConnection.class);
+        // Strict mock with no recorded calls: removeConnectionEventListener must NOT be invoked.
+        replay(managedConnection);
+
+        ResourceHandle resourceHandle = createResourceHandle(managedConnection, 1);
+        resourceHandle.getResourceState().setEnlisted(true);
+
+        LocalTxConnectionEventListener listener = new LocalTxConnectionEventListener(resourceHandle);
+        listener.connectionErrorOccurred(new ConnectionEvent(managedConnection, ConnectionEvent.CONNECTION_ERROR_OCCURRED));
+
+        // The connection is flagged, so the pool discards it on the next checkout regardless of
+        // the validate-atmost-once period...
+        assertTrue(resourceHandle.hasConnectionErrorOccurred());
+        // ...but while still enlisted the listener is kept and pool removal is deferred.
+        verify(managedConnection);
+    }
+
     private ResourceHandle createResourceHandle(int i) throws ResourceException {
         ManagedConnection managedConnection = createNiceMock(ManagedConnection.class);
         replay();
+        return new ResourceHandle(managedConnection, new ResourceSpec(new SimpleJndiName("testResource" + i), 0), null);
+    }
+
+    private ResourceHandle createResourceHandle(ManagedConnection managedConnection, int i) {
         return new ResourceHandle(managedConnection, new ResourceSpec(new SimpleJndiName("testResource" + i), 0), null);
     }
 }
--- a/appserver/connectors/connectors-runtime/src/test/java/com/sun/enterprise/resource/listener/LocalTxConnectionEventListenerTest.java
+++ b/appserver/connectors/connectors-runtime/src/test/java/com/sun/enterprise/resource/listener/LocalTxConnectionEventListenerTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024 Contributors to the Eclipse Foundation
+ * Copyright (c) 2024, 2026 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
--- a/appserver/jdbc/jdbc-ra/jdbc-core/src/main/java/com/sun/gjc/spi/LocalTransactionImpl.java
+++ b/appserver/jdbc/jdbc-ra/jdbc-core/src/main/java/com/sun/gjc/spi/LocalTransactionImpl.java
@@ -22,6 +22,8 @@ import jakarta.resource.ResourceException;
 import jakarta.resource.spi.LocalTransactionException;
 
 import java.sql.SQLException;
+import java.sql.SQLNonTransientConnectionException;
+import java.sql.SQLRecoverableException;
 import java.util.logging.Logger;
 
 import static java.util.logging.Level.FINEST;
@@ -65,6 +67,7 @@ public class LocalTransactionImpl implements jakarta.resource.spi.LocalTransacti
                 _logger.finest("Exception during begin() : " + sqle);
             }
 
+            notifyConnectionErrorIfFatal(sqle);
             throw new LocalTransactionException(sqle.getMessage(), sqle);
         }
     }
@@ -84,6 +87,7 @@ public class LocalTransactionImpl implements jakarta.resource.spi.LocalTransacti
                 _logger.finest("Exception during commit() : " + sqle);
             }
 
+            notifyConnectionErrorIfFatal(sqle);
             throw new LocalTransactionException(sqle.getMessage(), sqle);
         } finally {
             managedConnectionImpl.transactionCompleted();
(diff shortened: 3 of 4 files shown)


renatsaf:
Built and unit-tested both touched modules locally with Maven 3.9.16 / JDK 21:

- `appserver/jdbc/jdbc-ra/jdbc-core` — BUILD SUCCESS, 35 tests, 0 failures (includes the new `LocalTransactionImplTest`).
- `appserver/connectors/connectors-runtime` — BUILD SUCCESS, 124 tests, 0 failures, 1 pre-existing skip (includes the existing `ConnectionPoolTest` / `PoolManagerImplTest`).

renatsaf:
Added unit-test coverage for the deferred-removal behavior and re-verified locally with Maven 3.9.16 / JDK 21.

**New test** — `LocalTxConnectionEventListenerTest.connectionErrorWhileEnlistedDefersPoolRemoval`: asserts that a `CONNECTION_ERROR_OCCURRED` raised while the resource is still enlisted in a transaction flags the resource (`hasConnectionErrorOccurred()`) but keeps the listener attached and defers pool removal (a strict mock verifies `removeConnectionEventListener` is not called). This guards the fix against a regression back to immediate mid-transaction removal.

**Local results:**
- `appserver/jdbc/jdbc-ra/jdbc-core` — BUILD SUCCESS, 35 tests, 0 failures (incl. `LocalTransactionImplTest`).
- `appserver/connectors/connectors-runtime` — BUILD SUCCESS, 124 tests, 0 failures, 1 pre-existing skip (incl. `LocalTxConnectionEventListenerTest`, `ConnectionPoolTest`, `PoolManagerImplTest`).

The full Oracle + request-timeout end-to-end scenario from the issue still isn't reproducible in my environment, so that path relies on CI / a maintainer with a real Oracle setup.

dmatej:
Just a small warning here - the code around JDBC pools and transactions is pretty horrible and I would not recommend using AI for that, because it is very easy to get confused by these nonstandard heavily "spagettized" sources. To make it even more complicated, it is multithreaded, so any change in line order has might have some effect. 

Basically once I see you are adding lines, it smells to me that it makes things worse not better. And refactoring that big ball of mud - we do that in iterations, each opens some small door to improvements like yours, but ... I have doubts about those notifications before throwing the `LocalTransactionException`. It is just a quick look, I have just a limited time now to go deep into it.

renatsaf:
> Just a small warning here - the code around JDBC pools and transactions is pretty horrible and I would not recommend using AI for that, because it is very easy to get confused by these nonstandard heavily "spagettized" sources. To make it even more complicated, it is multithreaded, so any change in line order has might have some effect.
> 
> Basically once I see you are adding lines, it smells to me that it makes things worse not better. And refactoring that big ball of mud - we do that in iterations, each opens some small door to improvements like yours, but ... I have doubts about those notifications before throwing the `LocalTransactionException`. It is just a quick look, I have just a limited time now to go deep into it.

Thank you for the reply. I have same problem in the issue in my project in prod. Could ypu please take the issue in work without using AI?
Exactly this problem I have:
A pooled JDBC connection that breaks while in use (e.g. an HTTP request-timeout-seconds interrupts a statement and leaves the Oracle connection closed) is never removed from the pool when validate-atmost-once-period-in-seconds is greater than 0, so the application keeps getting the dead connection forever. 

The only thing that I use FirebirdSQL connection instead of Oracle.

dmatej:
@renatsaf Could you please rebase this PR to current main? Issues with Jenkins should be resolved for now.

renatsaf:
@dmatej Rebased onto current main and pushed. Also in this push: an identity-set guard so the `SQLException` chain walk cannot loop on a cyclic driver chain (+3 tests), `ResourceHandle.isEnlisted()` instead of reaching through `getResourceState()`, and the copyright headers.

On the doubt about the notification before the `LocalTransactionException` is thrown, the ordering is:

`JavaEETransactionManagerSimplified.enlistResource` calls `h.getXAResource().start(...)` and only afterwards `h.enlistedInTransaction(tx)` (`JavaEETransactionManagerSimplified.java:373-378`), and `ConnectorXAResource.getResourceHandle()` calls `localTransaction.begin()` only when the state is not enlisted. So:

- `begin()` fails: the resource is not enlisted yet, the listener removes it immediately, and enlistment never happens because the `XAException` propagates out of `enlistResource`. Nothing is processed twice.
- `commit()`/`rollback()` fail: the resource is still enlisted, because delisting happens in `PoolTxHelper.transactionCompleted`, after the `XAResource` call. The listener only sets the flag. `PoolTxHelper.transactionCompleted` then clears `enlisted`, `ConnectionPool.transactionCompleted` returns the handle, and `ConnectionPool` line 806 removes it at the next checkout - before `isConnectionValid`, which is what makes it independent of `validate-atmost-once-period-in-seconds`.

One trade-off needs your call. The deferral means `ConnectionPool.resourceErrorOccurred` is not reached while the resource is enlisted, so `doFailAllConnectionsProcessing()` does not run for a pool with `fail-all-connections=true` until the transaction completed. `notifyWaitingThreads()` is not needed on that path, nothing was freed - the resource is still held by the transaction. The missing flush is a real difference for the other caller that reaches this listener while enlisted: `ManagedConnectionImpl.transactionCompleted()` fires `connectionErrorOccurred` when `markedForRemoval` is set. Options:

1. Leave it as is, documented in the code. This is the current state of the PR.
2. Move the enlisted check into `ConnectionPool.resourceErrorOccurred`, after the `failAllConnections` block. That keeps the flush, but it then also applies to the XA listener (`ConnectorAllocator.ConnectionListenerImpl`), which removes itself from the `ManagedConnection` unconditionally, so that listener would have to keep itself attached while enlisted too. Otherwise a handle the application has not closed yet can no longer find its way back to the pool.
3. A separate pool entry point for the enlisted case.

I did not take 2 or 3 because both widen the change beyond local-tx, which is the opposite of what you asked for. Both are small to implement if you prefer one of them.

Separately: `isConnectionError` is unit tested, but nothing asserts that `begin()`/`commit()`/`rollback()` actually call `connectionErrorOccurred` before throwing, because jdbc-core has no mocking framework on its test classpath. Tell me if that coverage is worth adding EasyMock to the module.

Local verification with Maven 3.9.16 / JDK 21 after the rebase:
- `appserver/jdbc/jdbc-ra/jdbc-core` - BUILD SUCCESS, 39 tests, 0 failures.
- `appserver/connectors/connectors-runtime` - BUILD SUCCESS, 125 tests, 0 failures, 1 pre-existing skip.

OndroMih:
This is superceeded by https://github.com/eclipse-ee4j/glassfish/pull/26193, which contains a functional fix proven by integration tests.

@renatsaf , can you review with Claude if any improvements in this PR still make sense on top of https://github.com/eclipse-ee4j/glassfish/pull/26193?

renatsaf:
@OndroMih thanks — I went through #26193 and compared it against this PR. Short answer: your root cause is the deeper one and mine did not cover it, but there is one piece here that is still a real functional delta on top of #26193, and I now have driver measurements that make the case concrete.

## What #26193 fixes that this PR missed

Moving `transactionStarted()` after a successful `setAutoCommit(false)` is the actual root cause. With the old order, a failed `begin()` left the managed connection with `transactionInProgress == true` forever, and `ManagedConnectionFactoryImpl.isValid()` returns immediately when `isTransactionInProgress()` — so that connection was permanently exempt from validation at *any* period, including `0`. That explains the "forever" in #25930 far better than my analysis did. Same for the `makeSureResourceIsBusy(resourceFromPool)` fix and the stale-MCF subject fix — both are independent bugs this PR never touched.

## What still isn't covered

The assertion in your new IT for `validate-atmost-once-period > 0`:

```java
assertThat("some early requests must fail (stale connections within validation window)",
           succeeded, lessThan(requestResults.size()));
```

This codifies as expected behaviour that requests keep failing for the whole validation window. #25930 goes from "broken forever" to "broken for the length of the window" — with `validate-atmost-once-period-in-seconds=60` or `300` in production that is still a stream of 500s, while the server already knows the connection is dead. The `connectionValidation=false` branch is observational only, so a pool without validation never self-heals at all.

That is what change **(1)** of this PR closes, and it composes with #26193 rather than competing with it: when `begin()`/`commit()`/`rollback()` fail with an 08-class `SQLState`, `SQLRecoverableException` or `SQLNonTransientConnectionException`, the RA raises `CONNECTION_ERROR_OCCURRED`. `ConnectionPool.java:806` checks `hasConnectionErrorOccurred()` *before* `isConnectionValid()`, so the handle is discarded on the next checkout regardless of the period, and regardless of whether validation is configured at all — one failed request instead of a whole window. The two edits to `begin()` are in different places (yours before the `try`, mine in the `catch`), and GitHub still reports this branch as mergeable.

## Driver measurements (Jaybird 5.0.10 / Firebird 5.0)

Since the IT only covers PostgreSQL, I measured what the driver in my production stack actually reports. Firebird 5.0.0.1306, Jaybird 5.0.10.java11, connection broken two ways — server-side (`DELETE FROM MON$ATTACHMENTS`) and by a real blocked socket read (`soTimeout`, the same shape as your `pg_sleep` outage):

| operation | thrown | SQLState |
|---|---|---|
| `setAutoCommit(false)` (== `begin()`) | **nothing** | — |
| `commit()` on killed attachment | `java.sql.SQLException` | `08003` "connection shutdown / Killed by database administrator" |
| `rollback()` on killed attachment | `java.sql.SQLException` | `08003` |
| statement on killed attachment | `java.sql.SQLException` | `08003` |
| statement blocked past `soTimeout` | `java.sql.SQLException` &larr; `java.net.SocketTimeoutException` | `08006` "Error reading data from the connection" |
| `commit()` after that timeout | `FBSQLException` | `08003` "You cannot commit a closed connection" |

Two consequences:

* **Jaybird's `setAutoCommit(false)` never round-trips** — it sets a local flag and the transaction starts lazily — so `LocalTransactionImpl.begin()` cannot throw on Firebird at all, and the stuck-`transactionInProgress` state your fix removes never arises there. On Firebird the only local-transaction detection point is `commit()`/`rollback()`, which is precisely the path change (1) covers and #26193 does not. Oracle in #25930 fails at `begin()`; Firebird fails at `commit()`. Both need covering.
* **None of these are `SQLRecoverableException` or `SQLNonTransientConnectionException`** — they are plain `java.sql.SQLException`. Only the `SQLState` class-`08` branch of the detection catches them, so that branch is not redundancy, it is the sole working check for this driver.

Happy to add a Firebird-flavoured variant of `JdbcPoolValidationIT` if you want a second driver covered — `DELETE FROM MON$ATTACHMENTS WHERE MON$ATTACHMENT_ID = ...` breaks a pooled connection cleanly without touching the container, which is a bit less timing-dependent than the `pg_sleep`/`socketTimeout` race.

## What I'd drop

Change **(2)**, the deferred pool removal in `LocalTxConnectionEventListener` while the resource is still enlisted. @dmatej's warning applies squarely to it, it carries the unresolved `fail-all-connections` trade-off I described in https://github.com/eclipse-ee4j/glassfish/pull/26063#issuecomment-5226112635, and the double-processing it guards against is so far only demonstrated by a strict mock, not by observed behaviour. Your Testcontainers harness makes it testable for real — I'd rather prove it with an IT in a separate PR than carry it here on speculation. (Side note: `ConnectionPool:806` already removes the resource bypassing `failAllConnections` — there's a `// Possible failAllConnections bug` comment right there — so that gap isn't new, but change (2) does widen it.)

## One more gap, larger than either PR

Your IT comment notes that "the error is not propagated as a connection error" when a pooled connection dies mid-statement. That is true beyond local transactions: `ConnectionHolder` never calls `connectionErrorOccurred` for a `SQLException` out of statement execution, so local-tx begin/commit/rollback is only one detection point. My measurements above show the same on Firebird — the `08006` from the interrupted read is swallowed at the RA level. That looks like a genuinely separate (and much larger) issue — worth filing, not worth folding into either PR.

Tell me which you prefer: narrow this PR to change (1) and rebase it on #26193, or close it and open a small focused PR once #26193 is merged. The second is closer to the iterative approach @dmatej asked for and I'm fine with either.

dmatej:
Try to run locally:
```
mvn clean install -T4C -Pfastest
./runtests.sh jdbc_group3
./runtests.sh jdbc_group4
```
Tests in these groups failed.

>
[Tests](https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/job/PR-26063/21/testReport/) (8 failures / +8)
[Test / Ant-Light / jdbc_group3 / jdbc_group3..Mark-Connection-As-Bad [Local-XA -Shareable-Shareable - Read-Read] : ID](https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/job/PR-26063/21/testReport/junit/jdbc_group3/(empty)/Test___Ant_Light___jdbc_group3___Mark_Connection_As_Bad__Local_XA__Shareable_Shareable___Read_Read____ID/)
[Test / Ant-Light / jdbc_group3 / jdbc_group3..Mark-Connection-As-Bad [Local-XA -Shareable-Shareable - Read-Write] : ID](https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/job/PR-26063/21/testReport/junit/jdbc_group3/(empty)/Test___Ant_Light___jdbc_group3___Mark_Connection_As_Bad__Local_XA__Shareable_Shareable___Read_Write____ID/)
[Test / Ant-Light / jdbc_group3 / jdbc_group3..Mark-Connection-As-Bad [Local-XA -Shareable-Shareable - Write-Read] : ID](https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/job/PR-26063/21/testReport/junit/jdbc_group3/(empty)/Test___Ant_Light___jdbc_group3___Mark_Connection_As_Bad__Local_XA__Shareable_Shareable___Write_Read____ID/)
[Test / Ant-Light / jdbc_group3 / jdbc_group3..Mark-Connection-As-Bad [Local-XA -Shareable-Shareable - Write-Write] : ID](https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/job/PR-26063/21/testReport/junit/jdbc_group3/(empty)/Test___Ant_Light___jdbc_group3___Mark_Connection_As_Bad__Local_XA__Shareable_Shareable___Write_Write____ID/)
[Test / Ant-Light / jdbc_group4 / jdbc_group4..DMMCF Mark-Connection-As-Bad [Local-XA -Shareable-Shareable - Read-Read] : ID](https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/job/PR-26063/21/testReport/junit/jdbc_group4/(empty)/Test___Ant_Light___jdbc_group4___DMMCF_Mark_Connection_As_Bad__Local_XA__Shareable_Shareable___Read_Read____ID/)
[Test / Ant-Light / jdbc_group4 / jdbc_group4..DMMCF Mark-Connection-As-Bad [Local-XA -Shareable-Shareable - Read-Write] : ID](https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/job/PR-26063/21/testReport/junit/jdbc_group4/(empty)/Test___Ant_Light___jdbc_group4___DMMCF_Mark_Connection_As_Bad__Local_XA__Shareable_Shareable___Read_Write____ID/)
[Test / Ant-Light / jdbc_group4 / jdbc_group4..DMMCF Mark-Connection-As-Bad [Local-XA -Shareable-Shareable - Write-Read] : ID](https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/job/PR-26063/21/testReport/junit/jdbc_group4/(empty)/Test___Ant_Light___jdbc_group4___DMMCF_Mark_Connection_As_Bad__Local_XA__Shareable_Shareable___Write_Read____ID/)
[Test / Ant-Light / jdbc_group4 / jdbc_group4..DMMCF Mark-Connection-As-Bad [Local-XA -Shareable-Shareable - Write-Write] : ID](https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/job/PR-26063/21/testReport/junit/jdbc_group4/(empty)/Test___Ant_Light___jdbc_group4___DMMCF_Mark_Connection_As_Bad__Local_XA__Shareable_Shareable___Write_Write____ID/)

I have restarted the build once again to see if same tests fail again. However if they failed once I believe something is wrong. However ... it might be the test too.

renatsaf:
@dmatej thanks. Build #21 has 8 failures, all in the same scenario:

- jdbc_group3: `Mark-Connection-As-Bad [Local-XA -Shareable-Shareable - Write-Write | Read-Write | Read-Read | Write-Read]`
- jdbc_group4: `DMMCF Mark-Connection-As-Bad [Local-XA -Shareable-Shareable - ...]`, same four variants

Cause: the deferred removal in `LocalTxConnectionEventListener`. In `test9`-`test12` a `jdbc-local-pool` connection is marked bad inside a container transaction together with an XA connection. At transaction completion `ManagedConnectionImpl.transactionCompleted()` fires `connectionErrorOccurred` for the marked connection while the resource is still enlisted. Before this PR the listener removed it immediately; with the PR it only sets the flag and the connection is destroyed at the next checkout. The client reads `NumConnDestroyed` of `jdbc-local-pool` right after each call and expects 1..4, so all four local-pool assertions fail.

Your original doubt about notifying before throwing the `LocalTransactionException` was also correct. My statement in the August comment that a failed `begin()` is not processed twice is wrong: `connectionErrorOccurred` removes the resource through `PoolManager.resourceErrorOccurred`, and then the catch block in `PoolManagerImpl.getResource` calls `putbackDirectToPool`, which runs `resourceClosed` on the already removed handle. It does not show up in these devtests, but the double processing is there.

The deferral cannot be dropped on its own either: without it, notifying from `commit()`/`rollback()` removes the resource while it is still enlisted, which is the double processing the deferral was added for.

Since #26193 fixes #25930, I'm closing this PR. Detecting a broken connection in `commit()`/`rollback()` (the only detection point for drivers that start the transaction lazily, e.g. Jaybird) would need a different design and devtest coverage. That would be a separate issue, if it is wanted at all.

Issue #25607, opened 2025-07-17
EmbeddedVirtualServerHostNameTest doesn't dispose GlassFish runtime, leaving temporary directories not deleted


OndroMih:
### GlassFish Version (and build number)

7.1.0-SNAPSHOT

### JDK version

Any

### OS

Any

### Problem Description

The test [EmbeddedVirtualServerHostNameTest](https://github.com/eclipse-ee4j/glassfish/blob/e0ccdaaee7f7c2d03188fca0e9340da68e43bb01/appserver/tests/embedded/web/web-api/src/test/java/org/glassfish/tests/embedded/web/EmbeddedVirtualServerHostNameTest.java#L64) doesn't call glassfish.dispose() method in the cleanup phase. This leaves a temporary domain directory after running the Embedded GlassFish test suite.

Cause of the issue:
GlassFish runtime is stored in a local variable instead of storing it into the instance variable. Since the instance variable is null, dispose is not called later in the `@AfterAll` method.

### Steps to reproduce

Build GlassFish (e.g. with `mvn -Pfastest -T4C -Dcyclonedx.skip=true`) and then execute the `EmbeddedVirtualServerHostNameTest` test.

### Impact of Issue

The temporary directory is polluted after the test suite runs.

HamzaElzarw-2022:
Hello @OndroMih , can you assign me this issue please?

dmatej:
Duplicates #25545

OndroMih:
This is not a duplicate of #25545, @dmatej . This is about a simple fix in the test to call the dispose() method, nothing else. #25545 is about improving Embedded GlassFish to delete temporary directories on JVM shutdown even if dispose() method is not called.

PR #26069, opened 2026-06-08
Delete embedded GlassFish temporary directories via a JVM shutdown hook and restrict them to the current user

### LINKED ISSUE #25545: Embedded GlassFish doesn't remove temporary directories

7.1.0-SNAPSHOT

### JDK version

any

### OS

any

### Database

_No response_

### Problem Description

Following directories should be removed using shutdown hooks, including all content.
 
* gfembed
* fileinstall

They must not be also publicly visible - just for the current user.

### Steps to reproduce

Run tests with embedded GlassFish.

### Impact of Issue

_No response_

renatsaf:
Fixes #25545

## Problem

Embedded GlassFish creates a temporary instance directory (`gfembed*`) that:
- was created with default, world-readable permissions, and
- was only deleted from `dispose()`. If a caller invoked `start()`/`stop()` without `dispose()` (e.g. `EmbeddedVirtualServerHostNameTest`, or any program that forgets to dispose), the directory leaked.

## Changes

- **Owner-only temp directory** — `EmbeddedGlassFishRuntime` now creates the temp instance directory accessible only by the current user: atomic `rwx------` on POSIX file systems, best-effort owner-only restriction on non-POSIX (Windows). This replaces the previous `createTempFile` + `delete` + `mkdir` dance.
- **Shutdown-hook safety net** — `AutoDisposableGlassFish` registers a JVM shutdown hook in `start()` that disposes the instance (so the server stops cleanly and releases its files) and deletes the temporary directory if `dispose()` was never called. The hook is removed on a normal `dispose()`, and registered only once across `start()`/`stop()`/`start()` cycles. Deletion remains in `dispose()` too, so an instance can still be recycled without a JVM restart.
- **Test** — added a forked-JVM test (`LifeCycleTest.shutdownHookDeletesTempDirWhenNotDisposed` + `StartStopNoDisposeRunner`) that starts and stops an embedded instance without calling `dispose()` and verifies the temp directory is removed when the JVM exits. The fork is required because shutdown hooks only run at actual JVM termination.

## Notes

- The temporary directory name changes from `gfembed<n>tmp` to `gfembed<n>`; the `gfembed` prefix is preserved.
- The Windows file-lock scenario mentioned in the issue (Derby `ejbtimer` files held by another process during deletion) is still handled best-effort — `deleteRecursive` logs and continues. Fully hardening Windows deletion (retry / `deleteOnExit`) can be a follow-up; see also #25607 for test-side cleanup.

## Testing

`mvn -pl appserver/tests/embedded/basic/lifecycle test` — both the existing `LifeCycleTest.test` and the new shutdown-hook test pass.


appserver/tests/embedded/basic/lifecycle/src/test/java/org/glassfish/tests/embedded/basic/lifecycle/LifeCycleTest.java
OndroMih: Update year to 2026

appserver/tests/embedded/basic/lifecycle/src/test/java/org/glassfish/tests/embedded/basic/lifecycle/StartStopNoDisposeRunner.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation.
```
OndroMih: Change year to 2026

nucleus/core/bootstrap-osgi/src/main/java/org/glassfish/main/boot/embedded/AutoDisposableGlassFish.java
OndroMih: Update year to 2026

nucleus/core/bootstrap-osgi/src/main/java/org/glassfish/main/boot/embedded/EmbeddedGlassFishRuntime.java
OndroMih: Update year to 2026

OndroMih (commented):
Looks good. Please update copyright years as suggested in comments.

renatsaf:
@OndroMih Thank you for the review! I've added 2026 to the copyright headers of all four files in ba23d89796.

One note: in the new `StartStopNoDisposeRunner.java` I appended the year (`2025, 2026`) instead of replacing it, to keep the existing year intact. Let me know if you'd prefer plain `2026` there.

OndroMih:
That file is completely new so copyright should be only for 2026

renatsaf:
Done in 2a7d0a59f1 - the new file now has only 2026. Thank you!

OndroMih (approved):

PR #25446, opened 2025-04-09
Add Core Profile 11 Runner for Glassfish Embedded Web

Thihup:

PR #26070, opened 2026-06-08
Propagate deployment failures and fix CDI/assembly issues in GlassFish Embedded

### LINKED ISSUE #26062: GlassFish Embedded: deployment failures are silently swallowed and CDI persistence extension NPEs without an active invocation
@
## Summary

While running the Jakarta EE TCKs against GlassFish Embedded (see #25446), two embedded-runtime bugs surface in the product code (independent of the TCK runners):

### 1. Embedded `deploy(...)` swallows deployment failures
`com.sun.enterprise.admin.cli.embeddable.DeployerImpl#deploy(File, String...)` runs the `deploy` command and then returns `actionReport.getResultType(String.class)` **without checking `actionReport.hasFailures()`**. A failed deployment therefore returns normally instead of raising `GlassFishException`, so callers cannot detect the failure. All other `deploy(...)` overloads delegate to this method, so they inherit the problem.

**Expected:** a failed deployment throws `GlassFishException` carrying the action-report message and failure cause.

### 2. `PersistenceExtension` NPEs when there is no habitat / no current invocation
`org.glassfish.cdi.persistence.PersistenceExtension#afterBean(...)` chains:

```java
Globals.get(InvocationManager.class).getCurrentInvocation().getContainer();
```

In embedded scenarios this is invoked when there is no default habitat or no current invocation, producing an NPE during `AfterBeanDiscovery`.

**Expected:** the extension returns early (adds no persistence beans) when there is no habitat or no active invocation, instead of throwing.

### 3. (Minor) stray `beans.xml` in the Embedded Web assembly
The `glassfish-embedded-web` assembly bundles `META-INF/beans.xml`, which alters CDI bean-discovery behavior of the assembled artifact. It should be excluded from the uber-jar.

## Environment
- GlassFish Embedded (Web)

## Notes
These fixes already exist as isolated commits in PR #25446 and can be applied independently of the (still-under-discussion) TCK runner files.
@

renatsaf:
Extracts the three self-contained product fixes from #25446 (GlassFish Embedded), independent of the TCK runner files still under discussion there. Original author: @Thihup.

## Fixes

- **Embedded deploy failures are now propagated.** `DeployerImpl.deploy(File, ...)` returned normally even when the action report contained failures, so callers could not detect a failed deployment. It now throws `GlassFishException` carrying the report message and failure cause.
- **`PersistenceExtension` no longer NPEs** when there is no default habitat or no active invocation (e.g. GlassFish Embedded). `afterBean()` returns early in those cases instead of dereferencing the current invocation.
- **`META-INF/beans.xml` is excluded** from the `glassfish-embedded-web` assembly (both unpacked dependency sets) so a bundled descriptor doesn't alter CDI bean discovery of the uber jar.

Fixes #26062

### DIFF
--- a/appserver/extras/embedded/web/src/assembly/package.xml
+++ b/appserver/extras/embedded/web/src/assembly/package.xml
@@ -31,6 +31,7 @@
                     <exclude>META-INF/MANIFEST.MF</exclude>
                     <exclude>META-INF/hk2-locator/*</exclude>
                     <exclude>META-INF/loggerinfo/*</exclude>
+                    <exclude>META-INF/beans.xml</exclude>
                     <exclude>module-info.class</exclude>
                     <exclude>META-INF/versions/*/module-info.class</exclude>
                     <exclude>*.jar</exclude>
@@ -58,6 +59,7 @@
                     <exclude>META-INF/MANIFEST.MF</exclude>
                     <exclude>META-INF/hk2-locator/*</exclude>
                     <exclude>META-INF/loggerinfo/*</exclude>
+                    <exclude>META-INF/beans.xml</exclude>
                     <exclude>module-info.class</exclude>
                     <exclude>META-INF/versions/*/module-info.class</exclude>
                     <exclude>*.jar</exclude>
--- a/appserver/web/weld-integration/src/main/java/org/glassfish/cdi/persistence/PersistenceExtension.java
+++ b/appserver/web/weld-integration/src/main/java/org/glassfish/cdi/persistence/PersistenceExtension.java
@@ -51,9 +51,17 @@ public class PersistenceExtension implements Extension  {
     private PersistenceUnitDescriptor firstPersistenceUnitDescriptor;
 
     public void afterBean(final @Observes AfterBeanDiscovery afterBeanDiscovery, BeanManager beanManager) {
-        var container = Globals.get(InvocationManager.class)
-                                  .getCurrentInvocation()
-                                  .getContainer();
+        if (Globals.getDefaultHabitat() == null) {
+            return;
+        }
+
+        var currentInvocation = Globals.get(InvocationManager.class).getCurrentInvocation();
+
+        if (currentInvocation == null) {
+            return;
+        }
+
+        var container = currentInvocation.getContainer();
 
         if (container instanceof ApplicationInfo applicationInfo) {
 
--- a/appserver/web/weld-integration/src/main/java/org/glassfish/cdi/persistence/PersistenceExtension.java
+++ b/appserver/web/weld-integration/src/main/java/org/glassfish/cdi/persistence/PersistenceExtension.java
@@ -40,6 +40,7 @@ import jakarta.persistence.metamodel.Metamodel;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
 
+import org.glassfish.api.invocation.ComponentInvocation;
 import org.glassfish.api.invocation.InvocationManager;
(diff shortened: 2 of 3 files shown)


appserver/web/weld-integration/src/main/java/org/glassfish/cdi/persistence/PersistenceExtension.java
```diff
@@ -51,9 +51,17 @@ public class PersistenceExtension implements Extension  {
...
 
     public void afterBean(final @Observes AfterBeanDiscovery afterBeanDiscovery, BeanManager beanManager) {
-        var container = Globals.get(InvocationManager.class)
-                                  .getCurrentInvocation()
-                                  .getContainer();
+        if (Globals.getDefaultHabitat() == null) {
+            return;
+        }
+
+        var currentInvocation = Globals.get(InvocationManager.class).getCurrentInvocation();
```
dmatej: The problem with var is that when we refactor something, methods change, interfaces, class hierarchy, and suddenly compiler shows that method doesn't exist, but I have no idea what class was declaring that before.

On the other hand when we go just for the `instanceof` check, there it is probably even better than using `Object`.

appserver/extras/embedded/web/src/assembly/package.xml
```diff
@@ -58,6 +59,7 @@
                     <exclude>META-INF/MANIFEST.MF</exclude>
                     <exclude>META-INF/hk2-locator/*</exclude>
                     <exclude>META-INF/loggerinfo/*</exclude>
+                    <exclude>META-INF/beans.xml</exclude>
```
dmatej, line 62: Just for the record, could you check which dependency contained the `beans.xml` file?

dmatej (approved):

appserver/extras/embedded/web/src/assembly/package.xml
renatsaf: Traced it: the `META-INF/beans.xml` comes from **`org.glassfish.jersey.ext.cdi:jersey-cdi1x-servlet`**.

It ships an empty bean-discovery marker:

```xml
<beans/>
```

(just that element under an old Oracle 2014–2018 license header). I confirmed it by unpacking the assembled `glassfish-embedded-web.jar` and diffing — its `META-INF/beans.xml` is byte-identical to the one inside `jersey-cdi1x-servlet`. Of the CDI-related libraries bundled in the uber jar, `jersey-cdi1x-servlet` is the only one carrying a `beans.xml` (the Weld jars don't), so it's the sole source.

renatsaf:
Thanks for the review, @dmatej! Addressed both points:

**The `var` on the container check** — pushed a follow-up commit (`21763b8`). I gave `currentInvocation` an explicit `ComponentInvocation` type and inlined `getContainer()` directly into the `instanceof`, dropping the throwaway local:

```java
ComponentInvocation currentInvocation = Globals.get(InvocationManager.class).getCurrentInvocation();
if (currentInvocation == null) {
    return;
}
if (currentInvocation.getContainer() instanceof ApplicationInfo applicationInfo) {
    ...
}
```

`getContainer()` returns `Object`, so this avoids both the `var` and an `Object`-typed local — which lines up with your point about preferring the plain `instanceof`.

**The `beans.xml` source** — it's `org.glassfish.jersey.ext.cdi:jersey-cdi1x-servlet`; details in the inline reply above.

OndroMih (approved):

dmatej (approved):

PR #26073, opened 2026-06-09
Fix publishing the glassfish.org website

OndroMih:
The artifact org.glassfish.docs:distribution:5.1.0 depends on nucleus parent SNAPSHOT version. After the SNAPSHOT version was deleted, the dependency plugin fails to resolved it. We download the artifact manually from Maven Central instead of resolving it.

### DIFF
--- a/docs/publish/pom.xml
+++ b/docs/publish/pom.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
+    Copyright (c) 2026 Contributors to the Eclipse Foundation.
     Copyright (c) 2017, 2021 Oracle and/or its affiliates. All rights reserved.
 
     This program and the accompanying materials are made available under the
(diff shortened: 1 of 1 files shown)


dmatej:
I am against reintroduction of Ant again, there must be some easier way. Isn't it possible to configure the dependency plugin to avoid the issue? Or use exclusions?

OndroMih:
> Isn't it possible to configure the dependency plugin to avoid the issue? Or use exclusions?

No, there's no way around it with the dependency plugin. I think the best option is to introduce our own Maven plugin or add a new goal to the existing buildhelper plugin.

dmatej:
> > Isn't it possible to configure the dependency plugin to avoid the issue? Or use exclusions?
> 
> No, there's no way around it with the dependency plugin. I think the best option is to introduce our own Maven plugin or add a new goal to the existing buildhelper plugin.

Unfortunately they broke it in 3.11.0, with 3.10.0 it worked. I will report that.

dmatej:
Reported here: https://github.com/apache/maven-dependency-plugin/issues/1642

OndroMih:
Closing in favor of https://github.com/eclipse-ee4j/glassfish/pull/26074

PR #26088, opened 2026-06-13
Correct links for the 8.0.3 platform TCK results

arjantijms:
Per feedback from https://github.com/jakartaee/platform/issues/1308#issuecomment-4680940061


dmatej (approved):
Maybe you could create a script which would list all links and checked that there's always http 200 for all of them.

OndroMih (approved):

OndroMih:
@dmatej , I’m pretty sure that both the old and new links work and give 200. The changes are because the certification process requested using more “correct” links, which are more durable.

arjantijms:
> Maybe you could create a script which would list all links and checked that there's always http 200 for all of them.

That's already done. The links are all "live" links; meaning the TCK runner job actually download the TCK Big ZIps from those URLs and calculate the SHA from the downloaded zip from those.

The CCR validation again checks if the links are valid and indeed have a 200 status.

PR #26090, opened 2026-06-13
Remove thread-context-classloader hack in Weld BDA; rely on per-BDA ResourceLoader SPI

### LINKED ISSUE #25593: Instruct Weld to use a specific classloader instead of hacky solution with thread context classloader

7.0.25

### JDK version

N/A

### OS

N/A

### Database

_No response_

### Problem Description

Because of https://issues.jboss.org/browse/WELD-781, a hacky solution was applied a long time ago (see https://github.com/eclipse-ee4j/glassfish/issues/18152) to load CDI classes using a specific classloader. So far, no issues with it. However, the Weld issue has been fixed and GlassFish can probably specify the classloader, as suggested by this comment in GlassFish, in https://github.com/eclipse-ee4j/glassfish/blob/bf590b8a5da1bcdf800cd243180775c52080be3c/appserver/web/weld-integration/src/main/java/org/glassfish/weld/BeanDeploymentArchiveImpl.java#L210:

```
// The TCL is unset at the end of deployment of CDI beans in WeldDeployer.event
            // XXX: This is a workaround for issue https://issues.jboss.org/browse/WELD-781.
            // Remove this as soon as the SPI comes in.
            Thread.currentThread().setContextClassLoader(this.moduleClassLoaderForBDA);
```

### Steps to reproduce

Currently no issues.

I hit issues when I modified GlassFish code to define a bean deployment archive with a classloader of an OSGi module (for a specific module). Then the thread context classloader was set to this classloader and remained set during the whole Weld deployment, because the method that sets it (`getBeanClasses()`), is called also in a place where GlassFish doesn't expect that the thread context classloader is modified and doesn't reset it afterwards.

### Impact of Issue

Currently no impact. Only cleanup.

renatsaf:
## Summary

Fixes #25593.

`BeanDeploymentArchiveImpl.getBeanClasses()` set the thread context classloader (TCL) to the BDA's module classloader as a side effect — a long-standing workaround for [WELD-781](https://issues.jboss.org/browse/WELD-781), introduced back in #18152, from before Weld offered an SPI to control the classloader per bean deployment archive.

That side effect is fragile: a *getter* mutates global thread state, so it leaks into any caller (e.g. `toString()`), and when a BDA uses a custom classloader the TCL stays set for the rest of the Weld deployment because GlassFish only restores it at the end of `WeldDeployer.enable()`.

The SPI has long since landed and GlassFish already uses it:
- **per-BDA `ResourceLoader`** — registered in `WeldDeployer.enable()` with `getModuleClassLoaderForBDA()`, so Weld loads each archive's classes/resources with the correct classloader;
- **`ProxyServices`** — resolves the bean's own classloader for proxy generation.

The TCL manipulation is therefore redundant.

## Changes

- `BeanDeploymentArchiveImpl.getBeanClasses()` is now a pure getter (no TCL side effect).
- Removed the dependent machinery: the `deploymentComplete` flag and its `is/set` accessors, the `originalClassLoader` save/restore and `deploymentComplete()` call in `WeldDeployer.enable()`, and the now-orphaned `AS-CDI-00001` *"Setting Context Class Loader"* log message.

## Notes / risk

Behavioral change: during Weld bootstrap the TCL is no longer mutated to each BDA's module classloader; it stays as the deployment-time classloader. `ProxyServicesImpl` still falls back to the TCL only for *non-application* beans (e.g. `UserTransaction` proxies), which now resolve against the application classloader — a reasonable choice. As the issue notes there is no functional reproducer today, so correctness rests on the CDI/Weld TCKs (multi-module EARs, RARs, WEB-INF/lib bean archives, built-in-bean proxies).

### DIFF
--- a/appserver/web/weld-integration/src/main/java/org/glassfish/cdi/CDILoggerInfo.java
+++ b/appserver/web/weld-integration/src/main/java/org/glassfish/cdi/CDILoggerInfo.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2022 Contributors to the Eclipse Foundation.
+ * Copyright (c) 2022, 2026 Contributors to the Eclipse Foundation.
  * Copyright (c) 2013, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -39,9 +39,6 @@ public class CDILoggerInfo {
     public static Logger getLogger() {
         return CDI_LOGGER;
     }
-
-    @LogMessageInfo(message = "Setting Context Class Loader for {0} to {1}.", level = "FINER")
-    public static final String SETTING_CONTEXT_CLASS_LOADER = CDI_LOGMSG_PREFIX + "-00001";
 
     @LogMessageInfo(message = "BeanDeploymentArchiveImpl::addBeanClass - adding {0} to {1}.", level = "FINE")
     public static final String ADD_BEAN_CLASS = CDI_LOGMSG_PREFIX + "-00002";
--- a/appserver/web/weld-integration/src/main/java/org/glassfish/weld/BeanDeploymentArchiveImpl.java
+++ b/appserver/web/weld-integration/src/main/java/org/glassfish/weld/BeanDeploymentArchiveImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, 2025 Contributors to the Eclipse Foundation.
+ * Copyright (c) 2021, 2025, 2026 Contributors to the Eclipse Foundation.
  * Copyright (c) 2009, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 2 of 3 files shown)


dmatej (approved):

dmatej:
Rebased, I hope now Jenkins will like it :)

dmatej:
Weird failure in GHA Ubuntu ... I will retry, so just for the record. 
```
[ERROR] Failures: 
66914
[ERROR] org.glassfish.main.test.app.monitoring.ThreadPoolMonitoringTest.testThreadPoolSizeDecrease
66915
[INFO]   Run 1: PASS
66916
[ERROR]   Run 2: ThreadPoolMonitoringTest.testThreadPoolSizeDecrease:466 afterDecrease (2 failures)
66917
	java.lang.AssertionError: current threads
66918
Expected: <0>
66919
     but: was <22>
66920
	java.lang.AssertionError: busy threads
66921
Expected: <0>
66922
     but: was <10>
66923
[INFO] 
66924
[ERROR] Errors: 
66925
[ERROR]   ManagedExecutorDefinitionWebTest.testCopyCompletableFutureTwice:72 » Socket Unexpected end of file from server
66926
[ERROR]   ManagedExecutorTasksSubmittedDuringDeploymentTest.testSubmittedTaskExecuted:70 » Socket Unexpected end of file from server
66927
[INFO] 
66928
[ERROR] Tests run: 100, Failures: 1, Errors: 2, Skipped: 3
```

Issue #24808, opened 2024-02-13
Class Loader Delegate false not working


kurtomerfaruk:
[//]: # "Please remove these comments"
[//]: # "Provide a general summary of the issue in the Title above"

## Environment Details
* GlassFish Version (and build number): 7.0.9-7.0.12
* JDK version: JDK 11
* OS: Windows 11
* Database: Microsoft Sql Server

----------

## Problem Description
Although I use <class-loader delegate="false"/> in glassfish-web.xml, Glassfish uses the jars on the server.

```
             <dependency>
                <groupId>org.glassfish</groupId>
                <artifactId>jakarta.faces</artifactId>
                <version>4.0.4</version>
            </dependency>
```
`[2024-02-13T15:55:38.537894+03:00] [GF 7.0.9] [INFO] [faces.config.listener.version] [jakarta.enterprise.resource.webcontainer.faces.config] [tid: _ThreadID=57 _ThreadName=admin-listener(2)] [levelValue: 800] [[
  Initializing Mojarra 4.0.5 for context `

I also have the same problem with other jars.

OndroMih:
Hi, which other jars you have a problem with?

In general, it's not possible to just add a library into your application and expect that GlassFish will use it. GlassFish will usually use the components inside GlassFish - for example Mojarra will be used when GlassFish detects that your application uses JSF and it needs to start a JSF engine. When you disable classLoader delegation, what changes is where classes are loaded from if they are imported into other classes in the application. 

However, specifically for Jakarta Faces, there's an option supported in glassfish-web.xml file. You can specify the property `useBundledJsf` to `true` if you want to use a Jakarta Faces implementation packaged in your application:

```
<glassfish-web-app error-url="">
  <class-loader delegate="false"/>
  <property name="useBundledJsf" value="true"/>
</glassfish-web-app>
```
To explain how classloader delegation works for other components, I'll give you an example with Jackson JSON mapper, which is present in GlassFish and is also often bundled in applications. By default, you don't need to add Jackson JARs to your application and you can still access the Jackson's ObjectMapper class directly, e.g. from a servlet. If you add Jackson JARs to your application, the application will still use the Jackson library in GlassFish. WIth delegate=false, your application will use Jackson library in your app.

There is an exception: If a class's package starts with `jakarta`, it's always loaded from GlassFish. Even if delegate=false and you have that class somewhere in your application.

[...]

PR #26091, opened 2026-06-13
Document class loader delegation exceptions and the useBundledJsf property

renatsaf:
## Problem

As reported in #24808, setting `<class-loader delegate="false"/>` in `glassfish-web.xml` does not make GlassFish use a Jakarta Faces implementation bundled in the application — the server's Mojarra is used instead. This is by design: a fixed set of packages is always delegated to the server regardless of the `delegate` setting, including the Faces API (`jakarta.faces.*`) and implementation (`com.sun.faces.*`). The documented workaround (the `useBundledJsf` property) was not actually documented, and the always-delegated package set was described only as "`java.*` and `javax.*`".

## Changes (docs only)

- **Application Development Guide → Class Loaders → Delegation**: list the full set of always-delegated packages (matching `WebappClassLoader.DELEGATED_PACKAGES`) and explain the `useBundledJsf` workaround with an example.
- **Application Deployment Guide → `class-loader` descriptor reference**: document the previously-undocumented `useBundledJsf` property (and its `useMyFaces` synonym) in the properties table, and expand the `delegate` attribute description to mention the Faces exception.

No behavior change.

Refs #24808

### DIFF
--- a/docs/application-deployment-guide/src/main/asciidoc/dd-elements.adoc
+++ b/docs/application-deployment-guide/src/main/asciidoc/dd-elements.adoc
@@ -1218,11 +1218,15 @@ Servlet specification and looks in its class loader before looking in
 the parent class loader. It's safe to set this to `false` only for a web
 module that does not interact with any other modules.
 
-For a number of packages, including `java.*` and `javax.*`, symbol
-resolution is always delegated to the parent class loader regardless of
-the delegate setting. This prevents applications from overriding core
-Java runtime classes or changing the API versions of specifications that
-are part of the Jakarta EE platform.
+For a number of packages, including `java.*`, `javax.*`, `jakarta.*`,
+and the Jakarta Faces implementation (`com.sun.faces.*` and
+`jakarta.faces.*`), symbol resolution is always delegated to the parent
+class loader regardless of the delegate setting. This prevents
+applications from overriding core Java runtime classes or changing the
+API versions of specifications that are part of the Jakarta EE platform.
+To use a Jakarta Faces implementation bundled in the application instead
+of the one provided by the {productName}, set the `useBundledJsf`
+property described in xref:#gcfjs[Table C-24].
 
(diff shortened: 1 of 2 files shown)


docs/application-deployment-guide/src/main/asciidoc/dd-elements.adoc
```diff
@@ -1218,11 +1218,15 @@ Servlet specification and looks in its class loader before looking in
...
 module that does not interact with any other modules.
 
-For a number of packages, including `java.*` and `javax.*`, symbol
-resolution is always delegated to the parent class loader regardless of
-the delegate setting. This prevents applications from overriding core
-Java runtime classes or changing the API versions of specifications that
-are part of the Jakarta EE platform.
+For a number of packages, including `java.*`, `javax.*`, `jakarta.*`,
+and the Jakarta Faces implementation (`com.sun.faces.*` and
+`jakarta.faces.*`), symbol resolution is always delegated to the parent
```
OndroMih: The complete list is in https://github.com/eclipse-ee4j/glassfish/blob/87e5ef7215fe05236a8fe1729563dc84d6561978/appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java#L138-L147

renatsaf: Thanks — aligned both docs with `WebappClassLoader.DELEGATED_PACKAGES` exactly (`jakarta`, `javax`, `sun`, `org.xml.sax`, `org.w3c.dom`, `org.glassfish.wasp.taglibs.standard`, `com.sun.faces`, `org.eclipse.microprofile`). `java.*` was in the previous wording but isn't part of that set, so it's now described separately as delegated by the JVM bootstrap class loader. Fixed in 40302199b5.

OndroMih (approved):

PR #24570, opened 2023-09-03
Classes in ScatteredArchive are processed only once.

OndroMih:
They are not processed again when the parser is called for JARs (web fragments). Signed-off-by:Ondro Mihalyi 

Fixes #24415.

To do:

* write a test that covers #24415 with both ScatteredArchive and with a plain WAR file. The issue #24415 doesn't happen for a plain WAR file even without this fix. A reproducer needs to bundle at least 1 JAR file in the lib directory.
  * I will wait until https://github.com/eclipse-ee4j/glassfish/pull/24576 is merged and then add a test in the newly enabled GF embedded test suite
* Decide what to do for `WebFragmentDescriptor`. This solution returns no items for it. Maybe it should return all classes in the nested JAR file referenced from the descriptor.

PR #26100, opened 2026-06-16
Scan web-fragment nested JARs for annotations in embedded ScatteredArchive

### LINKED ISSUE #24592: GlassFish Embedded" WebFragmentDescriptor in ScatteredArchive should return all classes in the nested JAR file referenced from the descriptor
## Environment Details
* GlassFish Version (and build number): 7.0.8

----------

## Problem Description

When application is deployed as a ScatteredArchive to GlassFish Embedded, nested JAR files should be scanned for annotations. Currently, the [ScatteredWebArchivist.java](https://github.com/eclipse-ee4j/glassfish/pull/24570/files#diff-a862208ea5334ea07096ef60a5454174f94789987e2a9b57b4fe8c63be6d6a34) ignores nested JAR files and only scans `*.class` files in the archive but in in subarchives.

The issue is related to https://github.com/eclipse-ee4j/glassfish/pull/24570, which fixes related issues with ScatteredArchive but leaves this one out of scope.

## Impact of Issue

Currently no impact, it's just nice to add support for this usecase.

Type: bug
Component: GlassFish Embedded (ScatteredArchive)
Priority: low

renatsaf:
## Problem

When a web application is deployed to GlassFish Embedded as a `ScatteredArchive`, annotated components (servlets, filters, listeners, …) bundled inside a **nested JAR** (a web fragment in `WEB-INF/lib`) were silently ignored. Only `WEB-INF/classes/*.class` was scanned.

Root cause (two cooperating gaps):
- `EmbeddedWebScanner.process()` in `ScatteredWebArchivist` only kept entries under `WEB-INF/classes/`.
- `WebFragmentDescriptor.getArchiveFileEntries()` returned `Collections.emptyEnumeration()`, so when the scanner processed a fragment descriptor it got nothing.

This is the follow-up that was explicitly left out of scope in #24570.

## Fix

- `WebFragmentDescriptor.getArchiveFileEntries()` now opens the nested JAR referenced by its `jarName` (`WEB-INF/lib/<jarName>`) via `getSubArchive(...)`, eagerly materializes the entry names (so the sub-archive can be closed), and returns them.
- `EmbeddedWebScanner.process()` selects the class-name prefix by descriptor type: `WEB-INF/classes/` for the web module, JAR-relative (`""`) for a `WebFragmentDescriptor`; it also skips `module-info.class`.

## Test

Adds `ScatteredArchiveTest#testWebFragmentInNestedJar`, which builds a web-fragment JAR at runtime (an `@WebServlet` plus `META-INF/web-fragment.xml`), bundles it via `addClassPath` (→ `WEB-INF/lib`), deploys, and asserts the servlet responds. The servlet class is present **only** inside the nested JAR, so it is reachable only if the JAR is scanned.

Verified end-to-end against a rebuilt `glassfish-embedded-all` uber jar:
- With the fix: new test passes; full `ScatteredArchiveTest` (3 tests) green.
- Without the fix: the new test fails (servlet 404) — confirming it is a genuine regression guard.

Fixes #24592

### DIFF
--- a/appserver/web/web-glue/src/main/java/org/glassfish/web/embed/impl/ScatteredWebArchivist.java
+++ b/appserver/web/web-glue/src/main/java/org/glassfish/web/embed/impl/ScatteredWebArchivist.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2025 Contributors to the Eclipse Foundation.
+ * Copyright (c) 2022, 2025, 2026 Contributors to the Eclipse Foundation.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -37,6 +37,7 @@ import org.glassfish.internal.api.Globals;
 import org.glassfish.web.LogFacade;
 import org.glassfish.web.deployment.archivist.WebArchivist;
 import org.glassfish.web.deployment.descriptor.WebBundleDescriptorImpl;
+import org.glassfish.web.deployment.descriptor.WebFragmentDescriptor;
 import org.jvnet.hk2.annotations.Service;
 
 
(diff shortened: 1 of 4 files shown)


OndroMih (approved):

appserver/web/web-glue/src/main/java/org/glassfish/web/embed/impl/ScatteredWebArchivist.java
```diff
@@ -37,6 +37,7 @@
 import org.glassfish.web.LogFacade;
 import org.glassfish.web.deployment.archivist.WebArchivist;
 import org.glassfish.web.deployment.descriptor.WebBundleDescriptorImpl;
+import org.glassfish.web.deployment.descriptor.WebFragmentDescriptor;
 import org.jvnet.hk2.annotations.Service;
```
OndroMih:
```suggestion
import org.jvnet.hk2.annotations.Service;

```

dmatej:
@OndroMih You don't have to add empty commits, jobs can be reexecuted manually (if it is enabled in the workflow file). Unfortunately Microsoft is doing something with Maven Central reference, I have noticed in logs of another build that it tried to use Apache instead of Sonatype - and failed. Here the ubuntu job just froze for 3 hours. I will try to restart it again now ...

OndroMih:
I didn’t see any option to start the workflows in the UI. Typically, when they run, I can restart them. But here they didn’t even run and I didn’t see an option to run them.

PR #26106, opened 2026-06-16
Move JSFTemplate to GlassFish adminui

arjantijms:
JSFTemplate is exclusively used by the AdminUI in GlassFish, and therefor should be bundled with it as a private internal dependency.

### DIFF
--- a/nucleus/parent/pom.xml
+++ b/nucleus/parent/pom.xml
@@ -143,8 +143,8 @@
 
         <!-- Jakarta Persistence -->
         <jakarta.persistence-api.version>3.2.0</jakarta.persistence-api.version>
-        <eclipselink.version>5.0.0</eclipselink.version>
-        <eclipselink.asm.version>9.9.0</eclipselink.asm.version>
+        <eclipselink.version>5.0.1-RC2</eclipselink.version>
+        <eclipselink.asm.version>9.10.0</eclipselink.asm.version>
 
         <!-- Jakarta Mail -->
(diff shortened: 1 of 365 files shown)


qa/pom.xml
```diff
@@ -33,7 +33,7 @@
                 </plugin>
                 <plugin>
                     <artifactId>maven-surefire-plugin</artifactId>
-                    <version>3.5.5</version>
```
dmatej, line 36: Why do you need a milestone here?

nucleus/parent/pom.xml
```diff
@@ -275,10 +275,10 @@
         <jmh.version>1.37</jmh.version>
         <osgi-resource-locator.version>3.0.0</osgi-resource-locator.version>
         <slf4j.version>2.0.18</slf4j.version>
-        <shrinkwrap.version>3.3.5</shrinkwrap.version>
+        <shrinkwrap.version>3.3.7</shrinkwrap.version>
         <testcontainers.version>2.0.5</testcontainers.version>
 
-        <surefire.version>3.5.5</surefire.version>
```
dmatej, line 281: ... and here

dmatej (approved):

OndroMih (approved):

nucleus/parent/pom.xml
arjantijms: No specific need, just updated to the latest of what was out there and not specifically for the mythical Maven 4.

(it helped, as it did catch an error that was tolerated before, but now flagged as error)

PR #26103, opened 2026-06-16
Sync domain-passwords.p12 to instances after PKCS12 migration

renatsaf:
## Problem

This is a regression introduced in 7.1.0. Please backport also to 7.1.x stream (7.x branch).

Password aliases stored in `domain-passwords.p12` are not synchronized to clustered instances on startup (regression in 7.1.0 / 8.0.x; 7.0.25 and earlier are fine). As a result, any instance-side use of an alias fails, e.g. a JDBC connection pool whose password is `${ALIAS=...}`:

```
There is no such alias [mssql-pwd] ... defined in the default store for this domain.
```

## Root cause

The list of config files synced to instances is hardcoded in `nucleus/cluster/admin/src/main/resources/META-INF/config-files`. It still named the alias keystore `domain-passwords` — the old JKS/JCEKS filename. Commit 096ace25 ("Switching from JKS and JCEKS to PKCS12") renamed it to `domain-passwords.p12` (`PasswordAdapter.PASSWORD_ALIAS_KEYSTORE`). The stale entry no longer matches any file, so the keystore is never propagated.

## Fix

- Update the sync list entry to `domain-passwords.p12`.
- Fix the matching stale references in the HA Administration Guide (`instances.adoc`).

## Notes

This restores the 7.0.x behavior. Reported and root-caused by @cluisquijada; triage/fix confirmed by @OndroMih in the issue.

Fixes #26093

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #26109, opened 2026-06-17
Migrate legacy JKS/JCEKS security stores to PKCS12 when upgrading from 7.0.x

### LINKED ISSUE #26107: SecurityUpgradeService should migrate keystore/cacerts/domain-passwords to PKCS12 when upgrading from 7.0.x or older
### Background

Follow-up requested in PR #26103 (regression fix for #26093): https://github.com/eclipse-ee4j/glassfish/pull/26103

Commit 096ace25 ("Switching from JKS and JCEKS to PKCS12") changed GlassFish 7.1.0+ to read and write its security stores exclusively in PKCS12 format under fixed filenames:

- `keystore.p12`
- `cacerts.p12`
- `domain-passwords.p12` (`PasswordAdapter.PASSWORD_ALIAS_KEYSTORE`)

A domain created by 7.0.x or earlier instead contains the legacy JKS/JCEKS files (`keystore.jks`, `cacerts.jks`, `domain-passwords`). When such a domain is opened by 7.1.0+, there is **no automatic migration**: `SecurityUpgradeService` only upgrades the JACC provider, the JDBC-realm digest algorithm, cleans up generated policy files, and warns on NSS — it does not convert the keystores. As a result an upgraded domain can fail to read its alias store / keystore, since the runtime only looks for the `.p12` filenames.

### Proposal

Extend `SecurityUpgradeService.postConstruct()` to detect legacy JKS/JCEKS stores in `<domain>/config` and convert them to PKCS12 under the new names, using the domain's master password:

- `keystore.jks` → `keystore.p12`
- `cacerts.jks` → `cacerts.p12`
- `domain-passwords` (JKS/JCEKS) → `domain-passwords.p12`

Notes / open questions:

- Migration should be idempotent and skip files that are already PKCS12 / already migrated.
- Decide whether to delete or retain (e.g. `.bak`) the legacy files after conversion.
- Needs the master password; should fail gracefully (clear log message) if conversion isn't possible.
- Coordinate with the cluster sync list fix in #26103 so migrated instances receive the `.p12` files.

### Related

- Regression fix: PR #26103 (`Fixes #26093`)
- Original PKCS12 switch: commit 096ace25

renatsaf:
## Problem

Commit `096ace25` ("Switching from JKS and JCEKS to PKCS12") changed GlassFish 7.1.0+ to read and write its security stores exclusively in PKCS12 under fixed file names (`keystore.p12`, `cacerts.p12`, `domain-passwords.p12`). A domain created by 7.0.x or older instead contains `keystore.jks`, `cacerts.jks` and a JCEKS `domain-passwords`. `SecurityUpgradeService` upgraded the JACC provider, the JDBC-realm digest, generated policy files and warned on NSS — but **never converted the keystores**, so an upgraded domain could fail to read its stores. Follow-up requested in PR #26103.

On top of that, the documented upgrade procedure (copy the old domain, run `asadmin start-domain --upgrade`) failed before the server — and therefore the migration — could even start (#25800): `asadmin` verifies the master password against `config/cacerts.p12`, which does not exist in a copied 7.0.x domain, so **every** password was rejected:

```
Enter master password (3) attempt(s) remain)>
Sorry, incorrect master password, retry
...
Number of attempts (3) exhausted, giving up
Command start-domain failed.
```

### Server side: migrate the stores during `start-domain --upgrade`

Extend `SecurityUpgradeService.postConstruct()` with a `migrateLegacyKeystores()` step that detects the legacy stores and converts them to PKCS12 under the new names using the domain master password:

- `keystore.jks` → `keystore.p12`
- `cacerts.jks` → `cacerts.p12`
- `domain-passwords` (JCEKS) → `domain-passwords.p12`
- `master-password` (JCEKS, saved master password in the domain dir, protected by a fixed password) → `master-password.p12`

After a successful migration, `javax.net.ssl.keyStore`/`javax.net.ssl.trustStore` JVM options in every `java-config` that still point to `.../config/keystore.jks` / `.../config/cacerts.jks` are rewritten to the `.p12` successors — otherwise they would reference the renamed `.bak` files. Options pointing to custom (non-migrated) stores are left untouched.

Properties:

- **Idempotent** — a store is skipped if the legacy file is absent or the `.p12` target already exists; JVM options are only rewritten while they reference a migrated store.
- **Non-destructive** — each migrated legacy file is retained as `<name>.bak`.
- **Graceful** — if the master password is unavailable, or a conversion fails, it logs a warning (new `NCLS-SECURITY-01016…01020` messages), removes any half-written `.p12`, and never aborts the upgrade.
- Conversion uses `java.security.KeyStore` directly (no external `keytool` process) and copies private keys, trusted certificates and secret keys.

### Client side: let `asadmin` verify the master password against the legacy stores (#25800)

`LocalServerCommand.getJKS()` (used by `start-domain`, `stop-domain`, `restart-domain`, …) now falls back to `cacerts.jks` when `cacerts.p12` is absent, and `getMasterPasswordFile()` falls back to the legacy `master-password` JCEKS file — `KeyTool` loads stores via `KeyStore.getInstance(File, char[])`, which auto-detects JKS/JCEKS/PKCS12, so only the file names were the blocker. The same fallback is applied to the instance-side commands (`LocalInstanceCommand`, `CreateLocalInstanceCommand`, `ChangeNodeMasterPasswordCommand`). With this, `start-domain --upgrade` can verify the master password, launch the server and run the migration above.

Runs on the `start-domain --upgrade` path. NSS (`.db`) stores remain out of scope (existing warning unchanged). Coordinates with the cluster sync-list fix in #26103 so migrated instances receive the `.p12` files.

## Testing

Added `SecurityUpgradeServiceTest` covering the secret-key JCEKS `domain-passwords` case (entries and store type preserved) and the JVM-option rewrite (migrated stores rewritten; custom/non-migrated stores and unrelated options untouched). Affected modules (`common-util`, `admin/server-mgmt`, `cluster/cli`, `security/core`) compile and tests pass locally; JCEKS auto-detection by `KeyStore.getInstance(File, char[])` verified against a 7.0-style `master-password` store.

Fixes #26107
Fixes #25800

### DIFF
--- a/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/LocalServerCommand.java
+++ b/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/LocalServerCommand.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2025 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022, 2025, 2026 Contributors to the Eclipse Foundation
  * Copyright (c) 1997, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -55,8 +55,10 @@ import static com.sun.enterprise.universal.process.ProcessUtils.waitWhileIsAlive
 import static com.sun.enterprise.util.SystemPropertyConstants.KEYSTORE_PASSWORD_DEFAULT;
 import static com.sun.enterprise.util.SystemPropertyConstants.MASTER_PASSWORD_ALIAS;
 import static com.sun.enterprise.util.SystemPropertyConstants.MASTER_PASSWORD_FILENAME;
+import static com.sun.enterprise.util.SystemPropertyConstants.MASTER_PASSWORD_FILENAME_LEGACY;
 import static com.sun.enterprise.util.SystemPropertyConstants.MASTER_PASSWORD_PASSWORD;
 import static com.sun.enterprise.util.SystemPropertyConstants.TRUSTSTORE_FILENAME_DEFAULT;
+import static com.sun.enterprise.util.SystemPropertyConstants.TRUSTSTORE_FILENAME_LEGACY;
 import static java.lang.System.Logger.Level.DEBUG;
 
 /**
@@ -514,11 +516,18 @@ public abstract class LocalServerCommand extends CLICommand {
         if (serverDirs == null) {
             return null;
         }
-        File mp = new File(new File(serverDirs.getServerDir(), "config"), TRUSTSTORE_FILENAME_DEFAULT);
-        if (mp.canRead()) {
-            return mp;
+        File configDir = new File(serverDirs.getServerDir(), "config");
+        File trustStore = new File(configDir, TRUSTSTORE_FILENAME_DEFAULT);
+        if (trustStore.canRead()) {
+            return trustStore;
         }
-        LOG.log(DEBUG, "File does not exist or is not readable: {0}", mp);
+        // A server copied from 7.0.x or older still has the JKS stores; they are migrated to PKCS12
+        // by start-domain --upgrade, which must be able to verify the master password first.
+        File legacyTrustStore = new File(configDir, TRUSTSTORE_FILENAME_LEGACY);
+        if (legacyTrustStore.canRead()) {
+            return legacyTrustStore;
+        }
+        LOG.log(DEBUG, "Neither {0} nor {1} exists or is readable.", trustStore, legacyTrustStore);
         return null;
     }
 
@@ -530,7 +539,11 @@ public abstract class LocalServerCommand extends CLICommand {
 
         File mp = new File(serverDirs.getServerDir(), MASTER_PASSWORD_FILENAME);
         if (!mp.canRead()) {
-            return null;
+            // A server copied from 7.0.x or older still keeps the saved master password in a JCEKS store.
+            mp = new File(serverDirs.getServerDir(), MASTER_PASSWORD_FILENAME_LEGACY);
+            if (!mp.canRead()) {
+                return null;
+            }
         }
 
         return mp;
--- a/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/LocalServerCommand.java
+++ b/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/LocalServerCommand.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2025, 2026 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022, 2026 Contributors to the Eclipse Foundation
  * Copyright (c) 1997, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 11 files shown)


nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/LocalServerCommand.java
```diff
@@ -514,11 +516,18 @@ private File getJKS() {
...
-        File mp = new File(new File(serverDirs.getServerDir(), "config"), TRUSTSTORE_FILENAME_DEFAULT);
-        if (mp.canRead()) {
-            return mp;
+        File configDir = new File(serverDirs.getServerDir(), "config");
+        File trustStore = new File(configDir, TRUSTSTORE_FILENAME_DEFAULT);
+        if (trustStore.canRead()) {
+            return trustStore;
         }
-        LOG.log(DEBUG, "File does not exist or is not readable: {0}", mp);
+        // A server copied from 7.0.x or older still has the JKS stores; they are migrated to PKCS12
```
OndroMih: I think that having the code that checks the legacy keystores in each command is an overkill. The user should run start-domain --upgrade and all the keystores should be upgraded before any other admin command is called. Then admin commands would already have the migrated keystores and wouldn't need to care about the legacy keystores.

So it's enough to have this fallback only in the StartDomainCommand class, which needs it to verify that the user can run the --upgrade action.

OndroMih (commented):
Looks OK, I would just keep the fallback to the old keystores in StartDomainCommand, not in other commands.

nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/LocalServerCommand.java
renatsaf: Good point, done in d7114dc229: moved the legacy-JKS/JCEKS fallback out of `LocalServerCommand.getJKS()`/`getMasterPasswordFile()` (which is back to PKCS12-only, like before this PR) and into overrides in `StartDomainCommand`, since that's the only command that needs to verify the master password before running `--upgrade`. Other commands built on `LocalServerCommand` no longer see the legacy stores at all.

OndroMih:
@renatsaf , also revert modifications to ChangeNodeMasterPasswordCommand.java, CreateLocalInstanceCommand.java, and LocalInstanceCommand.java - they also don't need to respect old jks files and can use only the p12 files, assuming that upgrade already happened.

renatsaf:
@OndroMih Done for two of the three in 862c7365bf: `ChangeNodeMasterPasswordCommand` and `CreateLocalInstanceCommand` are back to the PKCS12-only lookup. You're right that those read the instance's `config` directory, which is synchronized from the DAS, so after `start-domain --upgrade` they can only ever see `keystore.p12`/`cacerts.p12`.

I left the fallback in `LocalInstanceCommand.getMasterPasswordFile()` though, because I don't think the "upgrade already happened" assumption covers it. That method resolves `nodes/<node>/agent/master-password`, which is outside the domain directory and typically on a different host — `SecurityUpgradeService` only converts stores under `<domain>` and `<domain>/config`, and synchronization never touches the node agent directory. So a node created by 7.0.x keeps its JCEKS `master-password` indefinitely, and without the fallback `start-local-instance` would stop finding the saved master password on every remote node after the upgrade.

Happy to change the approach if you prefer — the alternative would be to convert `agent/master-password` in place instead of falling back to it (it's protected by the fixed `master-password` password, so no user input is needed). Just let me know which you'd rather have.

OndroMih:
I agree, @renatsaf . I would prefer the alternative approach and convert agent/master-password in place, so that afterwards the old approach is not needed and could be dropped in later GlassFish versions.

renatsaf:
@
Done in a024202224, @OndroMih. `LocalInstanceCommand.getMasterPasswordFile()` now converts the node agent `master-password` (JCEKS) to `master-password.p12` in place on first use — keeping the old file as `master-password.bak` — instead of falling back to reading it. So once an upgraded node runs its first local-instance command, only the `.p12` remains and the legacy handling can be dropped entirely in a later GlassFish version.

The conversion reads the legacy store explicitly as JCEKS (KeyTool/PasswordAdapter can't open the extension-less `master-password` file) and writes the PKCS12 via PasswordAdapter; it needs no user input since the store is protected by the fixed `master-password` password. Added two tests (migration round-trip + no-store).
@

OndroMih (approved):

PR #26111, opened 2026-06-17
Synchronization with later branches 7.x and main

dmatej:
* https://commons.apache.org/proper/commons-io/changes.html: 2.18.0 => 2.22.0
* https://commons.apache.org/proper/commons-lang/changes.html: 3.17.0 => 3.20.0
* JBoss Logging
  * https://github.com/jboss-logging/jboss-logging/releases/tag/v3.6.3.Final
  * https://github.com/jboss-logging/jboss-logging/releases/tag/v3.6.2.Final
* https://github.com/jboss-javassist/javassist/releases/tag/rel_3_31_0_ga
* https://asm.ow2.io/versions.html: 9.8 => 9.10.1
* Ant
  * https://dlcdn.apache.org/ant/README.html
  * https://downloads.apache.org/ant/RELEASE-NOTES-1.10.16.html
* Upgrades Jackson because Jersey upgrade requires Jackson >= 2.19.0
  * https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.19
  * https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.20
  * https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.21
  * https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.22
* https://github.com/FasterXML/java-classmate/compare/classmate-1.7.0...classmate-1.7.3
* https://github.com/jettison-json/jettison/releases/tag/jettison-1.5.5
* https://github.com/eclipse-ee4j/metro-mimepull/releases/tag/1.11.0
* https://docs.junit.org/5.14.4/release-notes.html
* https://github.com/apache/felix-dev/tree/master/webconsole#releases: 5.0.10 => 5.0.18


nucleus/parent/pom.xml
```diff
@@ -132,43 +139,57 @@
...
         <jsch.version>0.2.24</jsch.version>
         <antlr.version>2.7.8</antlr.version>
-        <ant.version>1.10.15</ant.version>
-        <ant.external.version>1.10.2</ant.external.version>
-        <jackson.version>2.18.3</jackson.version>
-        <fasterxml.classmate.version>1.7.0</fasterxml.classmate.version>
+        <!-- Keep Ant versions in separated properties to be able to test snapshot while keeping antrun runnable -->
+        <ant.version>1.10.17</ant.version>
+        <ant-osgi.version>1.10.17</ant-osgi.version>
+        <jackson.version>2.22.0</jackson.version>
```
OndroMih, line 159: Isn't it enough to upgrade only to Jackson 2.19.4 for 7.0.x GlassFish branch? We aim for maximum stability, mostly CVE and critical fixes, so I would choose minimal upgrades.

Alternatively, we can also upgrade to 2.21. as the 2.21 Jackson version is LTS while 2.22 is not an LTS.

dmatej: In general it depends, usually it is the best option to use the latest patch version and consider even minor version upgrades, but we both know that some projects struggle with compatibility (including GlassFish).

I will document later in the description why I chose which change. Now I am running in iterations especially because 7.0.25 tests seem not much stable these days. 

Ant - we had issues with it before, that is why we mixed several ant versions in dependencies of submodules. Then we fixed the "osgi repackage" and it made things much easier.

Jackson - 2.22 is latest patch, if it passes all tests, I would not hold it back. Note that there's already 3.1 and we did not update even main branch - dependabot doesn't offer it as it changed group id. Previous 2.18.3 failed some tests in combination with other already done updates. If we would go just to 2.19, nobody would care if we would find some bug. Also I don't believe much in those Jackson LTS versions there as they don't receive any patches.

When I finish with all tests and some would detect some related error, we could try degrading the version. Also, in older versions is higher risk of discovering some CVEs. That applies to every dependency.

dmatej: Also, at this moment this branch is rather some "sniffing around" what is possible with bringing this branch back to live again. When it passes all tests including TCK, I will reconstruct it, separate some commits in a cleaner way, etc.

OndroMih: OK, no problem. In this case it seems that the Jackson upgrade is pretty safe. 

In general, I would be cautious and always consider upgrading only patch versions in 7.0. For Jackson, we already need to upgrade it from 2.18 to a higher minor version because of the Jersey upgrade.

OndroMih (approved):

PR #26115, opened 2026-06-18
Extra Header Validation

Pandrex247:
Spun off from https://github.com/eclipse-ee4j/glassfish-grizzly/pull/2237

### DIFF
--- a/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/services/impl/GlassfishNetworkListener.java
+++ b/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/services/impl/GlassfishNetworkListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, 2025 Contributors to the Eclipse Foundation
+ * Copyright (c) 2021, 2026 Contributors to the Eclipse Foundation
  * Copyright (c) 2007, 2022 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -60,6 +60,7 @@ import org.glassfish.hk2.utilities.AbstractActiveDescriptor;
 import org.glassfish.hk2.utilities.BuilderHelper;
 import org.glassfish.hk2.utilities.ServiceLocatorUtilities;
 import org.glassfish.internal.grizzly.V3Mapper;
+import org.glassfish.kernel.KernelLoggerInfo;
 import org.jvnet.hk2.config.types.Property;
 
 public class GlassfishNetworkListener extends GenericGrizzlyListener {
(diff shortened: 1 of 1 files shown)


OndroMih:
Hi @Pandrex247, which issue does this address? Grizzly should handle this already by ignoring Conten-tLength header if Transfer-Encoding header is present, as explained in https://github.com/eclipse-ee4j/glassfish-grizzly/pull/2237 you linked. Is there a reason why GlassFish needs this validation on top of Grizzly and refuse requests with both headers instead of just ignoring the Content-Length header?

Pandrex247:
Grizzly itself does, but not necessarily Payara/GlassFish. This is stricter validation.
We had a reproducer very similar to one of the recent CVE fixes that Grizzly wasn't catching (I previously left a comment on the issue in GitLab)

OndroMih:
@Pandrex247 , I replied to you on Gitlab: https://gitlab.eclipse.org/security/vulnerability-reports/-/work_items/293#note_7934313

In short, I believe that the fix in Grizzly is enough and it works in GlassFish 8.0.3, this this PR is not needed. It adds a performance penalty and I don't see how it addresses any issue that isn't already addressed.

dmatej:
Can we close this? The fix was distributed with Grizzly 5.0.2 and GlassFish 8.0.3. Basically it is enabled the strict validation by default, while older releases were benevolent by default. The extra validation seems redundant to me now.

OndroMih:
I understand now.

Grizzly now addresses the issue by ignoring the Content-Length header and continues processing.

This PR for GlassFish instead raises an error in this situation and doesn't continue processing. This is explained in https://gitlab.eclipse.org/security/vulnerability-reports/-/work_items/293#note_8157118

The RFC explicitly says that the Content-Length header should be ignored, but also strongly suggests raising an error instead of processing normally without the Content-Length header. 

So the solution proposed here is more robust, however it introduces performance penalty because the check happens for every request and is not very cheap.

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/services/impl/GlassfishNetworkListener.java
```diff
@@ -401,10 +402,19 @@ protected boolean onHttpHeaderParsed(final HttpHeader httpHeader,
...
                 final FilterChainContext ctx) {
 
-            final boolean result = super.onHttpHeaderParsed(httpHeader,
+            boolean result = super.onHttpHeaderParsed(httpHeader,
                     buffer, ctx);
 
             final HttpRequestPacket request = (HttpRequestPacket) httpHeader;
+
+            // Here we can add extra validation to prevent Http Server invalid use of headers
+            if (request.containsHeader(Header.ContentLength) && request.containsHeader(Header.TransferEncoding)) {
```
OndroMih, line 411: This doesn't seem to be a valid check because Grizzly always computes the value of the Content-Length header from the content so I believe that `request.containsHeader(Header.ContentLength)` always returns `true`. That would lead to refusing all requests that contain `Transfer-Encoding` header. 

See: https://github.com/eclipse-ee4j/glassfish-grizzly/blob/52ef17dfe6d5f71016d7584d3f92c2157af3739b/modules/http/src/main/java/org/glassfish/grizzly/http/HttpHeader.java#L901

@Pandrex247, can you verify that correct requests with just the `Transfer-Encoding` are not always refused?

Pandrex247: Apologies, this got lost in the flood of GitHub notifications I usually get.
I'll see if I can find some time to test on Payara/GlassFish as another issue just got raised on GitLab which seems related (issue 867)

OndroMih: @Pandrex247 , did you have some time to verify this?

PR #26124, opened 2026-06-23
No Executor nodes available

fredg02:
https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/work_items/7610

Keep only one build with artifacts.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -257,8 +257,10 @@ pipeline {
    }
 
    options {
-      buildDiscarder(logRotator(numToKeepStr: '2'))
+      buildDiscarder(logRotator(numToKeepStr: '2', artifactNumToKeepStr: '1'))
 
+      disableConcurrentBuilds()
+     
       parallelsAlwaysFailFast()
 
       // to allow re-running a test stage


dmatej:
> https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/work_items/7610
> 
> Keep only one build with artifacts.

That is a problem, sometimes, because when we have some flaky build, it is useful to compare logs (artifacts) between executions. However that can be enabled on demand. 
I did not notice your PR, so this morning I started to play with Jenkins here: https://github.com/eclipse-ee4j/glassfish/pull/26125

First I had the same idea, to reduce this, but yesterday I made some list of things we could do, so I am following that.

* Stashes are useful just when we want to rerun just tests; we usually rerun tests just for failures, so we can delete stashes if the build passed.
* I don't know how to delete archived logs if the build passed. But maybe it is not necessary.
* We used archived artifacts to run the TCK, however that is not possible any more, because these Jenkins instances are isolated now. Maybe we could kick them out for now.

Now I am trying to find any large garbage we keep on the disk when job finished.

dmatej:
Finished in #26125.
Concurrent builds are still allowed, the setup allows it.

PR #26129, opened 2026-06-27
Released 7.0.26

dmatej:


arjantijms (approved):

Jenkinsfile
```diff
@@ -449,4 +452,16 @@ pipeline {
          }
       }
    }
+
+   post {
+      // Some files allocate a lot of disk space, but when the build succeeded, we often don't need them any more.'
+      always {
+         cleanWs()
```
fredg02, line 459: Since the build runs on epheremal build agents, there is no need to manually clean up the workspace. At the end of the build, the agent is thrown away anyway.

dmatej: I thought so, but as at least one disk is persistent (m2 cache), so I tried to be rather defensive. It takes few seconds.

PR #26138, opened 2026-07-08
Avoid going through `Optional`

pzygielo:


appserver/appclient/client/acc-standalone/src/main/java/org/glassfish/appclient/client/acc/agent/GFDependenciesClassLoader.java
```diff
@@ -104,7 +104,7 @@ private static URL[] findGlassFishJars() {
         Stream<File> modules = findFiles(gfModules, moduleFilter);
         Stream<File> libs = findFiles(appInstall, libFilter);
         Stream<File> derbyLibs = findFiles(derbyLibDir, derbyFilter);
-        return Stream.of(modules, libs, derbyLibs).reduce(Stream::concat).orElseGet(Stream::empty).map(TO_URL)
+        return Stream.of(modules, libs, derbyLibs).flatMap(s -> s).map(TO_URL)
```
dmatej, line 107: `s -> s` is the same as `Function.identity()`, am I right? 
But ... where is the `Optional`? Aha, I see, the original code's `Stream.reduce` method returns `Optional<Stream<File>>`, while `Stream.empty` returns `Stream<File>`. 

And you are right: 

> "Each mapped stream is closed after its contents have been placed into this stream. (If a mapped stream is null an empty stream is used, instead.)"

dmatej (approved):

PR #26139, opened 2026-07-10
Use default value for `.publish`

pzygielo:
The only other explicit `publish` I've found is (setting to the non-default): https://github.com/eclipse-ee4j/glassfish/blob/a51f9c4ed6b024223737c7d39b975151d8007837/appserver/connectors/work-management/src/main/java/com/sun/enterprise/connectors/work/OneWork.java#L33-L37

OTOH - this `publish` flag does not seem to be used by GF at all.

Maybe it _could be used_ or - dropped from `@LogMessageInfo`? :thinking:

### DIFF
--- a/appserver/connectors/work-management/src/main/java/com/sun/enterprise/connectors/work/CommonWorkManager.java
+++ b/appserver/connectors/work-management/src/main/java/com/sun/enterprise/connectors/work/CommonWorkManager.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
  * Copyright (c) 1997, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -71,8 +72,7 @@ public final class CommonWorkManager implements WorkManager {
             comment = "Failed to find thread pool",
             level = "SEVERE",
             cause = "Could not find a thread pool according to the pool ID.",
-            action = "Check the thread-pool-id property in Resource Adapter Config.",
-            publish = true)
+            action = "Check the thread-pool-id property in Resource Adapter Config.")
     private static final String RAR_THREAD_POOL_NOT_FOUND = "AS-RAR-05001";
 
     @LogMessageInfo(
@@ -80,8 +80,7 @@ public final class CommonWorkManager implements WorkManager {
             comment = "Failed to find the default thread pool.",
             level = "SEVERE",
             cause = "Could not find the default thread pool for resource adatper.",
-            action = "Check the thread-pool-id property in Resource Adapter Config.",
-            publish = true)
+            action = "Check the thread-pool-id property in Resource Adapter Config.")
     private static final String RAR_DEFAULT_THREAD_POOL_NOT_FOUND = "AS-RAR-05002";
 
     /**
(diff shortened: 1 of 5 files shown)


dmatej (approved):
My question is what exactly the `publish` argument does?
Relates to https://github.com/eclipse-ee4j/glassfish-logging-annotation-processor/pull/107

pzygielo:
> My question is what exactly the `publish` argument does? 

- Good question that I also ask myself. But as this property was not placed in properties - it was also not used so far on GF side (At least **I** failed to locate such usage).
Per https://github.com/eclipse-ee4j/glassfish-logging-annotation-processor/blob/d2c1831ea651b9bbf6c2813dec30748e4b4e1cb2/src/main/java/org/glassfish/logging/annotation/LogMessageInfo.java#L33 it is described as:
```
publish: Boolean value indicates whether this log message should be
         published in the Error Reference guide. (default: true)
```
(a bit similar to https://github.com/eclipse-ee4j/glassfish-logging-annotation-processor/blob/d2c1831ea651b9bbf6c2813dec30748e4b4e1cb2/src/main/java/org/glassfish/logging/annotation/LoggerInfo.java#L29).
- My initial thought was: "Maybe it is or could be used by `glassfish-doc-maven-plugin`". I could not confirm this to be the case. I also do not think java sources (out of /docs) or binaries are used for documentation preparation, but I'm not 100% sure about that.
- Also coming from https://github.com/eclipse-ee4j/glassfish-logging-annotation-processor/pull/107
   - that whole `_details` properties with `action`, `cause` and `level` (and possibly  publish) - what are they used for? Were they used in GF ever before (i.e. before migration to JUL)? Perhaps we could drop (first from glassfish-logging-annotation-processor) ALL attributes that land in `_details` and not produce `_details`? I'm slowly exploring this.
   - I'm also learning how files generated into `META-INF`
      - `loggerinfo/LoggerInfoMetadata.properties`
      - `logmessages/LogMessagesMetadata.properties`
     are (or are not) being used.
     (And this feels like I've already done that few summers ago (the commit in eclipse-ee4j/glassfish-logging-annotation-processor#107 was _waiting_ since 2022, but I remember nothing now...)


But - by using default for `publish` - it will leave only one place to modify if `publish` is removed from `@`. :grin:

OR GF could use `_details` in logging. (Or I could find that is does.)

dmatej:
Aha, I believe around 2003 there was a trend to create multilanguage applications including logging (JUL was added in Java 1.4.2 if I remember well), so at some point they tried to integrate everything to the Ant build, later to Maven, but never finished. 
I think some man pages are generated somehow (different story), I am also quite sure somewhere was a list of those "id based" log messages.

At some point there was a ML distribution of GlassFish, surprisingly some versions differed in behavior between ML and SL, so I remember far far in the past we were switching between them with SJAS9. 

However now we slowly remove these things, logs are in plain english, UI too, so this doesn't make much sense. Perhaps with AI, however AI can read logs without a need of IDs and translations, so ... no.

pzygielo:
[Error Message Reference](https://glassfish.org/docs/latest/error-messages-reference.html) lists some IDs but they seem to come from checked-in `docs/error-messages-reference/src/main/asciidoc/error-messages.adoc`. (And they have `cause` and `action`.) That's why I suspected `glassfish-doc-maven-plugin` to play a part.

I'm yet to learn the level of consistency there. :smile:

PR #26144, opened 2026-07-12
Upgrade Nimbus 9.47 -> 10.4.2

OndroMih:
To fix CVE-2025-53864

Nimbus 10.4.2 is the oldest version with the fix that is compatible with Soteria and Omni-JWT.

Nimbus changelog: https://bitbucket.org/connect2id/nimbus-jose-jwt/src/master/CHANGELOG.txt

Newer versions already have a newer Nimbus and aren't impacted.

### DIFF
--- a/appserver/pom.xml
+++ b/appserver/pom.xml
@@ -99,10 +99,10 @@
         <jakarta.interceptor-api.version>2.1.0</jakarta.interceptor-api.version>
 
         <!-- Jakarta Security + Authentication/Authorization -->
-        <soteria.version>3.0.3</soteria.version>
+        <soteria.version>3.0.4</soteria.version>
         <exousia.version>2.1.3</exousia.version>
         <epicyro.version>3.0.0</epicyro.version>
-        <nimbus.version>10.0.2</nimbus.version>
+        <nimbus.version>10.4.2</nimbus.version>
         <jcip.version>1.0.2</jcip.version>
 
         <!-- Jakarta Messaging -->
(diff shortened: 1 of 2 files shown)


arjantijms:
Note that this may trigger the OSGi protection guards for Soteria as the dependency version number changes a major version. May need to either patch the protection file (MANIFEST.MF) in Soteria, or release a new Soteria version for that range.

OndroMih:
You're right. The 2 tests in the JakartaSecurityTest class fail, because of:

```
Caused by: org.osgi.framework.BundleException: Unable to resolve org.omnifaces.jwt [32](R 32.0): missing requirement [org.omnifaces.jwt [32](R 32.0)] osgi.wiring.package; (&(osgi.wiring.package=com.nimbusds.jose)(version>=9.23.0)(!(version>=10.0.0))) Unresolved requirements: [[org.omnifaces.jwt [32](R 32.0)] osgi.wiring.package; (&(osgi.wiring.package=com.nimbusds.jose)(version>=9.23.0)(!(version>=10.0.0)))]
```

OndroMih:
It should be enough to raise the version of omni-jwt from 2.0.2 to 2.0.3, which raises Nimbus version to 10.4 and doesn't change anything else in the functionality or dependencies: https://github.com/OmniFish-EE/omni-jwt/releases/tag/2.0.3

OndroMih:
In the end, I had to upgrade Soteria and Omni JWT to a newer patch version, they didn't support Nimbus >10. And then I upgraded also Nimbus to 10.4.2, because newer versions of Soteria and Omni JWT required it.

dmatej (approved):

PR #26145, opened 2026-07-12
Fix JNDI validation failure for app-scoped resources on server restart

renatsaf:
Fixes #25961

Applications with app-scoped resources defined in `glassfish-resources.xml` (e.g. `java:app/jdbc/mydb`) deployed fine but failed to load after `asadmin restart-domain` with:

```
DeploymentException: JNDI lookup failed for the resource: Name: my_persistence_unit, Lookup: java:app/jdbc/mydb, Type: javax.sql.DataSource
```

leaving the application inaccessible (404) until manual redeploy.

## Root cause

The deploy path (`ResourcesDeployer.processArchive`) parses `glassfish-resources.xml` and publishes the declared JNDI names into the deployment context as `APP_SCOPED_RESOURCES_JNDI_NAMES` transient metadata, which `ResourceValidator` loads into its in-memory namespace — so validation passes without a live JNDI lookup.

The load path (server restart, app/app-ref enable) reads the resource config from domain.xml via `populateResourceConfigInAppInfo()` but never produced this metadata. The validator then fell back to a live `InitialContext.doLookup()`, which runs before the app-scoped resources are bound to naming (and without a component invocation context for `java:app`), so it always failed and aborted the deployment.

## Changes

- **Commit 1**: `ResourcesDeployer.populateResourceConfigInAppInfo()` now also publishes `APP_SCOPED_RESOURCES_JNDI_NAMES` built from the domain.xml config beans, keyed identically to the deploy path (app name / module name). Includes unit tests.
- **Commit 2**: Restores `ResourceValidator` activation. Commit 6f269c13a0 removed the seemingly unused `ResourceValidator` injection from `DolProvider`, but that injection was the only thing instantiating the validator (whose `postConstruct` registers it as a deployment event listener — nothing else references its `ResourceValidatorVisitor` contract). Since then resource validation never ran at all on 8.0.3+, which also masked this issue there: invalid resource references were no longer rejected at deployment time and surfaced later with less helpful errors (e.g. from JPA prepare). Re-added with a comment explaining the injection is load-bearing.

## Verification

Using the reporter's Derby-based reproducer from the issue (deploy → visit → `restart-domain` → visit), on Windows 11 / JDK 25:

| Build | Restart result |
|---|---|
| 8.0.0 release, stock | 404, exception from the issue |
| 8.0.0 release + patched `resources-runtime.jar` | 200 across two restart cycles |
| current main, stock | 200, but only because validation never runs |
| current main + commit 2 only | 404 — the bug resurfaces once validation is restored |
| current main + both commits | 200 |

Negative control on both fixed builds: an app referencing a nonexistent datasource (`java:app/jdbc/DOESNOTEXIST`) is still rejected at deploy time with the validator's error, so validation is not weakened — the load path now only recognizes names that are genuinely declared in the application's resource configuration.

### DIFF
--- a/appserver/deployment/jakartaee-core/src/main/java/org/glassfish/javaee/core/deployment/DolProvider.java
+++ b/appserver/deployment/jakartaee-core/src/main/java/org/glassfish/javaee/core/deployment/DolProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023, 2026 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022, 2026 Contributors to the Eclipse Foundation
  * Copyright (c) 2009, 2022 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
--- a/appserver/deployment/jakartaee-core/src/main/java/org/glassfish/javaee/core/deployment/DolProvider.java
+++ b/appserver/deployment/jakartaee-core/src/main/java/org/glassfish/javaee/core/deployment/DolProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022, 2023, 2026 Contributors to the Eclipse Foundation
  * Copyright (c) 2009, 2022 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -31,6 +31,7 @@ import com.sun.enterprise.deployment.archivist.DescriptorArchivist;
 import com.sun.enterprise.deployment.deploy.shared.DeploymentPlanArchive;
 import com.sun.enterprise.deployment.deploy.shared.InputJarArchive;
 import com.sun.enterprise.deployment.deploy.shared.Util;
+import com.sun.enterprise.deployment.util.ResourceValidator;
 import com.sun.enterprise.util.LocalStringManagerImpl;
 import com.sun.enterprise.util.io.FileUtils;
 import com.sun.enterprise.v3.common.HTMLActionReporter;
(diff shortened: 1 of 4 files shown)


appserver/deployment/jakartaee-core/src/main/java/org/glassfish/javaee/core/deployment/DolProvider.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022, 2023, 2026 Contributors to the Eclipse Foundation
```
pzygielo: 1. it no longer needs to be updated
2. if updated - it shall be range ("...(c) [initial], [the latest] Contributors..."), not list

renatsaf: Thanks — changed to the range form in 007bec5a6, here and in ResourcesDeployer.java (which had the same list-form update).

appserver/resources/resources-runtime/src/test/java/org/glassfish/resources/module/ResourcesDeployerTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+        Map<String, List<SimpleJndiName>> result =
+            ResourcesDeployer.collectAppScopedResourceJndiNames(app, APP_NAME);
+
+        assertThat(result.get(APP_NAME), containsInAnyOrder(
+            new SimpleJndiName("java:app/jdbc/mydb"), new SimpleJndiName("java:app/jdbc/other")));
+        assertThat(result, not(hasKey("web.war")));
+    }
+
+    // --- config-bean proxies (no Mockito available in this module) ---
```
dmatej: Could Claude try to use EasyMock?
```
        <dependency>
            <groupId>org.easymock</groupId>
            <artifactId>easymock</artifactId>
        </dependency>
```

renatsaf: Done in 9e4db9af14 — replaced the Proxy helpers with EasyMock (createNiceMock/expect/replay) and added the managed easymock dependency; all 5 tests pass.

dmatej (approved):

PR #26146, opened 2026-07-13
Slim down Jenkins build resource requirements and disable concurrent builds

fredg02:
Also lower timeout to 4 hours.

Related to https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/work_items/7610

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -53,11 +53,11 @@ spec:
       mountPath: "/home/jenkins/.m2/repository/org/glassfish/main"
     resources:
       limits:
-        memory: "8Gi"
-        cpu: "5500m"
+        memory: "4Gi"
+        cpu: "4000m"
       requests:
-        memory: "8Gi"
-        cpu: "5500m"
+        memory: "4Gi"
+        cpu: "4000m"
   volumes:
   - name: "jenkins-home"
     emptyDir:
@@ -276,8 +276,10 @@ pipeline {
       // show timestamps in logs
       timestamps()
 
-      // global timeout, abort after 6 hours
-      timeout(time: 8, unit: 'HOURS')
+      // global timeout, abort after 4 hours
+      timeout(time: 4, unit: 'HOURS')
+
+      disableConcurrentBuilds()
    }
 
    stages {


Jenkinsfile
```diff
@@ -276,8 +276,10 @@ pipeline {
       // show timestamps in logs
       timestamps()
 
-      // global timeout, abort after 6 hours
-      timeout(time: 8, unit: 'HOURS')
+      // global timeout, abort after 4 hours
+      timeout(time: 4, unit: 'HOURS')
+
+      disableConcurrentBuilds()
```
dmatej, line 282: This would block processing PRs until the every single one finishes. Jobs are split to phases and stages exactly for that reason, to be able to run them in parallel as nodes are available.

In general Jenkins performance is deteriorating, we managed to get on original times by more parallel processing months ago.

The build failed now, because Jenkins server is somehow broken (restart failed?), all I can see is just some textual response: https://ci.eclipse.org/glassfish/

Jenkinsfile
```diff
@@ -53,11 +53,11 @@ spec:
       mountPath: "/home/jenkins/.m2/repository/org/glassfish/main"
     resources:
       limits:
-        memory: "8Gi"
-        cpu: "5500m"
+        memory: "4Gi"
```
dmatej, line 56: I guess this will not be enough, some tests run several JVMs. Maybe it could pass as we fixed a lot of memory leaks in the past, but some integration tests still need memory.

We can try that, however, the original values were tested somewhere in 2022 and when we tried to change them in the past, results were always worse, but any improvement is welcome.

Jenkinsfile
fredg02: > This would block processing PRs until the every single one finishes. Jobs are split to phases and stages exactly for that reason, to be able to run them in parallel as nodes are available.

It's a good practice to split up the jobs into different phases. During the test phase you are still spawing ~25+ pods in parallel, AFAICT. If 6 different PRs trigger builds at the same time (like yesterday) this results in resource exhaustion, keeping in mind that Glassfish has a massive amount of resources assigned to it.

In a world of infinite resources, this would probably work just fine. Here it doesn't for more than one reason. Since I don't have the cycles to search for the golden bullet, I can only offer you more robust builds that might have to run more sequentially and take a bit longer. I'd like to focus on robustness first and consider performance (speed) as a second step.

Please note, `disableConcurrentBuilds()` might also not solve all the problems at once.

Jenkinsfile
fredg02: At least one build ran fine. I'm testing it again now.

fredg02: There were two more successful runs (https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/view/change-requests/job/PR-26146/4 and the one before), so I guess the resource limits seem to work.

dmatej:
Last two builds: 35 and 34 minutes. Older of another PR 25 and 29 minutes.

fredg02:
> Last two builds: 35 and 34 minutes. Older of another PR 25 and 29 minutes.

There might still be potential to improve build times. For now, I'd rather wait 10 minutes longer but have successful builds and no crashes.

dmatej:
The last build failed because you were doing some work on your side or is it some new problem?

```
07:41:43  + mvn -V -B -e -fae clean validate -Ptck,set-version-id,snapshots
07:41:43  Apache Maven 3.9.11 (3e54c93a704957b63ee3494413a2b544fd3d825b)
07:41:43  Maven home: /usr/share/maven
07:41:43  Java version: 21.0.9, vendor: Eclipse Adoptium, runtime: /opt/java/openjdk
07:41:43  Default locale: en_US, platform encoding: UTF-8
07:41:43  OS name: "linux", version: "6.7.4-200.fc39.x86_64", arch: "amd64", family: "unix"
07:41:44  [INFO] Error stacktraces are turned on.
07:41:45  [INFO] Scanning for projects...
07:41:45  [INFO] Downloading from eclipse.maven.central.mirror: https://repo.eclipse.org/repository/maven-central/org/eclipse/ee4j/project/2.0.4/project-2.0.4.pom
07:41:45  [WARNING] Failed to create tracking file parent '/home/jenkins/.m2/repository/org/eclipse/ee4j/project/2.0.4/project-2.0.4.pom.lastUpdated'
07:41:45  java.nio.file.AccessDeniedException: /home/jenkins/.m2/repository/org/eclipse
07:41:45      at sun.nio.fs.UnixException.translateToIOException (UnixException.java:90)
07:41:45      at sun.nio.fs.UnixException.rethrowAsIOException (UnixException.java:106)
07:41:45      at sun.nio.fs.UnixException.rethrowAsIOException (UnixException.java:111)
07:41:45      at sun.nio.fs.UnixFileSystemProvider.createDirectory (UnixFileSystemProvider.java:462)
07:41:45      at java.nio.file.Files.createDirectory (Files.java:700)
07:41:45      at java.nio.file.Files.createAndCheckIsDirectory (Files.java:808)
07:41:45      at java.nio.file.Files.createDirectories (Files.java:794)
```

>> Last two builds: 35 and 34 minutes. Older of another PR 25 and 29 minutes.
>
> There might still be potential to improve build times. For now, I'd rather wait 10 minutes longer but have successful builds and no crashes.
> Please note, disableConcurrentBuilds() might also not solve all the problems at once.

Ok, would it make sense to try just this first, so builds would finish faster?

fredg02:
> The last build failed because you were doing some work on your side or is it some new problem?

Not sure if this aligns time-wise, but we had a repo.eclipse.org maintenance event on Thursday evening at 17:00 CEST.

> Ok, would it make sense to try just this first, so builds would finish faster?

What are you referring to by "try just this first"?

dmatej:
> What are you referring to by "try just this first"?

I meant the `disableConcurrentBuilds()`. Then we would race just with all other jobs, which are mostly sequential, but we would still get the lowest possible time for a single PR.

fredg02:
According to https://stackoverflow.com/questions/41492688/how-to-limit-jenkins-concurrent-multibranch-pipeline-builds the `disableConcurrentBuilds()` option only works on a per-branch basis, meaning that several different branches/PRs could still be built at the same time, which could trigger resource exhaustion. It's still a good first step to avoid running multiple builds on the same branch/PR at the same time. 

We will need to test plugins like [Throttle Concurrent Builds](https://plugins.jenkins.io/throttle-concurrents/).

dmatej:
> According to https://stackoverflow.com/questions/41492688/how-to-limit-jenkins-concurrent-multibranch-pipeline-builds the `disableConcurrentBuilds()` option only works on a per-branch basis, meaning that several different branches/PRs could still be built at the same time, which could trigger resource exhaustion. It's still a good first step to avoid running multiple builds on the same branch/PR at the same time.
> 
> We will need to test plugins like [Throttle Concurrent Builds](https://plugins.jenkins.io/throttle-concurrents/).

We already do the opposite, when somebody pushes changes to the same branch, we ask Jenkins to cancel the previous run instead of waiting until it finishes. See https://www.jenkins.io/doc/pipeline/steps/pipeline-milestone-step/#milestone-the-milestone-step-forces-all-builds-to-go-through-in-order .

Hmm, but that means that the `disableConcurrentBuilds` would disable this.

```Groovy
      stage('StopOld') {
         steps {
            script {
               milestone ordinal: Integer.parseInt(env.BUILD_NUMBER), label: "Build ${env.BUILD_NUMBER}"
            }
         }
      }
```

dmatej:
So, latest news in short:

* There's a problem with Jenkins/Kubernetes/backend, where is some 15 seconds timeout killing open inactive TCP connections. Errors are cached.
* The job killer I mentioned above is ineffective under heavy load, because to start aborting jobs it needs a free node. Then the queue must be simply processed or jobs removed manually.
* disableConcurrentBuilds is ineffective too, jobs will be just added to the queue later.
* Reducing memory allows to run more jobs at once, however not too much, and it makes jobs slower. In effect it doesn't help. 
* Single PR needs 29 nodes, while max I have seen running was 19. We should do something with that, so I should probably resume work on migration of Ant jobs to Maven - it makes these tests much faster and we can get rid of some parts of this "multiheaded zombie".
* #26155 Reduced the size of Junit Reports to process, so we don't hit the timeout any more. I have disabled also the Checkstyle reporting, because it made Jenkins UI very slow, showing the build result page took around a minute. And finally, one maven plugin poisoned Maven's truststore. It depended on the order of tests and if required artifacts were already in the cache, so sometimes it hit the wall, sometimes it did not.
* For some time after merging it jobs were randomly freezing, not sure why, build succeeded, then before it started the step with uploading results it paused for 30 minutes until I hard killed whole build. In the evening I started creating new PRs and rebasing some older, and merging them, and all problems with Jenkins went away. It took me a week of my life, but hooray ...

dmatej:
@fredg02 I will close it for now, as it seems, that @arjantijms used some AI black magic in #26191 with a success. I don't like it, because it made Jenkinsfile twice larger, does some things I don't like, but ... passed.
So perhaps I could try to make it more simple again in the main branch, but learn something from it.

Still weird that it worked in the past and now it doesn't without any changes on our side in last few weeks (last change was August 2nd, #26155 ). Last passing [build](https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/job/PR-26184/) was in August 12th.
Now even this PR doesn't pass.

PR #26148, opened 2026-07-15
Generate JMS admin password  of the broker on first run (fixes CVE-2026-22886)

OndroMih:
Fixes [CVE-2026-22886](https://nvd.nist.gov/vuln/detail/CVE-2026-22886).

When the default JMS broker instance is created, this generates a random password for both the client and in the broker configuration to keep them in synch.

WARNING: If upgrading an existing Glassfish installation, the existing JMS broker configuration isn't changed. It's recommended to change the default JMS admin password manually, as described in the GlassFish Administration guide. 

The generated password can be retrieved from the GlassFish configuration if needed: `get configs.config.config-name.jms-service.jms-host.jms-host-name.admin-password`

Knowing the generated password, it's possible to change the password to a custom value, as described in the guide.

## Why

Currently, by default the OpenMQ broker's admin interface is protected by the default well-known password. This allows anybody to access the admin interface if they have access to the JMS port.

It's possible to disable the admin interface but it's needed for several Glassfish commands, which would be broken without access to the OpenMQ admin interface. I chose the solution to generate the password because it doesn't break anything and should be safe as the default setup.

## How

The EMBEDDED and LOCAL type of the broker are initialized from the JMS resource adapter. The EMBEDDED is initialized lazily, the LOCAL is initialized immediately after GlassFish starts if it's not created yet.

The resource adapter provided by OpenMQ doesn't support providing the admin password to initialize the new broker instance with, and always uses the default admin password. That's why I extended it in the new GlassFishResourceAdapter class, which waits until the new broker instance is created in the filesystem and then modifies the admin password to a generated password and updates the admin password in the jms host in GlassFish configuration to match it.

A few more changes were required to align with this:
* the jms-ping command, if it was executed before the JMS broker was created, failed. The reason was that it used the default admin password and its action initialized the broker and the password get generated. The jms-ping command then failed because it operated on a resource that was created with the previous default password. I modified the jms-ping command so that when it fails and the password was meanwhile updated, it recreates the temporary resource and restarts the ping action with the new generated password
* An ant test was using hard-coded default admin password. I changed the tests to create a new test user on the JMS broker and use it instead of the admin user. This not only made the tests pass again but it's a good practice to use a test user intead of mis-using the existing admin user.


OndroMih:
@pzygielo , do you think it's worth porting this directly to the OpenMQ resource adapter? 

My new GlassFishResourceAdapter doesn't use any GlassFish-specific features, it only relies on the value of the "AdminPassword" property and updated the admin password in the broker config when the broker is created. It could also accept a property toggle, e.g. "UpdateAdminPassword" and only upadte the admin password if it's true.

OndroMih:
I want to review the clustered scenario and cover by tests. Just to be sure that the either that each instance generates the same password or that it works even if generated passwords are different on each instance.

OndroMih:
This solution fails in clustered setups.

I'll see if it's possible to generate a random password at server startup and save it in the configuration eagerly, if no brokers with that password are created yet. It would be better than generating the password lazily when a broker is created, possibly after a cluster instance is created with the default password.

dmatej:
There's also some 8.0.4-SNAPSHOT, current is 8.0.5-SNAPSHOT

PR #26154, opened 2026-07-27
Autodeployer and FileArchive improvements

dmatej:
- using Path instead of String and File
- added softlink resolution
- added protection against files escaping the archive file
- tested locally with future docker image improvements

### DIFF
--- a/nucleus/deployment/autodeploy/src/main/java/org/glassfish/deployment/autodeploy/AutoDeployer.java
+++ b/nucleus/deployment/autodeploy/src/main/java/org/glassfish/deployment/autodeploy/AutoDeployer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022, 2026 Contributors to the Eclipse Foundation
  * Copyright (c) 1997, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -20,7 +20,10 @@ package org.glassfish.deployment.autodeploy;
 import com.sun.enterprise.util.LocalStringManagerImpl;
 
 import java.io.File;
+import java.io.IOException;
 import java.net.URI;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
@@ -106,7 +109,7 @@ public class AutoDeployer {
     private static final boolean DEFAULT_INCLUDE_SUBDIR = false;
     private static final boolean DEFAULT_ENABLED = true;
 
-    static final String STATUS_SUBDIR_PATH = ".autodeploystatus";
+    static final Path STATUS_SUBDIR_PATH = Path.of(".autodeploystatus");
 
     /**
      * Creates a new autodeployer.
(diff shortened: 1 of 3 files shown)


nucleus/deployment/autodeploy/src/main/java/org/glassfish/deployment/autodeploy/AutoDeployer.java
```diff
@@ -246,54 +249,55 @@ boolean mkdirs(final File baseDir, final File targetDir) {
...
 
     private void validateAutodeployDirectory(String autodeployDirPath) throws AutoDeploymentException {
-        File autodeployDir = new File(autodeployDirPath);
+        Path autodeployDir = domainRoot().resolve(new File(autodeployDirPath).toPath());
         validateDirectory(autodeployDir);
-        File statusDir = new File(autodeployDir, STATUS_SUBDIR_PATH);
+        Path statusDir = autodeployDir.resolve(STATUS_SUBDIR_PATH);
         validateDirectory(statusDir);
+        Path bundlesDir = autodeployDir.resolve("bundles");
+        validateDirectory(bundlesDir);
```
OndroMih, line 263: This could throw an exception if the bundles directory doesn't exist and can't be created. 

It would mean that autodeployment would stop working i the autodeploy directory exists and contains apps, but it doesn't contain directory bundles if it cannot be created.

Is it OK?

OndroMih: Maybe this second call for "bundles" directory could be in a try catch block that catches AutoDeploymentException and just logs a warning instead of throwing it.

dmatej: Path.resolve doesn't throw, it doesn't validate existence, just possibility of construction of such Path object, ie using invalid characters. See UnirPath.resolve:
```Java
    // Resolve child against given base
    private static byte[] resolve(byte[] base, byte[] child) {
        int baseLength = base.length;
        int childLength = child.length;
        if (childLength == 0)
            return base;
        if (baseLength == 0 || child[0] == '/')
            return child;
        byte[] result;
        if (baseLength == 1 && base[0] == '/') {
            result = new byte[childLength + 1];
            result[0] = '/';
            System.arraycopy(child, 0, result, 1, childLength);
        } else {
            result = new byte[baseLength + 1 + childLength];
            System.arraycopy(base, 0, result, 0, baseLength);
            result[base.length] = '/';
            System.arraycopy(child, 0, result, baseLength+1, childLength);
        }
        return result;
    }
```

dmatej: If it would be true, some test would fail, or we don't have any coverage here?

dmatej: And the validateDirectory does this. I agree it is confusing, that is why I wrote you that I am quite sure we will come back to these classes.

```Java
    private void validateDirectory(Path dirFile) throws AutoDeploymentException {
        final Path realPath;
        if (Files.exists(dirFile)) {
            // Can be a softlink
            try {
                realPath = dirFile.toRealPath();
            } catch (IOException e) {
                throw new AutoDeploymentException(e);
            }
            if (!Files.isDirectory(realPath)) {
                throw new AutoDeploymentException(localStrings.getLocalString(
                    "enterprise.deployment.autodeploy.invalid_source_dir", "invalid source directory {0}", realPath));
            }
        } else {
            try {
                realPath = Files.createDirectories(dirFile);
            } catch (IOException e) {
                throw new AutoDeploymentException(e);
            }
        }
        if (!Files.isReadable(realPath)) {
            throw new AutoDeploymentException(localStrings.getLocalString(
                "enterprise.deployment.autodeploy.dir_not_readable", "directory {0} not readable", realPath));
        }
        if (!Files.isWritable(realPath)) {
            throw new AutoDeploymentException(localStrings.getLocalString(
                "enterprise.deployment.autodeploy.dir_not_writeable", "directory {0} not writable", realPath));
        }
    }
```

OndroMih: I meant validateDirectory should be wrapped in try catch and log a warning in case of AutoDeploymentException and continue. It now throws AutoDeploymentException if it can't create the directory, with this: 

```
try {
                realPath = Files.createDirectories(dirFile);
            } catch (IOException e) {
                throw new AutoDeploymentException(e);
            }
```

If the "bundles" directory can't be created, I don't think that AutoDeploymentException should bubble out - it's not a fatal issue, the autodeploy directory still works and can be used for app deployments. Instead, a warning should be logged and the autodeployer should initialize normally.

dmatej: I disagree, if we cannot control own directories, it will just crash later, but user will not receive any info. That is why I added it here, because it happened.

OndroMih: I was thinking how it would work e.g. in Docker if the directory is mounted as a volume and the bundle directory can’t be created. But you’re right - the current implementation requires writable autodeploy directory so bundles directory should be always possible to create.

It would be better if the autodeploy supported a read-only mode which is a better fit for Docker. But that’s out of scope of this PR.

OndroMih (approved):

nucleus/deployment/autodeploy/src/main/java/org/glassfish/deployment/autodeploy/AutoDeployer.java
dmatej: I think (not sure now) that I inverted some locations in the docker image, created a softlink and it just broke, but it was perhaps in May, so I forgot a bit.

I don't like autodeploy too much, because in the past I usually needed more control over the deployment, ability to process errors and warnings, etc. When we were in Payara for a while, I fixed the JSR-88 deployment, which had ugly API, but was somehow nice when I created another layer over it.

renatsaf: > I don't like autodeploy too much

I don't know any scenarios where auto-deploy can be used.

OndroMih: Auto deploy is often used in development and in Docker, where it's convenient to deploy applications just by copying the application into a directory.
However, it's much better to use the asadmin deploy command, which provides more control and configuration options.

renatsaf: > it's much better to use the asadmin deploy command

I agree.

PR #23495, opened 2021-06-25
Integrate Exousia 1.0.0-M1.

arjantijms:
This is the first step, which integrates Exousia for the Web component
only.

PR #26159, opened 2026-08-02
Issue #25903 Removal of the CachedPermission - dead code

dmatej:
- If it would not be dead, the code would throw NPE in tck tests all the time as noted in issue #25903
- I tried to compare the PermissionCache with 9.0 - no changes.
- Fixes #25903


OndroMih:
Yes, this is a dead code. It was made unused in this PR: https://github.com/eclipse-ee4j/glassfish/pull/23495 when refactored authorization was refactored and the logic was delegated to Exousia. The cache became unused but the code to create it and delete it remained. 

However, I don't see a similar cache implemented in Exousia. So I'm not sure if it's correct to just remove the dead code, or fix it and use used again, before delegating to Exousia in the permitAll method. Since the method `createWebResourcePermission`, which is needed to use the cache, was removed in the PR 23495, I assume that the intention was really to stop using the cache.

@arjantijms , what do you think? Do we need the uncheckedPermissionCache cache used in the `GlassFishAuthorizationService.permitAll` method, before delegating to Exousia?

arjantijms:
We've gone a while without the cache and did not see any issues.

Maybe create an issue in Exousia to investigate a potential replacement?

arjantijms (approved):

PR #26167, opened 2026-08-07
Fixed NPE when check 'behind-proxy' is enabled or not on an ssh node.

JackienStephan:
When `isBehindProxyEnabled()` method is executed on a remote node instance,
for example, execute the `start-instance` command on das, and the target is an instance on an ssh node,
a server element named "server" cannot be found in the instance's `domain.xml`.

So `domain.getServerNamed("server")` here returns null, resulting in NPE when call `getConfig()`.
It would be better to find the server element of the current instance.

With this issue, `start-instance` command targeting on ssh instances would succeed, but `list-instance` cannot get the correct running status of target instances. It would always be "stopped".
Then `stop-instance` command on the target instances always timeout.
The only temporary workaround I found for now, is to ssh to the remote instance server and execute `stop-local-instance`.

### DIFF
--- a/nucleus/admin/util/src/main/java/com/sun/enterprise/admin/util/GenericAdminAuthenticator.java
+++ b/nucleus/admin/util/src/main/java/com/sun/enterprise/admin/util/GenericAdminAuthenticator.java
@@ -323,14 +323,8 @@ public class GenericAdminAuthenticator implements AdminAccessController, JMXAuth
      * When enabled, the server trusts X-Real-IP and X-Forwarded-For headers.
      */
     private boolean isBehindProxyEnabled() {
-        // When this method is executed on a remote node instance,
-        // for example, execute the start-instance command on das, and the target is an instance on an ssh node,
-        // a <server> element named "server" cannot be found in the instance's domain.xml.
-        // would result in NPE when call "getConfig()" method.
-
-        // Config config = domain.getServerNamed("server").getConfig();
-
-        // it would be better to find the <server> element of the current instance.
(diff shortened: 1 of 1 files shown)


dmatej (approved):

nucleus/admin/util/src/main/java/com/sun/enterprise/admin/util/GenericAdminAuthenticator.java
```diff
@@ -322,7 +323,20 @@ public String getRemoteHost() {
...
     private boolean isBehindProxyEnabled() {
-        Config config = domain.getServerNamed("server").getConfig();
+        // When this method is executed on a remote node instance,
+        // for example, execute the start-instance command on das, and the target is an instance on an ssh node,
+        // a <server> element named "server" cannot be found in the instance's domain.xml.
+        // would result in NPE when call "getConfig()" method.
+
+        // Config config = domain.getServerNamed("server").getConfig();
+
+        // it would be better to find the <server> element of the current instance.
```
OndroMih: This comment is redundant, it should be in pull request description but no need to have in code. If you want to keep some comment, just explain that "Server server = domain.getServerNamed(serverEnv.getInstanceName());" retrieves the current instance's config.

For example:



```suggestion
        // find the <server> element of the current instance. Needed if executed on non-DAS instances,
        // e.g. when calling admin commands against a remote instance
```

OndroMih (changes requested):

OndroMih (approved):

PR #26169, opened 2026-08-07
Fix download instance log file

JackienStephan:
Cannot download log file from ssh instance.
A `ProviderMismatchException` thrown when `logFileDirectoryOnServer.toPath().resolve(loggingFile.getFileName())`, because a `SFTPPath` is not a `UnixPath` or `WindowsPath`.

And also, `SFTPClient.download(..)` requires that local file does not exist, so the local log file needs to be deleted before download a new one.


dmatej (approved):

OndroMih:
Jenkins tests fail because of intra issues, no need to test this PR on Jenkins and block this PR, I'm merging it.

OndroMih (approved):

PR #26170, opened 2026-08-07
Before doing the node-agent element migration, check if the nodes element already exists

JackienStephan:
After doing `start-domain --upgrade`, all the nodes are removed.
I think `createDefaultNodeList()` should not be done when `nodes` exists and `node-agents` not exist.

### DIFF
--- a/nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/NodeAgentConfigUpgrade.java
+++ b/nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/NodeAgentConfigUpgrade.java
@@ -59,6 +59,29 @@ public class NodeAgentConfigUpgrade implements ConfigurationUpgrade, PostConstru
 
     public void postConstruct() {
 
+        // first, check if the nodes element already exists.
+        final Nodes nodes = domain.getNodes();
+        if (nodes != null) {
+            if (!nodes.getNode().isEmpty()) {
+                // if nodes element exists, and it's not empty, there's no need to do the upgrade.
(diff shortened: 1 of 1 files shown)


nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/NodeAgentConfigUpgrade.java
```diff
@@ -59,6 +59,29 @@ public class NodeAgentConfigUpgrade implements ConfigurationUpgrade, PostConstru
...
+            try {
+                ConfigSupport.apply(new SingleConfigCode<Domain>() {
+                    @Override
+                    public Object run(Domain d) throws PropertyVetoException, TransactionFailure {
+                        d.setNodes(null);
+                        return null;
+                    }
+                }, domain);
+            } catch (Exception e) {
+                Logger.getAnonymousLogger().log(Level.SEVERE,
```
dmatej, line 79: Wouldn't be better to move the log to the message of the runtime exception?

JackienStephan: I just want to follow the same style of the other try-catch of `ConfigSupport.apply(..)` in this soruce code.

OndroMih (approved):

OndroMih:
All the tests on Jenkins passed except for a flaky test  ejb_group_3.schedule_on_ejb_timeout, which failed because the timeout was missed by a few milliseconds.

I'm merging this.

PR #26179, opened 2026-08-11
Missing ".level" for logger "org.glassfish.main.jdke" in the logging.properties

JackienStephan:
Causing error when saving Logger Settings using Administration Console.


OndroMih (approved):

OndroMih:
Thanks, @JackienStephan !

PR #26184, opened 2026-08-12
Add Zip file Extra field headers: Header ID & Data Size.

JackienStephan:
The improved validation of the ZIP64 Extra Fields (JDK-8302483) mentioned that each Extra Header Must consist of `Header ID` (2 bytes) and `Data Size` (2 bytes). (from the comments in `java.util.zip.ZipFile`)
Such header is absent in the current extra data bytes, causing "Invalid CEN header" errors when reading zip files exported by `ZipPayloadImpl.java`.

For example, `export-sync-bundle`, then `import-sync-bundle` the exported zip file.
The import command fails if system property `jdk.util.zip.disableZip64ExtraFieldValidation` not set to `true`.
And if using `zipinfo -v` to inspect the exported zip file, errors about the extra fields would also be found.

Using `0xFFFF` as the `Header ID` to avoid any [Known Types of Extra fields](https://libzip.org/specifications/extrafld.txt).
If a better value is suggested, please change it directly.

### DIFF
--- a/nucleus/common/common-util/src/main/java/org/glassfish/admin/payload/ZipPayloadImpl.java
+++ b/nucleus/common/common-util/src/main/java/org/glassfish/admin/payload/ZipPayloadImpl.java
@@ -22,6 +22,9 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.util.Arrays;
 import java.util.Iterator;
 import java.util.Properties;
 import java.util.zip.ZipEntry;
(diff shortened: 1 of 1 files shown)


dmatej (approved):

nucleus/common/common-util/src/main/java/org/glassfish/admin/payload/ZipPayloadImpl.java
```diff
@@ -315,14 +318,18 @@ public InputStream getInputStream() {
...
         private String contentType;
         private Properties props;
 
         private Extra(final byte[] extra) {
             try {
                 props = new Properties();
-                ByteArrayInputStream bais = new ByteArrayInputStream(extra);
+                // skip the header (first 4 bytes).
+                byte[] extraData = Arrays.copyOfRange(extra, EXT_HEADER_SIZE, extra.length);
+                ByteArrayInputStream bais = new ByteArrayInputStream(extraData);
```
OndroMih: Wouldn't this work without copying the array?

```suggestion
                ByteArrayInputStream bais = new ByteArrayInputStream(extraData, EXT_HEADER_SIZE, extra.length);
```

JackienStephan: Thank you for the suggestion. Applied it.

OndroMih (approved):

OndroMih:
Jenkins previously passed, the last run also passed all tests but failed due to a Jenkins hiccup in collecting test reports. I'm merging this.

Issue #24842, opened 2024-03-06
Embedded glassfish JAR doesn't contain JARS/RARS to support JMS


Blavo:
Embedded Glassfish JAR doesn't contain the JARs / RARs to support JMS. On attempted use embedded glassfish excepts saying unable to create an active RAR with a cause of ClassNotFoundException naming com.sun.messaging.jms.ra.ResourceAdapter. In standard glassfish this class is in a JAR inside the imqjmsra.rar in the MQ directory. The (contents of) this RAR are not present in the 7.0.12 embedded glassfish JAR.

If, as I understand, the glassfish admin facility uses JMS then this takes out that facility as well.

## Environment Details
Glassfish 7.0.12

java version "17.0.7" 2023-04-18 LTS
Java(TM) SE Runtime Environment (build 17.0.7+8-LTS-224)
Java HotSpot(TM) 64-Bit Server VM (build 17.0.7+8-LTS-224, mixed mode, sharing)

Mac OS 12.4 (Monterey)

Database not relevant

----------

## Problem Description
As above.

Reliably reproducible

[StackTrace.txt](https://github.com/eclipse-ee4j/glassfish/files/14503102/StackTrace.txt)

## Steps to reproduce
1. Start an instance of glassfish embedded 
2. Note the JMS service port in the domain.xml in the following line:
      <system-property name="JMS_PROVIDER_PORT" description="Port Number that JMS Service will listen for remote clients connection." value="7454"></system-property>
3. In a browser enter http://localhost:nnnn where nnnn is the noted port number; this will trigger JMS (it seems to use lazy initialisation)
4. Glassfish will except as above

## Impact of Issue
I don't use JMS so this doesn't affect me directly. But I believe the glassfish admin facility uses JMS; the admin facility does not work for me against an embedded glassfish and this is probably the cause.

dmatej:
That is quite good question, because embedded glassfish is quite specific. Years ago it was stripped from GF3. Naturally I don't think embedded glassfish should contain IMQ broker+JMS server, but throwing CNF exceptions is quite ugly. So it should support it OR it should nicely tell the user that JMS is not supported, but then we should document an example how to work with an external JMS with Embedded GlassFish.

That is quite a lot of work, I think ...

OndroMih:
[...]

PR #26187, opened 2026-08-15
Fix JMS support in glassfish-embedded-all and add embedded JMS test

hantsy:
## Summary

Fixes JMS support in the embedded GlassFish uber jar (`glassfish-embedded-all`). The JMS resource adapter could not run in embedded mode: the RA bean was loaded from the wrong classloader (`MQJMSRA_AS4001`), the broker stayed in REMOTE mode, the default JMS connection factory was never registered in JNDI, the embedded `jmsra.rar` was missing OpenMQ runtime jars, and — once the embedded broker was functional — shutdown failed with `NoClassDefFoundError: BrokerStateHandler$ShutdownRunnable` because the RAR classloader was closed before the resource adapters were stopped.

## Changes

- **ActiveRAFactory** (connectors-runtime): load the RA bean through the provided RAR classloader for all process types, falling back to the thread context classloader only when no loader is available. Resolves `MQJMSRA_AS4001` classloader mismatch.
- **ActiveJmsResourceAdapter** (jms-core): treat embedded as server for RA configuration so the embedded broker lifecycle (EMBEDDED/LOCAL/DIRECT) is applied instead of hardcoded REMOTE.
- **embedded domain.xml** (kernel): add the default JMS connector-connection-pool, connector-resource `jms/__defaultConnectionFactory`, server resource-ref (mirroring `jms-module-conf.xml`, which is only applied at domain creation in the full distribution), plus system properties `com.sun.enterprise.connectors.system.RevertToEmbedded=true` and `imq.jmsra.direct=false`. This makes ResourceManager bind the connection factory in JNDI in embedded mode.
- **embedded-all packaging** (`pom.xml`): unpack the OpenMQ broker properties (`**/props/broker/**`) from `mq-distribution` with `excludeTransitive=false` so the broker props reach the RAR and the embedded broker starts cleanly. `imqutil.jar` is not bundled in `jmsra.rar` — every class it provides already exists in the other bundled jars, so it is redundant.
- **ConnectorsClassLoaderUtil** (connectors-internal-api): in embedded mode, stop the active resource adapters during `PREPARE_SHUTDOWN` while the RAR classloader is still open (so the embedded broker's lazy class resolution such as `BrokerStateHandler$ShutdownRunnable` succeeds), and close the RAR classloaders on `SERVER_SHUTDOWN` to release the jar handles. The later ResourceManager shutdown re-runs `stopAllActiveResourceAdapters()`, which is idempotent. Embedded-only.
- **appserver/tests/embedded/jms**: a flat embedded test module that deploys a JAX-RS + JMS app (sender, MDB consumer, destination definition) on embedded GlassFish and verifies the send/receive round trip via the default connection factory, including a clean shutdown.

## Verification

Embedded integration test (`HelloJmsIT` against the rebuilt `glassfish-embedded-all`): `Tests run: 1, Failures: 0, Errors: 0, Skipped: 0`. Log shows `broker is EMBEDDED, connection mode is TCP` and `Started:EMBEDDED`, the send/receive round trip works (`sending message from HelloSender` → `received message: Hello JMS`), and shutdown is clean — the RA stop completes (`MQJMSRA_RA1101 ... stopping...` → `stopped.`) with no `BrokerStateHandler` / `ShutdownRunnable` / `NoClassDefFoundError` / `ClassNotFoundException` / `Could not close the`.

Embedded mode (minimal JMS-only Jakarta EE 11 app, `mvn clean package embedded-glassfish:run -Pglassfish-embedded`):

- No `MQJMSRA_AS4001`
- No `NameNotFoundException: jms` — `jms/__defaultConnectionFactory` resolves
- `GET /api/hellojms` returns 200 `sent`, MDB logs `received message: Hello JMS at ...`

Normal distribution regression (Arquillian managed test against the stock unpatched `glassfish-8.0.4.zip`): `HelloJmsResourceTest` passes — `Tests run: 1, Failures: 0, Errors: 0, Skipped: 0`. The patch does not affect the full distribution (the classloader/RA-stop change is embedded-only).

Rebased on upstream `main` and squashed into a single commit.

Related upstream issue: eclipse-ee4j/glassfish#24842

Original PR and discussion: https://github.com/hantsy/glassfish/pull/1

### DIFF
--- a/appserver/tests/embedded/jms/src/test/java/org/glassfish/tests/embedded/jms/HelloJmsIT.java
+++ b/appserver/tests/embedded/jms/src/test/java/org/glassfish/tests/embedded/jms/HelloJmsIT.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+
+package org.glassfish.tests.embedded.jms;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.net.URI;
+import java.net.URL;
+import java.net.URLConnection;
+import java.nio.file.Path;
+
+import org.glassfish.embeddable.Deployer;
+import org.glassfish.embeddable.GlassFish;
+import org.glassfish.embeddable.GlassFishProperties;
+import org.glassfish.embeddable.GlassFishRuntime;
+import org.glassfish.embeddable.archive.ScatteredArchive;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Test JMS functionality in embedded GlassFish.
+ */
+public class HelloJmsIT {
+
+    private static final Path PROJECT_DIR = detectBasedir();
+    private static final int HTTP_PORT = 8080;
+    private static final String APP_NAME = "hellojms";
+
+    static GlassFish glassfish;
+    static String appName;
+
+    @BeforeAll
+    static void setupServer() throws Exception {
+        GlassFishProperties props = new GlassFishProperties();
+        props.setPort("http-listener", HTTP_PORT);
+        glassfish = GlassFishRuntime.bootstrap().newGlassFish(props);
+        glassfish.start();
+
+        ScatteredArchive sa = new ScatteredArchive(APP_NAME, ScatteredArchive.Type.WAR,
+                PROJECT_DIR.resolve(Path.of("src", "main", "webapp")).toFile());
+        sa.addClassPath(PROJECT_DIR.resolve(Path.of("target", "classes")).toFile());
+        URI warURI = sa.toURI();
+
+        Deployer deployer = glassfish.getDeployer();
+        appName = deployer.deploy(warURI);
+    }
+
+    @Test
+    void testJmsSendAndReceive() throws Exception {
+        URL sendUrl = URI.create("http://localhost:" + HTTP_PORT + "/" + APP_NAME + "/api/hellojms").toURL();
+        String response = read(sendUrl);
+        assertEquals("sent", response);
+
+        // The MDB consumes the message asynchronously, poll until it shows up.
+        URL messagesUrl = URI.create("http://localhost:" + HTTP_PORT + "/" + APP_NAME + "/api/hellojms/messages").toURL();
+        String messages = "";
+        long deadline = System.currentTimeMillis() + 30_000;
+        while (System.currentTimeMillis() < deadline) {
+            try {
+                messages = read(messagesUrl);
+            } catch (IOException e) {
+                // Message not available yet, retry.
+            }
+            if (messages != null && messages.contains("Hello JMS")) {
+                break;
+            }
+            Thread.sleep(500);
+        }
+        assertTrue(messages != null && messages.contains("Hello JMS"), "MDB did not receive the message: " + messages);
+    }
+
+    @AfterAll
+    static void shutdownServer() throws Exception {
+        if (appName != null) {
+            glassfish.getDeployer().undeploy(appName);
+        }
+        if (glassfish != null) {
+            glassfish.dispose();
+        }
+    }
+
+    private static String read(URL url) throws IOException {
+        URLConnection connection = url.openConnection();
+        try (BufferedReader reader = new BufferedReader(new InputStreamReader(connection.getInputStream()))) {
+            return reader.readLine();
+        }
+    }
+
+    private static Path detectBasedir() {
+        // Maven would set this property.
+        final String basedir = System.getProperty("basedir");
+        if (basedir != null) {
+            return new File(basedir).toPath().toAbsolutePath();
+        }
+        // Maybe we are standing in the basedir.
+        final File target = new File("target");
+        if (target.exists()) {
+            return target.toPath().toAbsolutePath().getParent();
+        }
+        // Eclipse IDE sometimes uses target as the current dir.
+        return new File(".").toPath().toAbsolutePath().getParent();
+    }
+}
--- a/appserver/connectors/connectors-internal-api/src/main/java/com/sun/appserv/connectors/internal/api/ConnectorsClassLoaderUtil.java
+++ b/appserver/connectors/connectors-internal-api/src/main/java/com/sun/appserv/connectors/internal/api/ConnectorsClassLoaderUtil.java
@@ -21,6 +21,7 @@ import com.sun.enterprise.loader.ASURLClassLoader;
 import com.sun.logging.LogDomains;
 
 import jakarta.inject.Inject;
+import jakarta.inject.Provider;
 import jakarta.inject.Singleton;
 
 import java.io.File;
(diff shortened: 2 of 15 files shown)


appserver/tests/embedded/jms/src/test/java/org/glassfish/tests/embedded/jms/HelloJmsIT.java
```diff
@@ -0,0 +1,123 @@
...
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+
+package org.glassfish.tests.embedded.jms.hellojms;
```
OndroMih: This package is not aligned with the directory, which is just `org/glassfish/tests/embedded/jms`. How is it possible that it compiles and the test is executed?

pzygielo: > How is it possible that it compiles 

Perhaps because compiler does not check the file location but its content. And places `*.class` in proper directory.

> and the test is executed

Perhaps because framework does not work with sources but with binaries and looks for classes with specific names.

https://checkstyle.org/checks/coding/packagedeclaration.html#PackageDeclaration - with its optional [matchDirectoryStructure](https://checkstyle.org/checks/coding/packagedeclaration.html#matchDirectoryStructure) property.

pzygielo: Interestingly - the check is in config:
https://github.com/eclipse-ee4j/glassfish/blob/195eab2126396d9cd3d53486f3ce82dc1ce20f5f/qa/src/main/resources/org/glassfish/qa/config/checkstyle/checkstyle.xml#L120

:thinking:

pzygielo: Locally I'm getting:
```
[INFO] There are 7 errors reported by Checkstyle 13.9.0 with org/glassfish/qa/config/checkstyle/checkstyle.xml ruleset.
[ERROR] src/main/java/org/glassfish/tests/embedded/jms/HelloHandler.java:[17,1] (coding) PackageDeclaration: Package name is not same as directory.
[ERROR] src/main/java/org/glassfish/tests/embedded/jms/HelloSender.java:[17,1] (coding) PackageDeclaration: Package name is not same as directory.
[ERROR] src/main/java/org/glassfish/tests/embedded/jms/JmsResourcesSetup.java:[17,1] (coding) PackageDeclaration: Package name is not same as directory.
[ERROR] src/main/java/org/glassfish/tests/embedded/jms/HelloJmsResource.java:[17,1] (coding) PackageDeclaration: Package name is not same as directory.
[ERROR] src/main/java/org/glassfish/tests/embedded/jms/HelloConsumer.java:[17,1] (coding) PackageDeclaration: Package name is not same as directory.
[ERROR] src/main/java/org/glassfish/tests/embedded/jms/RestActivator.java:[17,1] (coding) PackageDeclaration: Package name is not same as directory.
[ERROR] src/test/java/org/glassfish/tests/embedded/jms/HelloJmsIT.java:[17,1] (coding) PackageDeclaration: Package name is not same as directory.
```

OndroMih: Yes, I thought it's an error from the compiler but it's from checkstyle. I thought that the compiler would give an error too but it seems it doesn't. That's why compilation and tests run in Github. Jenkins job failed way earlier on some infra issue but it would likely fail on the checkstyle error if it got to it.

appserver/extras/embedded/all/src/assembly/jmsra.xml
```diff
@@ -22,5 +22,19 @@
...
+            Bundle the OpenMQ runtime jars inside the RAR so the embedded JMS
+            Resource Adapter and the in-JVM broker load all their dependencies from
+            the RAR classloader (mirrors the full distribution's mq/lib classpath).
+            Start with the minimum missing jar (imqutil.jar); add more only if the
+            embedded broker reports further missing classes.
+        -->
+        <fileSet>
+            <directory>target/jars/mq/lib</directory>
+            <includes>
+                <include>imqutil.jar</include>
```
OndroMih: I'm not sure this is required. Full GlassFish distribution doesn't add imqutil.jar to the RAR and if I remove it and build locally, the HelloJmsIT test still passes in the same way.

The test HelloJmsIT passes with errors during showdown with or without imqutils.jar in the RAR:

```
14:39:43.929232  SEVERE            Thread-35                                                             . RAR6037 : Resource adapter threw an exception while stopping
java.lang.NoClassDefFoundError: com/sun/messaging/jmq/jmsserver/BrokerStateHandler$ShutdownRunnable
	at Connector(jmsra)//com.sun.messaging.jmq.jmsserver.BrokerStateHandler.initiateShutdown(BrokerStateHandler.java:279)
```

OndroMih: The not-found class `BrokerStateHandler` is in imqjmsra.jar, which is already in the RAR, in `lib/install/applications/jmsra`. So it wouldn't help to copy that jar there.

hantsy: > BrokerStateHandler 

I will check. Maybe it requires more libs. I am not sure.

appserver/tests/embedded/jms/src/test/java/org/glassfish/tests/embedded/jms/HelloJmsIT.java
hantsy: I noticed the package structure issue and refactored it locally, but I forgot to sync the git changes.

appserver/extras/embedded/all/src/assembly/jmsra.xml
hantsy: <del>I tried running HelloJmsIT on my local system today, and I didn't find any such error.</del> 

I used AI to output the full server log and found the error.

appserver/tests/embedded/jms/src/test/java/org/glassfish/tests/embedded/jms/HelloJmsIT.java
hantsy: @OndroMih @pzygielo Updated

appserver/extras/embedded/all/src/assembly/jmsra.xml
hantsy: @OndroMih Added a fix for the NoClassDef exception, which is caused by another class loader issue.

hantsy: This change seems not ideal. But I have no better idea to fix, the Rar class loader/resources was destroyed too early.

OndroMih:
All looks good to me. 

I just want to note that this includes OpenMQ broker server into the Embedded GlassFish JAR. It adds about 1MB to the size of the JAR - before this it's 92321371 bytes, after this it would be 93724821 bytes. The value of an embedded broker is only in development and testing mostly. In production, it would be useful if can cluster multiple embedded broker instances together, so that they can exchange messages. However, I don't think this is easy to configure. Embedded GlassFish doesn't support clustering and wouldn't configure OpenMQ cluster out of the box, some manual configuration would be needed and maybe even not possible with the embedded broker.

I would wait until a few more people review this as this can have a big impact for a long time, with potentially little value in production scenarios.

I would rather have 2 PRs - one fixing the REMOTE mode (classloader fixes, etc.), which I would happily accept, and another one adding OpenMQ server into Embedded GlassFish and support EMBEDDED and LOCAL modes, which I would like to have a longer time to think about it, and possibly make sure that it's possible and easy to form a cluster of OpenMQ instances.

hantsy:
> I would rather have 2 PRs - one fixing the REMOTE mode (classloader fixes, etc.), which I would happily accept, and another one adding OpenMQ server into Embedded GlassFish and support EMBEDDED and LOCAL modes, which I would like to have a longer time to think about it, and possibly make sure that it's possible and easy to form a cluster of OpenMQ instances.

I am not sure all the changes are the best solutions. Claude Code mainly fixed them based on my instructions. I hope the experts can review them and provide feedback. 

For the pr, there are 3 fixes: one is about the classloader, the second is the naming registration, and the third is applying the remote mode defined in the embedded domain.xml. Together, these fixes can resolve the JMS issue.

OndroMih:
As I wrote, this is good enough if we want to have OpenMQ bundled inside GlassFish JAR. It makes the JAR 1MB bigger but it probably doesn’t matter.

I think we can accept this and improve it later.

OndroMih (approved):

Issue #24060, opened 2022-08-02
Remove obsoleted logging-related attributes from domain.xml


dmatej:
I believe ie. server.http-service.virtual-server.__asadmin.log-file=${com.sun.aas.instanceRoot}/logs/server.log came from SunOne, where admin had own JVM and other instances had own instances too. But somehow it survived to these days.

There were also some XML-properties mappings, which should be checked if they are or they are not still meaningful, usable and useful.

PR #26192, opened 2026-08-23
Remove obsolete log-service config from new domains

renatsaf:
Logging has been configured through `logging.properties` since v3, but `domain.xml` still carries leftovers from the `<log-service>` era. This is the uncontroversial part of #24060: nothing user-facing is removed, and the config beans stay so that existing `domain.xml` files still parse and the v2 upgrade path keeps working.

### Stop writing `<log-service>` into new domains

Nothing reads the element at runtime any more:

- `MiniXmlParser.parseConfig()` has no `log-service` branch; the launcher resolves the server log from the `GlassFishLogHandler` output file property in `logging.properties`.
- `WebContainer` takes the server log path from `ServerLogFileManager.getCurrentLogFile()`.
- The only remaining `Config.getLogService()` callers are `Config.addIndex()`, which tolerates `null`, and the two upgrade services.

Yet both `domain.xml` templates wrote it into `default-config`, advertising a `log-rotation-limit-in-bytes` that has no effect — while `UpgradeLogging` exists precisely to delete the element from domains being upgraded. `server-config` has not carried it for a while; `default-config` now matches.

### Fix the legacy name translation

`LoggingXMLNames.xmltoPropsMap` translates the v2 `domain.xml` names that `set-log-levels` and `set-log-attributes` still accept. Two defects:

`LogDomains.DOMAIN_ROOT` is a logger name *prefix* ending with a dot, so concatenating it with `".level"` produced `jakarta..level` — a property for a logger literally named `jakarta.`. `asadmin set-log-levels root=FINE` configured nothing. It now maps to `jakarta.level`, the key the shipped `logging.properties` already uses. Same family as #26179.

`LoggingConfigImpl` translated a key read from `logging.properties` and then looked the value up under the *translated* name, which by definition is not the name present in the file. Any legacy key surviving in a file yielded a `null` value from `getLoggingProperties()` and `getDefaultLoggingProperties()`, and made `getLoggingFileDetails()` return `null` instead of the configured `server.log` path.

### Stop the upgrade from polluting `logging.properties`

`UpgradeLogging` pushed every `log-service` attribute and every module log level through `LoggingConfigImpl`, which writes any name it cannot translate verbatim. A domain upgraded from v2 gained about a dozen meaningless lines: `alarms`, `use-system-logging`, `log-filter`, `log-to-console`, `retain-error-statistics-for-hours`, and the module levels dropped after v2 (`verifier`, `connector`, `synchronization`, `node-agent`, `self-management`, `group-management-services`, `management-event`) — all of which default to a non-null value. Only settings with a counterpart are migrated now.

`log-handler` named a handler *added* to the chain, but the key it translates to is the complete list of root handlers, so migrating it replaced them — dropping the `GlassFishLogHandler` that writes the server log. It is appended instead.

A `<log-service>` without a `<module-log-levels>` child threw an NPE out of the upgrade, as did one without a `file` attribute; both are optional.

### Virtual server without `log-file`

`VirtualServer.getLogFile()` declared a default of `${com.sun.aas.instanceRoot}/logs/server.log`, so the config bean reported a log file for every virtual server even though no `domain.xml` has carried the attribute for years. `WebContainer` feeds that into `VirtualServer.setLogFile()`, which starts a virtual-server-specific `FileLoggerHandler` whenever the value differs from the server log.

The two agree in a default domain, so nothing happened there. Relocate the server log, though:

```
asadmin set-log-attributes org.glassfish.main.jul.handler.GlassFishLogHandler.file=/var/log/glassfish/server.log
```

and the phantom default no longer matches — the web container opens a second handler writing into `${com.sun.aas.instanceRoot}/logs/server.log`, the file the administrator just moved away from. This is derived from the code path, not reproduced on a running server.

The attribute is optional now, and `setLogFile()` compares against the server log path when it is not set. A virtual server without `log-file` stays attached to the server log wherever that log currently is; one that sets `log-file` behaves as before.

### Not in this PR

Left for a follow-up, since they need a maintainer decision:

- Deleting the `LogService` / `ModuleLogLevels` config beans, `NodeAgent.getLogService()`, `DefaultConfigUpgrade.LogServiceConfigCode` and `UpgradeLogging` itself. That breaks `asadmin get/set` on `configs.config.*.log-service.*` and drops the v2 to v3 logging migration.
- Removing `virtual-server@log-file` outright — attribute, admin console field, and the `devtests/web/virtualServerLogFile` devtest. That is a feature removal, not a cleanup.
- `DefaultConfigUpgrade` creates `<log-service>` while `UpgradeLogging` removes it, and `DomainXml.upgrade()` iterates the upgrade services without explicit ordering.

### Testing

`LoggingXMLNamesTest` is new and fails on `main` (`expected: <jakarta.level> but was: <jakarta..level>`). Green locally: `common-util` 108, `config-api` 89, `server-mgmt` 73, `launcher` 9, `web-glue` 2. Both templates verified well-formed.

Refs #24060

### DIFF
--- a/appserver/admin/template/src/main/resources/config/domain.xml
+++ b/appserver/admin/template/src/main/resources/config/domain.xml
@@ -302,9 +302,6 @@
          <jms-service type="EMBEDDED" default-jms-host="default_JMS_host" addresslist-behavior="priority">
              <jms-host name="default_JMS_host" host="localhost" port="${JMS_PROVIDER_PORT}" admin-user-name="admin" admin-password="admin" lazy-init="true"/>
          </jms-service>
-         <log-service log-rotation-limit-in-bytes="2000000" file="${com.sun.aas.instanceRoot}/logs/server.log">
-             <module-log-levels />
-         </log-service>
          <security-service>
              <auth-realm classname="com.sun.enterprise.security.auth.realm.file.FileRealm" name="admin-realm">
                  <property name="file" value="${com.sun.aas.instanceRoot}/config/admin-keyfile" />
--- a/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/VirtualServer.java
+++ b/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/VirtualServer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2024 Contributors to the Eclipse Foundation.
+ * Copyright (c) 2022, 2026 Contributors to the Eclipse Foundation.
  * Copyright (c) 1997, 2018 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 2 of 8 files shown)


OndroMih:
There are too many AI-generated PRs right now, I'm closing this temporarily until there's more throughput to review them.

PR #26193, opened 2026-08-26
Fix connection pool validation if at most once > 0 and other connection pool fixes

### LINKED ISSUE #25930: JDBC Connection Pool validation doesn't work if "Validate At Most Once" is set to >0

OndroMih:
Fixes #25930:

* when connection was invalid and not validated, new transaction failed and kept the connection associated wit ha transaction, which then prevented any further recreation of the connection during validation

Additional fixes:

* when validation was for each request (atMostOnce=0) and it failed, new connection was recreated in a busy state, throwing an error later
* when connections to the database are secured by a user and password, any change in configuration of the pool resets the pool but it didn't reset the connection factory assigned to the cached credentials. Any change in connnection pool without a restart lead to failure to establish new connections, reporting that the credentials don't match the connection factory

All scenarios covered by new IT tests.

I also added support for glassfish.debug system property which, if true, runs the GlassFish Docker container with debug port exposed and GlassFish inside in debug mode. It's then possible to debug Glassfish running in the Docker container. Testcontainers randomize the ports so each time it's needed to query Docker to find out the port exposed to the host machine and attach the debugger to it.

### DIFF
--- a/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/connectors/ConnectionManagerImpl.java
+++ b/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/connectors/ConnectionManagerImpl.java
@@ -24,7 +24,6 @@ import com.sun.appserv.connectors.internal.api.PoolingException;
 import com.sun.appserv.connectors.internal.spi.ConnectionManager;
 import com.sun.enterprise.config.serverbeans.BindableResource;
 import com.sun.enterprise.connectors.authentication.AuthenticationService;
-import com.sun.enterprise.connectors.util.ConnectionPoolObjectsUtils;
 import com.sun.enterprise.connectors.util.ResourcesUtil;
 import com.sun.enterprise.deployment.ConnectorDescriptor;
 import com.sun.enterprise.deployment.ResourcePrincipalDescriptor;
@@ -123,7 +122,7 @@ public class ConnectionManagerImpl implements ConnectionManager, Serializable {
      * functionality might be achieved.
      */
     @Override
-    public Object allocateNonTxConnection(ManagedConnectionFactory mcf, ConnectionRequestInfo connectionRequestInfo)
+    public Object allocateNonTxConnection(ManagedConnectionFactory managedConnectionfactory, ConnectionRequestInfo connectionRequestInfo)
         throws ResourceException {
         LOG.finest("Allocating NonTxConnection");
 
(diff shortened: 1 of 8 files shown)


appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/connectors/ConnectionManagerImpl.java
```diff
@@ -255,10 +255,12 @@ protected Object internalGetConnection(ManagedConnectionFactory mcf,
             ResourceSpec resourceSpec = new ResourceSpec(jndiNameToUse, ResourceSpec.JNDI_NAME, poolMetaData);
             resourceSpec.setPoolInfo(this.poolInfo);
             ManagedConnectionFactory freshManagedConnectionFactory = poolMetaData.getMCF();
+            boolean managedConnectionFactoryChanged = !freshManagedConnectionFactory.equals(mcf);
 
-            if (!freshManagedConnectionFactory.equals(mcf)) {
+            if (managedConnectionFactoryChanged) {
                 LOG.info("conmgr.mcf_not_equal");
             }
+            ManagedConnectionFactory effectiveManagedConnectionFactory = freshManagedConnectionFactory;
```
pzygielo: Is `effectiveManagedConnectionFactory` always `freshManagedConnectionFactory` then?
Is `mcf` used only to determine if to log "conmgr.mcf_not_equal"?
If so - could we rename `freshManagedConnectionFactory` for example to `managedConnectionFactoryFromPool`, drop [new one] `effectiveManagedConnectionFactory` and use `managedConnectionFactoryFromPool` from here?

🤔

OndroMih: You're right. I removed usage of the original factory completely. The code just uses the factory from the pool. No need to compare the factories, it was only to log a message but that message is not needed at all.

Before, the factory passed as an argument from the datasource was used, but it didn't match the subject if the factory was recreated by the pool after a config update. However, as the original fctory can be recreated, it's not relevant. The factory retrieved from the pool is always relevant instead.

appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/pool/ConnectionPool.java
```diff
@@ -872,9 +872,9 @@ private ResourceHandle getUnenlistedResource(ResourceAllocator resourceAllocator
 
             // Resource from the pool must be marked busy when it is returned from the pool
             if (resourceHandle != null) {
```
dmatej: resourceHandle or resourceFromPool?

OndroMih: You are right. I also moved the definition of the resourceHandle variable inside the try block - it should be used only as an iteration variable in the while loop.

arjantijms (approved):

PR #26195, opened 2026-08-29
Rewritten Jenkinsfile - based on Jenkinsfile in 9.0 branch

dmatej:
### Original State

* Most jobs used base docker image from CBI team.
* Jobs were split to groups, but all was running in parallel
* Since June we had issues with Jenkins reliability, see https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/work_items/7610 ; issues sometimes went away, we reduced size of the history, tried to optimize Jenkinsfile, it always helped for a while but it broke again and again.
* The build in June took around 25 minutes. The nearly same build in September took around 40 minutes (depends on ordering of stages, which is decided by Jenkins, but it is always much slower than before).

### Story

* So finally @arjantijms lost his patience and found workarounds to avoid starving and failing builds in #26190 and #26191 . It smells by something what should not be necessary, because I believe it duplicates and hacks into meaning of life of Kubernetes, but it works. He also added some monitoring of peak memory usage including IO buffers. So I used most of it and just avoided the K8S stuff.

Then I started thinking ... why we create podTemplate for each job? Could it be possible to make it more simple? I also noticed that order of some configuration lines differ. And replacing numbers by different numbers confused me several times. After some maintenance and cleanup I took a paper and used some bash script to filter peak memory usages of every stage. Then I separated them to create configurations for pod templates... 

Finally I started tuning mem and cpu requests to find the best combination. We could probably try some simulations using measured data...

### Changes

* Current times are around 35 - 39 minutes, depends on ordering of stages, hardware chosen for nodes by k8s and also on the fact that the real hardware can be shared with other projects on the background.
* Refactored Pod/Node/PodTemplate configuration to make it more maintainable and logical.
  * The main agent is `none` now, so it doesn't allocate resource. That also mean it cannot run much. It is used just to orchestrate stages.
  * Actually, it also runs the first stage StopOld, so when the system is overloaded, this is still executed - so when you push to the same PR, this stage interrupts previous running builds of the same PR.
  * yaml configurations have now a template. I could also create a global podTemplate directly in the groovy script, but the syntax would look worse than this work with strings.
  * The only thing which changes in the template, is cpu and memory request for k8s.
  * Each pod template now uses own label and name, Jenkins uses labels to avoid starvation - if there is a capacity for small pods, they can run, while larger still wait. Yes, that means those large can starve for some time.
  * All pods now use the same docker image as a base, the Maven-Temurin docker image. That resolved also few older problems with PATH in Ant pod.
  * Instead of Ant and Maven container we have just the Action container as they were principialy the same.
  * Finally we have 5 yaml configurations and respectively 5 different pod definitions, each has own label. The usage of each label can be limited, but the limit is currently 20, which is much higher than CBI limit of 10 concurrent nodes.
  * Configuration templates are based on the cgroup mem peak usage monitoring as it was implemented in #26191
* G1GC for maven replaced by ShenandoahGC. On tight environments it has an advantage of returning unused memory to the system. I use it also with Eclipse IDE.
* GIT clone is not executed 5 times, but just once. Where needed, we stash/unstash it.
* 469 lines in main, 1066 in 9.0, 657 lines here.

### What limits we really have?

```
// Total GF project limit as of September 2026 is 13 sponsored resource packs (+1 is free), which means:
// cpu limit is 14 * 2000 = 28 000
// mem limit is 14 * 8 Gi = 112 Gi

// But ... these lines are collected from logs:
// maximum cpu usage per Pod is 8300m
// maximum cpu usage per Container is 8
// maximum cpu usage is 44800m
```
I guess I computed it wrong, so somebody explain me, please.

### Hardware Influence

Every stage can run on different hardware, these results are based just on our logs.

```
| CPU        | AMD EPYC 9255       | Intel Xeon Silver 4210R |
| CPU        | 96 threads <4.3 GHz | 40 threads <3.2 GHz     |
| BogoMIPS   | 6400                | 4800                    |
| Build      | ?                   | 7:31 - 8:12)            |
| Main Tests | 7:39                | 13:43 - 14:30           | (just mvn clean verify)
```

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,72 +15,200 @@
 * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
 */
 
+// See:
+// https://hub.docker.com/r/eclipsecbi/jiro-agent-basic-ubuntu/tags
+// https://github.com/jenkinsci/kubernetes-plugin/blob/master/README.md
+// https://kubernetes.io/docs/concepts/workloads/pods/#working-with-pods
+// https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/
+
+// These limits are collected from logs:
+// maximum cpu usage per Pod is 8300m
+// maximum cpu usage per Container is 8
+// maximum cpu usage is 44800m
+
 def mvnVersion = '3.9.16'
 def javaVersion = '21'
-def jdkTool = "temurin-jdk${javaVersion}-latest"
-def mvnTool = "apache-maven-${mvnVersion}"
 
-def mvnContainerCfg = """
+def podYamlTemplate = """
 apiVersion: v1
 kind: Pod
 spec:
+  nodeSelector:
+    kubernetes.io/os: "linux"
   containers:
-  - name: maven
+  - name: jnlp
+    imagePullPolicy: IfNotPresent
+    tty: true
+    workingDir: "/home/jenkins/agent"
+    env:
+    - name: "JENKINS_REMOTING_JAVA_OPTS"
+      value: "-showversion -XshowSettings:vm -Xmx256m -Dorg.jenkinsci.remoting.engine.JnlpProtocol3.disabled=true -Dorg.jenkinsci.plugins.gitclient.CliGitAPIImpl.useSETSID=true"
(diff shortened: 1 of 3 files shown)


arjantijms (approved):
Did not read through every line of code, but the general approach is okay!

PR #26196, opened 2026-09-02
Fix issues when set defaultWebModule for virtual server or delete a virtual server

JackienStephan:
When set `defaultWebModule` using Admin Console, error occurs.

This is because `WebAppHandlers.ensureDefaultWebModule` not only handle with the config currently being edited.
For example, `a-web-app` targeting the clusters or standalone instances referencing configuration `a-config`.
those clusters or standalone instances not referencing `a-config` apparently not having `a-web-app` as `application-ref`, causing failures when updating `virtualServers` attribute on those `application-ref`.

And a wrong target ('domain') is given when calling `DeployUtil.reloadApplication(..)`, this would definitely fail unless a cluster or instance happened to have the same name as `domain`.

Further more, I think it also ought to update the `application-ref` of associated clusters. The original process only handle with instances.

Similarly, the problem (not considering config being edited and clusters) also appears when invoking additional handler (`checkVsOfAppRef`) after delete a virtual server.

### DIFF
--- a/appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/WebAppHandlers.java
+++ b/appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/WebAppHandlers.java
@@ -35,6 +35,7 @@ import com.sun.jsftemplating.annotation.HandlerOutput;
 import com.sun.jsftemplating.layout.descriptors.handler.HandlerContext;
 
 import java.net.URLEncoder;
+import java.nio.file.Files;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


pzygielo:
- https://github.com/eclipse-ee4j/glassfish/actions/runs/33607076586/job/100173398200?pr=26196#step:4:191 - I have seen this step to fail for another time. Wonder, if that is so extremely helpful to be there...

- :arrow_right: https://github.com/eclipse-ee4j/glassfish/issues/26197

OndroMih (approved):

PR #26201, opened 2026-09-07
Add an HTTP transport for remote EJB and JNDI, alongside IIOP

AngeloRubens:
Remote EJB and JNDI are reachable only over RMI-IIOP today. That protocol
predates NAT, container networking and the assumption that 443 is the one port
you can count on, and since JEP 320 removed CORBA from the JDK the client also
has to ship `glassfish-corba` and `pfl` to speak it at all.

This adds a second transport rather than replacing the first. Both can be
enabled at once and nothing in `orb-iiop` changes.

This is the first of three stacked pull requests:

| PR | Adds |
|---|---|
| **this one** | the protocol, the client, container-independent dispatch, and the codecs |
| #26214 | transaction propagation |
| #26215 | the GlassFish adapter: the endpoint on the server's HTTP listeners, the container, naming, realm and transaction bindings |

### Modules

Four, under `appserver/orb`, each packaged as an OSGi bundle like every other
module there:

| module | contents |
|---|---|
| `orb-http-protocol` | the wire format, the codec SPI and codec discovery; no dependency beyond the JDK |
| `orb-http-client` | thin client: JDK `HttpClient` plus dynamic proxies |
| `orb-http-server` | dispatch, behind interfaces so it is testable without a server |
| `orb-http-codec-fory` | an optional codec built on Apache Fory |

### Design

It leans on seams the container already has. `ContainerBridge` is
`EjbContainerFacade` restated without `java.rmi` — `getTargetObject` and
`releaseTargetObject`, whose implementation in `BaseContainer` already carries
the comment *"Called from the ProtocolManager when a remote invocation
arrives."* `NamingBridge` is the nine methods of `SerialContextProvider`. The
EJB key keeps the layout `POARemoteReferenceFactory` documents,
`ejbId(8) + length(4) + key`, so an HTTP invocation and an IIOP invocation name
the same instance.

The wire shape follows the WildFly HTTP Client specification v1 — same paths,
same header semantics, same `JSESSIONID` affinity — so that adding a JBoss
Marshalling codec later would make the two interoperable. The codec token is
in every content type for exactly that reason.

What is supported through that seam:

- stateless, singleton and stateful session beans; a stateful conversation is
  pinned by the affinity cookie and keeps its state across separate requests
- the business view and the EJB 2.x home and component views, for session
  beans (entity beans are not supported, and were pruned from the platform)
- asynchronous methods returning `Future`, including HTTP 202
- the Enterprise Beans exception contract: application exceptions arrive as
  themselves, with their state and stack traces
- JNDI lookup, list, bind and the rest

### Codecs

Java serialization is the default, not JSON. Remote EJB is defined as
pass-by-value with RMI semantics: cyclic graphs, shared references,
application exceptions with their causes and stack traces. Every reader is
given an `ObjectInputFilter`; there is no unfiltered path.

The codec is not fixed, though. Codecs are found through `META-INF/services`
and ranked, so adding one is adding a jar: no import, no property, no
application change. The choice is made per request, not per server — the codec
is read from the request's content type and the reply is written in the same
one — so one endpoint serves `jser` and `fory` clients side by side. A client
whose codec the server cannot read falls back to `jser` and retries once, so a
new codec never breaks a peer that lacks it.

`orb-http-codec-fory` is that mechanism exercised end to end with
[Apache Fory](https://fory.apache.org/). It is configured to preserve what
remote calls already rely on — identity, cycles, `transient` — and each is
pinned by a test. The same `ObjectInputFilter` guards it, and reading and
writing use separate instances so a class resolved on the way out is never
pre-admitted on the way in. Inside GlassFish, Fory's per-type code generation
cannot link across bundle class loaders, so it is off by default and the
reflective path runs; the README records the three arrangements that were
measured.

The Fory module is optional: nothing else depends on it, and it is not added
to any featureset. It is the only part of this work that brings in
third-party code, which needs a dependency review before release:

| Artifact | Version | License |
|---|---|---|
| `org.apache.fory:fory-core` | 1.7.1 | Apache-2.0 |
| `org.codehaus.janino:janino` | 3.1.12 | BSD-3-Clause |
| `org.codehaus.janino:commons-compiler` | 3.1.12 | BSD-3-Clause |

All three are embedded in the bundle, since none publishes OSGi metadata. The
janino entries arrive transitively and are not Apache-2.0.

The README in `orb-http-protocol` has a *Writing a codec* section: the
contract, the obligations the interface cannot enforce (the filter, framing,
separate read and write caches, weak class loader keys), and the bundle
packaging mistakes that build cleanly and fail at runtime, each tied to the
place in the Fory module that handles it.

### HTTP/2

No code. Grizzly already registers `Http2AddOn` from `GenericGrizzlyListener`
and the JDK client negotiates HTTP/2 by default. The work was keeping the
protocol free of trailers, server push and ordering assumptions, so h1/h2 — and
h3 on JDK 26 — stay a deployment choice. `Protocol` documents the two Grizzly
defaults that want raising for this traffic: a 64 KB initial window and 100
concurrent streams.

### Not in this PR

- **Transaction propagation.** Here it is refused explicitly rather than
  ignored: `TxContext` reserves the wire space, and an invocation arriving
  with a transaction attached is rejected instead of being run outside it.
  #26214 implements it on that reserved space.
- **The GlassFish adapter**, binding `ContainerBridge` to the real container
  and `ServerExchange` to Grizzly, is #26215. Keeping the dispatchers behind
  those interfaces is deliberate — these modules compile and test without the
  GlassFish reactor.

### Tests

145, on JDK 21: 38 in the protocol, 37 in the client, 55 in the server and 15
in the Fory codec. The round-trip tests drive a real client proxy through the
real wire format into the real dispatch, over a real socket where it matters,
on both codecs.

With #26214 and #26215 on top, the whole stack is also exercised against a
running GlassFish 8.0.4 on every change: the same bean over IIOP and over HTTP,
both codecs, transactions, and one server calling another over HTTP.

---

## Migrating an existing client

For a session bean reached through its remote business interface, the migration
is the factory and the URL:

```diff
 Properties env = new Properties();
-env.put(Context.INITIAL_CONTEXT_FACTORY, "com.sun.enterprise.naming.SerialInitContextFactory");
-env.put("org.omg.CORBA.ORBInitialHost", "host");
-env.put("org.omg.CORBA.ORBInitialPort", "3700");
+env.put(Context.INITIAL_CONTEXT_FACTORY, "org.glassfish.orb.http.client.HttpInitialContextFactory");
+env.put(Context.PROVIDER_URL, "https://host:8181/glassfish-services");

 InitialContext ctx = new InitialContext(env);
 Greeter greeter = (Greeter) ctx.lookup("java:global/myapp/GreeterBean!com.acme.Greeter");
```

The JNDI name, the interface and the invocation are unchanged. The classpath
loses `gf-client.jar` and the `glassfish-corba` / `pfl` chain behind it, and
gains `orb-http-client`, `orb-http-protocol` and `jakarta.ejb-api`. Credentials
still travel on `Context.SECURITY_PRINCIPAL` / `SECURITY_CREDENTIALS`, as
preemptive HTTP Basic, so over HTTPS; a credential the server rejects stops the
call rather than continuing anonymously.

`LegacyClientMigrationTest` is exactly that code, run against a real socket.
`LegacyEjb2xTest` does the same for a client written against an EJB 2.x home.

**What does not carry over:**

- `PortableRemoteObject.narrow(...)` — delete it. It returns the object
  unchanged when it is already an instance of the target interface, so it is
  harmless while the CORBA classes are still present, but keeping them is what
  this transport exists to avoid.
- **Entity beans** — finders, primary keys, `Handle`, `EJBMetaData`. Session
  beans only.
- **`corbaname:` / `corbaloc:` URLs** — they name an IIOP endpoint. Use the
  portable `java:global/...` name.
- **`com.sun.appserv.iiop.endpoints`** — replaced by ordinary HTTP load
  balancing, with the `JSESSIONID` affinity cookie pinning stateful
  conversations.

## Development

This was written with the help of a generative AI assistant, Claude
(Anthropic), working from the GlassFish and Eclipse ORB sources. The direction
and the design decisions are mine — which container seams to build on, HTTP/2
as a first-class target, Java serialization rather than a JSON binding,
refusing transaction propagation rather than silently ignoring it — and I
reviewed the result; the assistant produced the implementation and the tests.

Sources the design draws on are named in the commit messages and in the
javadoc: the WildFly HTTP Client wire specification v1 for the protocol shape,
GlassFish's own `EjbContainerFacade` / `SerialContextProvider` / 
`POARemoteReferenceFactory` for the server seams and the EJB key layout, and
Payara's Lite Remote EJB as the counter-example that motivated not using a
JSON binding.


OndroMih:
If this works, then all dreams come true :-D

Seriously, this would be really useful to modernize legacy apps, even sometimes for writing new apps in the cloud.

How much is this built with AI, @AngeloRubens ? If you used AI, please add information about that in the pull request description. According to the [Eclipse Foundation handbook](https://www.eclipse.org/projects/handbook/#genai):

> Quote your sources. And, to the extent possible, keep track both in your own records and in any content that you create, the generative AI platforms that you used to create that content and other relevant details.

AngeloRubens:
Thanks — and fair question. Let me answer it directly, then say what has changed since you looked.

**On AI:** yes, substantially. I used Claude (Anthropic) as a coding assistant. The direction and the design decisions are mine — which container seams to build on, HTTP/2 as a first-class target, Java serialization rather than a JSON binding, refusing transaction propagation rather than silently ignoring it — and I reviewed the result; the assistant produced the implementation and the tests. I've added a **Development** section to the description with that, and with the sources, as the handbook asks.

Those sources, briefly. The wire protocol follows the WildFly HTTP Client specification v1 deliberately — same paths, same header semantics, same `JSESSIONID` affinity — so that adding a JBoss Marshalling codec later would make the two interoperable. The server side leans on seams GlassFish already documents as remote-invocation entry points: `EjbContainerFacade.getTargetObject`, whose implementation in `BaseContainer` carries the comment *"Called from the ProtocolManager when a remote invocation arrives"*, and the nine methods of `SerialContextProvider`. The EJB key layout is `POARemoteReferenceFactory`'s rather than an invention, so an HTTP invocation and an IIOP invocation name the same instance. Payara's Lite Remote EJB was the counter-example that motivated not using a JSON binding.

**On "if this works":** that was fair too, and it wasn't tested — only against an in-memory fake, which is not a demonstration. It is now. The second commit puts the dispatchers behind a real socket and drives them through `new InitialContext(env)` and the real `java.net.http.HttpClient`:

- `aLegacyClientChangesOnlyTheFactoryAndTheProviderUrl` — the whole claim in one test: same JNDI name, same interface, same call
- a stateful conversation across separate requests, a 3 MB payload, eight threads sharing one client, an application exception arriving with its business state and cause chain, and credentials arriving as an identity at the seam `SecurityContext.setCurrent` hangs off
- `ObjectGraphFidelityTest` — cycles, shared references keeping their identity rather than becoming two equal copies, `transient`, `writeObject`/`readObject`, enum identity, a declared `Object` parameter carrying its runtime type, null arguments, overload resolution. Each of those is a case a JSON binding fails, and each appears in applications written against IIOP that nobody is going to rewrite.

Writing them found a real bug: the affinity cookie was documented and never actually set. Stateful addressing worked, but in a cluster nothing pinned a conversation to the instance holding it. That commit implements it — `GET {ctx}/common/v1/affinity`, and `open` sets it too — and corrects the javadoc that had overstated it.

**Transactions** are not in this PR; they're on a follow-up branch, [`orb-http-transactions`](https://github.com/AngeloRubens/glassfish/tree/orb-http-transactions). They cost far less than I expected: `JavaEETransactionManagerDelegate` already exposes `recreate(Xid, timeout)`, `release(Xid)` and `getXATerminator()` — the JCA transaction inflow contract, which the server already implements because an inbound resource adapter needs it. An HTTP request carrying a transaction is the same problem, so the server half is a mapping onto primitives the container already has. Two modes, because a thin client cannot be a coordinator: `txn/v1/ut/*` where the server coordinates, and `txn/v1/xa/*` where the caller does and this endpoint is one branch. There is a test running two servers under one coordinator that checks the outcomes which are *not* success — a refused prepare rolling back the branch that had already voted yes, a read-only branch dropping out and never being committed, recovery listing prepared branches.

**Serialization is not baked in.** The codec token appears in every content type and marshalling sits behind an SPI, so a second codec is not a protocol change or a version bump. JBoss Marshalling is the concrete one, for the WildFly interoperability above. Anything that round-trips a Java object graph fits the same SPI, and Apache Fory would be a natural candidate for speed. A schema-first format such as Protobuf would work for applications willing to constrain their remote interfaces to it, but is not a drop-in, since remote EJB is pass-by-value over arbitrary `Serializable` graphs.

The honest limit, also now in the description: none of this reaches a running server yet. The adapter binding `ContainerBridge` to the container and `ServerExchange` to Grizzly isn't written — what's here is the protocol, the client, the dispatch, and the tests that exercise all three. Happy to split the PR if smaller pieces would review better.

dmatej:
```
23:47:22  [INFO] There are 9 errors reported by Checkstyle 13.9.0 with org/glassfish/qa/config/checkstyle/checkstyle.xml ruleset.
23:47:22  [ERROR] src/main/java/org/glassfish/orb/http/protocol/EjbRoutes.java:[19,8] (imports) UnusedImports: Unused import - java.io.UnsupportedEncodingException.
23:47:22  [ERROR] src/main/java/org/glassfish/orb/http/protocol/InvocationEnvelope.java:[23,1] (imports) ImportOrder: Wrong lexicographical order for 'java.io.IOException' import. Should be before 'java.io.InputStream'.
23:47:22  [ERROR] src/test/java/org/glassfish/orb/http/protocol/ChunkedOutputTest.java:[23,1] (imports) ImportOrder: Import 'java.io.ByteArrayOutputStream' violates the configured relative order between static and non-static imports.
23:47:22  [ERROR] src/test/java/org/glassfish/orb/http/protocol/ChunkedOutputTest.java:[29,1] (imports) ImportOrder: 'org.junit.jupiter.api.DisplayName' should be separated from previous imports.
23:47:22  [ERROR] src/test/java/org/glassfish/orb/http/protocol/EjbKeyTest.java:[25,1] (imports) ImportOrder: Import 'java.util.Random' violates the configured relative order between static and non-static imports.
23:47:22  [ERROR] src/test/java/org/glassfish/orb/http/protocol/EjbKeyTest.java:[26,1] (imports) ImportOrder: 'org.junit.jupiter.api.DisplayName' should be separated from previous imports.
23:47:22  [ERROR] src/test/java/org/glassfish/orb/http/protocol/PathScannerTest.java:[25,1] (imports) ImportOrder: Import 'java.nio.charset.StandardCharsets' violates the configured relative order between static and non-static imports.
23:47:22  [ERROR] src/test/java/org/glassfish/orb/http/protocol/PathScannerTest.java:[27,1] (imports) ImportOrder: 'org.junit.jupiter.api.DisplayName' should be separated from previous imports.
23:47:22  [ERROR] src/test/java/org/glassfish/orb/http/protocol/RoutesAndContentTypeTest.java:[26,1] (imports) ImportOrder: Import 'org.junit.jupiter.api.DisplayName' violates the configured relative order between static and non-static imports.
23
```

AngeloRubens:
Thank you — that log isn't reachable without being signed in to the Eclipse CI, so I was guessing at the cause from the outside and guessing wrong.

Fixed and pushed. The rule is `ImportOrder` with `option=bottom`, so static imports go *after* the non-static ones rather than before, plus `groups=com,io,jakarta,java,javax,org` with `separated=true`. That's why every file was wrong in the same way — I'd used the more common convention throughout.

The 9 in your paste were only the first module before the build stopped. I ran checkstyle 13.9.0, the version the build reports, against this repository's own `qa/.../checkstyle.xml`, which found 41 errors across the three modules: 38 ordering and 3 genuinely unused imports. All 41 are gone, verified with the same command. Tests and compilation are unaffected — 72 tests still pass, no warnings under `-Xlint:all`.

Worth noting for anyone hitting this later: the `[WARN]` and `[INFO]` findings (missing `@param` tags, magic numbers, repeated string literals) don't fail the build, only `[ERROR]` does.

dmatej:
Now the job failed because of the WebSocket timeout, that is a Jenkins issue I am working on in #26202 , so please have a patience with me (and the CBI infrastructure we are not much happy with).

AngeloRubens:
No rush at all, and thanks for saying so — I'd have kept chasing it otherwise.

One data point that may save you a check: I built the three modules in the reactor locally against 8.0.5-SNAPSHOT and it's clean — `mvn install` on `orb-http-protocol`, `orb-http-client` and `orb-http-server` is BUILD SUCCESS, with the project's checkstyle running and surefire executing all 107 tests. So nothing is blocked on our side while the WebSocket issue is sorted.

While waiting I put a small harness on my fork that deploys a session bean to a real GlassFish 8.0.4 and invokes it over IIOP, checking the behaviours this transport claims to reproduce — cyclic graphs, shared references keeping their identity, transient fields, large payloads, stateful conversations. It found a real bug, and one reading the specification had not shown me: the transport was propagating whatever the bean threw, unchanged, in every case. The baseline's own test exception was an unannotated `RuntimeException`, the container had correctly wrapped it in `EJBException`, and that mismatch is what surfaced it.

That would have been a bad way to fail — an application moved from IIOP to HTTP would have started catching exception types its `catch` blocks had never been able to see before, in code nobody had touched. Fixed in the latest commit: `@ApplicationException` and declared checked exceptions pass through as themselves, everything else is wrapped, honouring `inherited`. Both halves are now checked over a real connection.

Mentioning it because it seems worth knowing that the transport is being measured against the ORB rather than against my reading of what the ORB does.

AngeloRubens:

appserver/orb/orb-http-client/src/main/java/org/glassfish/orb/http/client/HttpInitialContextFactory.java
```diff
@@ -0,0 +1,153 @@
...
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+
+package org.glassfish.orb.http.client;
+
```
dmatej, line 18: Why 5 empty lines? Same in other files.

Issue #24084, opened 2022-08-19
Kill Descriptor.boundsChecking


dmatej:
## Problem Description

* cargo cult synchronization of it's setter and getter when affected usages are not synchronized. Any change affects all parallel deployments, but as it takes just microseconds, collisions are not happening. Maybe in the year 2030 ;)
* effectively relies on a call of SetMethodAction, nothing else changes the value.
* usages of isBoundsChecking are not just in setters.

## Expected Changes

* Just remove it and remove also all related code in dead branches.
* Will throw exceptions if someone would use invalid values in descriptors elements covered by related code. Until now it would tolerate invalid values and just use defaults for some setters, see the image.
![image](https://user-images.githubusercontent.com/302904/185681527-19e3ce65-b834-465f-a478-36221ed406eb.png)

## Notes (may be expanded later)

* analyze effects of removal in the getType , it switches primitive types to object types if the bounds checking is disabled.

PR #26204, opened 2026-09-11
Deployment speed improvements

OndroMih:
Focused on big EARs but should speed up deployments also for smaller apps.

Each commit, except the first one, brings additional optimization and speed up of deployment, and is independent of other improvements, although some improvements touch the same components.

I used AI to help with analysis of profiling data, suggest improvements and generate some of the solutions, but I thoroughly reviewed all of the AI-generated code and often adjusted it in the way I would write it myself.

An important observation is that detection of implicit CDI archives dropped from dozens of seconds to around 5 seconds for the big EAR under test. That means that manual configuration to remove specific JARs from scanningwhich e.g. Payara supports would bring only a minor improvement and there's little need to implement it in GlassFish for usual applications.

## Measurements of deployment improvements

I did measurements of each individual commit with a big EAR application, which depends on several RAR deployables and takes almost 2 minutes to deploy on my laptop without these optimizations. After all of these optimizations, the deployment time dropped **by about 50%, from about 2 minutes to about 1 minute**.

### Setup

**Host system:** Linux amd64, 6 cpus, JDK 21

Each iteration creates a fresh domain, starts it, deploys 3 RARs, adds the RARs to the app classpath with `server.connector-service.class-loading-policy=global` , restarts, then times a single `asadmin deploy`. Timings include the `asadmin` client JVM startup, which is constant for each run.

The measurements and analysis were done by scripts that were produced by AI reviewed by me. 

### Per-revision results

| # | commit | n | median ms | mean ms | stddev | min | max | exit codes |
|---|---|---|---|---|---|---|---|---|
| 0 | 40fcf2ac890 Log exception that is not rethrown | 5 | 109910 | 109326 | 4089 | 104776 | 114132 | 1 |
| 1 | a8ac0b40704 Perf improvements inspired by FISH-5938 from Payara | 5 | 85159 | 86868 | 4948 | 83513 | 95625 | 1 |
| 2 | 6d7f50848ee Speed up WeldSniffer | 5 | 76011 | 77187 | 2475 | 74801 | 80072 | 1 |
| 3 | 88d229a417d Cache FileArchive during deployment | 5 | 65824 | 66581 | 4124 | 63055 | 73639 | 1 |
| 4 | c05d7888569 Speed up Weld bootstrap by replacing List with a Set | 5 | 62424 | 62637 | 2102 | 59913 | 65559 | 1 |
| 5 | 0150b59acf5 Cache the ejb indexes of ComponentValidator per validation pass | 5 | 58202 | 59178 | 2635 | 57411 | 63850 | 1 |
| 6 | 7542feee64d Ignore missing Class-Path entries in ASURLClassLoader | 5 | 57458 | 58780 | 2982 | 57192 | 64104 | 1 |
| 7 | e6d3af870e9 Enhance ASURLClassLoader with indexed URL handling for RARs | 5 | 56514 | 56725 | 702 | 55797 | 57533 | 1 |

### Attribution per commit

Each row is the change introduced by that commit, measured against the revision before it. Negative means faster.

| commit | before ms | after ms | change ms | change % |
|---|---|---|---|---|
| a8ac0b40704 Perf improvements inspired by FISH-5938 from Payara | 109910 | 85159 | -24751 | -22.5% |
| 6d7f50848ee Speed up WeldSniffer | 85159 | 76011 | -9148 | -10.7% |
| 88d229a417d Cache FileArchive during deployment | 76011 | 65824 | -10187 | -13.4% |
| c05d7888569 Speed up Weld bootstrap by replacing List with a Set | 65824 | 62424 | -3400 | -5.2% |
| 0150b59acf5 Cache the ejb indexes of ComponentValidator per validation pass | 62424 | 58202 | -4222 | -6.8% |
| 7542feee64d Ignore missing Class-Path entries in ASURLClassLoader | 58202 | 57458 | -744 | -1.3% |
| e6d3af870e9 Enhance ASURLClassLoader with indexed URL handling for RARs | 57458 | 56514 | -944 | -1.6% |

Cumulative: **109910 ms -> 56514 ms**, -48.6%, speedup **1.94x**.

### Control measurement

`start-domain` is unaffected by every commit under test. If its median moves with the revision, the machine drifted and the deploy numbers are not comparable.

| # | commit | start-domain median ms | deviation from overall |
|---|---|---|---|
| 0 | 40fcf2ac890 Log exception that is not rethrown | 8126 | +5.0% |
| 1 | a8ac0b40704 Perf improvements inspired by FISH-5938 from Payara | 7718 | -0.3% |
| 2 | 6d7f50848ee Speed up WeldSniffer | 7975 | +3.0% |
| 3 | 88d229a417d Cache FileArchive during deployment | 8132 | +5.1% |
| 4 | c05d7888569 Speed up Weld bootstrap by replacing List with a Set | 7940 | +2.6% |
| 5 | 0150b59acf5 Cache the ejb indexes of ComponentValidator per validation pass | 7709 | -0.4% |
| 6 | 7542feee64d Ignore missing Class-Path entries in ASURLClassLoader | 7323 | -5.4% |
| 7 | e6d3af870e9 Enhance ASURLClassLoader with indexed URL handling for RARs | 7688 | -0.7% |

Worst control deviation: **5.4%**. Below 10%, so the comparison holds.

### Deploy outcome consistency

All runs produced the same normalized `asadmin deploy` output (signature `4067ec3f6263`), so every revision behaved identically.


nucleus/deployment/common/src/main/java/com/sun/enterprise/deploy/shared/FileArchive.java
```diff
@@ -676,20 +707,24 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IO
...
+     * Returns the name of the entry relative to the archive root, or null if the entry escapes it.
+     * <p>
+     * Only a symbolic link can point outside the archive, so plain entries are relativized against
+     * the unresolved root instead of being resolved first.
+     */
+    private String getEntryName(Path root, Path realRoot, Path path, BasicFileAttributes attrs) throws IOException {
+        if (!attrs.isSymbolicLink() && path.startsWith(root)) {
+            return root.relativize(path).toString().replace(File.separatorChar, '/');
+        }
+        Path resolved = path.toRealPath();
```
dmatej: I would prefer the original name `resolvedPath`. At first glance I thought that `resolved` is a boolean. The `realRoot` is ok.

nucleus/deployment/common/src/main/java/org/glassfish/deployment/common/DeploymentContextImpl.java
```diff
@@ -677,7 +687,7 @@ private File initTenantDir() {
 
         File tenantDirectory = new File(getRootTenantDirForApp(originalAppName), tenant);
         if (!tenantDirectory.exists() && !tenantDirectory.mkdirs()) {
-            deplLogger.log(FINEST, "Unable to create directory {0}", tenantDirectory.getAbsolutePath());
+            deplLogger.log(FINEST, () -> "Unable to create tenant directory " + tenantDirectory.getAbsolutePath() + " for tenant=" + tenant + " and app=" + originalAppName);
```
dmatej: I would probably remove the `.getAbsolutePath()` call. The `toString` should be anough.

nucleus/deployment/common/src/main/java/com/sun/enterprise/deploy/shared/FileArchive.java
```diff
@@ -638,32 +660,41 @@ List<String> getListOfFiles(File directory, final Logger logger) {
...
         }
-        List<String> files = new ArrayList<String>();
+        final List<String> files = new ArrayList<>();
+        final Path root = archive.toPath();
+        final Path realRoot;
+        try {
+            // Resolved once here instead of per entry; resolving a path is a relatively costly syscall.
+            realRoot = root.toRealPath();
+        } catch (IOException e) {
+            deplLogger.log(Level.WARNING, FILE_LIST_FAILURE, directory.getAbsolutePath());
```
dmatej: The error log should log the exception and probably use the `root` variable.

nucleus/deployment/common/src/main/java/com/sun/enterprise/deploy/shared/FileArchive.java
```diff
@@ -638,32 +660,41 @@ List<String> getListOfFiles(File directory, final Logger logger) {
...
+            return files;
+        }
+        final Path start = directory.toPath();
```
dmatej: Better name than `start` - rootDirectory? Hmm, in current context it could be confusing too. Maybe you can rather leave it as it was, see File.toPath body:
```Java
    public Path toPath() {
        Path result = filePath;
        if (result == null) {
            synchronized (this) {
                result = filePath;
                if (result == null) {
                    result = FileSystems.getDefault().getPath(path);
                    filePath = result;
                }
            }
        }
        return result;
    }
```

nucleus/deployment/common/src/main/java/com/sun/enterprise/deploy/shared/FileArchive.java
```diff
@@ -638,32 +660,41 @@ List<String> getListOfFiles(File directory, final Logger logger) {
         if (archive == null || directory == null || !directory.isDirectory()) {
             return Collections.emptyList();
         }
-        List<String> files = new ArrayList<String>();
+        final List<String> files = new ArrayList<>();
+        final Path root = archive.toPath();
```
dmatej: `archivePath` rather than root? Is it the path of the archive or the root inside the archive? That would be `archiveRoot`.

nucleus/deployment/common/src/main/java/com/sun/enterprise/deploy/shared/FileArchive.java
```diff
@@ -144,6 +156,8 @@ public void open(URI uri) throws IOException {
         }
         isOpenedOrCreated = true;
         staleFileManager = StaleFileManager.Util.getInstance(archive);
+        // Opening does not change anything on disk, so only this archive's own entries are stale.
+        entryNamesCache.clear();
```
dmatej: I am not sure about consequences. Java uses some caching, classloading is parallelized now, it is easy to leak something.

nucleus/deployment/common/src/main/java/com/sun/enterprise/deploy/shared/FileArchive.java
```diff
@@ -210,6 +225,8 @@ public boolean delete() {
             return result;
         } catch (IOException e) {
             return false;
+        } finally {
+            releaseCachedEntryNames();
```
dmatej: Is it correct to first delete directory and then release caches, or should be release the cache first? Why this method doesn't need to be synchronized? Why close is empty? Very attractive rabbit hole questions for me ...

dmatej:
The first commit is relevant to #24084

OndroMih:
The commit https://github.com/eclipse-ee4j/glassfish/commit/6d7f50848ee452d37c0d43196627e3fe6afb5dff "Speed up WeldSniffer" fails some tests in "ejb_group_3". Reverting that commit passes the tests .

I tried reverting only the removal of getAnnotationNames() method but it didn't help. Investigating further what exactly in that commit broke the tests, so that I can either fix it or revert only that part instead of the whole commit.

nucleus/deployment/common/src/main/java/com/sun/enterprise/deploy/shared/FileArchive.java
OndroMih: I don't see what could be dangerous about this: 

* The entryNamesCache is local to this FileArchive instance
* It's used only by this instance, not by the JVM or anything else
* It only caches names of files in directories inside the archive, not any IO-bound objects or descriptors

In this `open` method, the cache for this archive instance is cleared so that entries from a previously opened file don't leak into the newly opened archive.

FileArchive is a `@PerLookup` service. The instances are stored into the DeploymentContext until the app is undeplyed, but the caches are cleared right after deploy to free the memory, in `DeploymentContextImpl.postdeployClean`

nucleus/deployment/common/src/main/java/com/sun/enterprise/deploy/shared/FileArchive.java
OndroMih: Good questions. I added lock-free synchronization mechanism to avoid using caches during dirty operations, until all dirty operations complete. Then it doesn't matter whether cache is cleared before or after delete directory.

nucleus/deployment/common/src/main/java/com/sun/enterprise/deploy/shared/FileArchive.java
OndroMih: It's path of the archive but it's the same as the archive root - FileArchive is always a directory. I'll rename it to archivePath because "path" is used everywhere, and "root" is never used as a term in this file.

nucleus/deployment/common/src/main/java/com/sun/enterprise/deploy/shared/FileArchive.java
OndroMih: OK, reverted to `directory.toPath()` as this call is very cheap

nucleus/deployment/common/src/main/java/com/sun/enterprise/deploy/shared/FileArchive.java
OndroMih: I changed it to realPath to follow the same pattern as in other places

nucleus/deployment/common/src/main/java/org/glassfish/deployment/common/DeploymentContextImpl.java
OndroMih: OK

OndroMih:
I reverted one more thing and the failed tests passed locally.

appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/annotation/impl/ModuleScanner.java
```diff
@@ -166,10 +168,15 @@ protected void completeProcess(T descriptor, ReadableArchive archive) throws IOE
 
     protected void calculateResults(T descriptor) {
         try {
-            classParser.awaitTermination();
+            final Exception[] parserExceptions = classParser.awaitTermination();
+            if (parserExceptions.length > 0) {
+                final MultiException e = new MultiException(List.of(parserExceptions));
+                LOG.log(Level.SEVERE, ANNOTATION_SCANNING_EXCEPTION, e);
+                throw e;
```
dmatej, line 175: Would not be better to wrap it to DeploymentException too? Or is it done elsewhere?

OndroMih: I was thinking about it too, but then I realized that anything can throw any runtime exception during deployment and it should be handled uniformly somehere at the upper layer. There's no reason to wrap MultiException into another DeploymentException. I wrapped InterruptedException only because it's not a runtime exception. I could use RuntimeException but DeploymentException looked like a better choice.

I checked what is catching that code and it's caught in the ApplicationLifecyle.deploy method, which sets it to the report object and returns null. I think the report object is transferred as JSON to the client, with the exception as as. Maybe we should wrap exceptions into DeploymentException in `ApplicationLifecyle.deploy` method, before setting it to the report, but that's out of scope of this PR.

dmatej: Reports are a bit of a black magic, I was there once, and touching any method has unpredictable consequences. So ok, thanks for the explanation.

dmatej (approved):

PR #26213, opened 2026-09-14
Fix flaky timeout duration assertion in JtaTimeoutLoggingITest

### LINKED ISSUE #26212: JtaTimeoutLoggingITest is flaky: timeout duration pattern tolerates only 9 ms of timer jitter
`JtaTimeoutLoggingITest` verifies the logged transaction timeout duration with this pattern ([JtaTimeoutLoggingITest.java#L89-L90](https://github.com/eclipse-ee4j/glassfish/blob/main/appserver/tests/jdbc/src/test/java/org/glassfish/main/test/jdbc/jta/JtaTimeoutLoggingITest.java#L89-L90)):

```java
private static final Pattern P_TIMEOUT = Pattern.compile( //
    "\s+Transaction with id=[0-9]+ timed out after 200[0-9] ms." + LOG_END);
```

The transaction timeout used by the test application is 2 s (`SlowJpaPartitioner.TIMEOUT_IN_SECONDS`), so the pattern accepts 2000-2009 ms, i.e. 9 ms of tolerance.

The duration in the message is not the configured timeout. It is measured in `JavaEETransactionImpl.run()`, which is the `TimerTask` body, so it includes the delay between the scheduled expiration and the moment the timer thread actually runs ([JavaEETransactionImpl.java#L173-L177](https://github.com/eclipse-ee4j/glassfish/blob/main/appserver/transaction/jta/src/main/java/com/sun/enterprise/transaction/JavaEETransactionImpl.java#L173-L177)):

```java
public void run() {
    timedOut = true;
    final long time = System.currentTimeMillis() - startTime;
    LOG.warning(() -> String.format("Transaction with id=%s timed out after %s ms.", txId, time));
```

That delay is unbounded on a loaded machine. On the Ubuntu CI runner it was 11 ms, which failed the build:

```
[ERROR] Failures:
[ERROR]   JtaTimeoutLoggingITest.timeoutOnly:152
Expected: iterable containing [a string matching the pattern '\s+Transaction with id=[0-9]+ timed out after 200[0-9] ms.[\|\#\]]*', ...]
     but: item 0: was "  Transaction with id=1 timed out after 2011 ms.]]"
[ERROR] Tests run: 38, Failures: 1, Errors: 0, Skipped: 2
...
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-failsafe-plugin:3.5.6:verify (verify) on project jdbc-tests: There are test failures.
```

Occurrence: https://github.com/eclipse-ee4j/glassfish/actions/runs/34840145032/job/103962956790 (a PR touching only `nucleus/admin` REST session code; the 41 minute job failed on this single assertion, everything else in the reactor was green).

The assertion is meant to prove that the 2 s timeout fired rather than some default one, which does not require millisecond precision. Widening the accepted range to the second following the timeout keeps that guarantee and removes the dependency on timer scheduling latency.

renatsaf:
Fixes #26212

`JtaTimeoutLoggingITest` accepted only `200[0-9] ms` as the logged duration of the 2 s transaction timeout, i.e. 9 ms of tolerance. The value comes from `JavaEETransactionImpl.run()`, the `TimerTask` body, so it is the configured timeout plus the delay before the timer thread runs - unbounded on a loaded machine. The Ubuntu CI runner reported 2011 ms and failed the whole build in a PR that touched only `nucleus/admin` REST session code ([run 34840145032](https://github.com/eclipse-ee4j/glassfish/actions/runs/34840145032/job/103962956790)).

The pattern now accepts the second following the timeout, and takes the value from `TIMEOUT_IN_SECONDS` rather than repeating the literal:

```java
"\s+Transaction with id=[0-9]+ timed out after " + TIMEOUT_IN_SECONDS + "[0-9]{3} ms." + LOG_END
```

The assertion still proves that the configured 2 s timeout fired and not a different one: `2000`-`2999` matches, `3000` and `200` do not.

Not executed locally - the test requires Docker plus a full distribution build; verification is the CI run on this PR.


dmatej:
It seems there was something wrong happening with GitHub Actions three days ago - you made this tolerant more, but another test failed. 
I just checked there can't be any false positives and it seems ok.

dmatej (approved):

PR #26214, opened 2026-09-14
Propagate transactions over the HTTP transport

AngeloRubens:
Depends on #26201. This branch is stacked on it; the change to review is the last commit.

#26201 refuses an invocation that carries a transaction. This makes it work. The wire space `TxContext` reserved there is used unchanged.

**Two families**, following the WildFly HTTP client wire specification v1:
- `ut`: begin, commit and rollback, driven by the client through `HttpUserTransaction`.
- `xa`: prepare, commit, rollback, forget and recover, for a caller with its own coordinator. The endpoint is enlisted as `HttpXAResource`.

**Container transactions.** Code already running in a container transaction reaches a remote bean without touching this API. `AmbientTransaction` enlists the endpoint with the caller's transaction manager, and keeps the branch with the transaction rather than with the thread. A pooled thread cannot carry a finished branch into its next call, and two endpoints in one transaction each carry their own branch.

**Server.** `TransactionBridge` is the seam to the container. `TransactionDispatcher` serves both families. `EjbDispatcher` imports the branch around each invocation and releases it afterwards, and it clears the thread for an invocation that carries none.

**Tests.** 61 new, against a fake bridge, so no server is needed: user transactions, rollback-only, the XA two-phase sequence across two endpoints, read-only votes, a refused prepare rolling everything back, recovery, XA error codes, and the ambient cases above.

**Not here.** The GlassFish implementation of `TransactionBridge` is in the next PR.

*Development: see the note in #26201; the same applies here.*


AngeloRubens:

OndroMih:
We are currently very busy with preparing the 8.0.5 release and resolving some blockers. We couldn’t get to reviewing this and the other related PRs yet.

These contributions seem solid and are really welcome, we just need a lot more time to get to them and review them properly. Probably in scope of several weeks. Please be patient, we’ll grt to it as we are really interested in the remoting over HTTP feature.

AngeloRubens:
Thanks @OndroMih , talk soon then. We'll be in touch once we start looking at the PR.

PR #26215, opened 2026-09-14
Bind the HTTP transport to the GlassFish container

AngeloRubens:
Depends on #26201 and #26214. The change to review is the last commit.

**New module `orb-http-glassfish`** binds the container-independent dispatchers to GlassFish:
- `OrbHttpEndpoint`, `OrbHttpHandler`, `GrizzlyServerExchange`: the endpoint, mounted on the server's HTTP listeners
- `GlassFishContainerBridge`, `EjbNameIndex`, `GlassFishNamingBridge`: beans and names resolved through the container
- `GlassFishSecurityBridge`, `RealmAuthenticator`: the Basic credential checked against the realm. Absent means anonymous; rejected means 403.
- `OsgiCodecScanner`: codecs installed as bundles are found
- `GlassFishTransactionBridge`: imported transactions run through GlassFish's own transaction manager
- 'introduce a new module codec ORB, based on apache fory' for fast perfomance on marshall and unmarshall. 

**Outside the transport's modules:**

| Module | Change |
|---|---|
| `orb-connector` | `EjbContainerFacade` gains `createSession` / `removeSession`. IIOP never names a session key because it rides inside a reference; HTTP has nowhere to hide it. |
| `ejb-container` | `StatefulSessionContainer` implements them; `BaseContainer` refuses them |
| `featuresets/glassfish`, `featuresets/embedded-all` | the four transport modules are added; the optional Fory codec is not |

**Transactions** go through the `XATerminator` that the JCA inflow contract already provides. Two of its behaviours, measured against a running server, shape the code:
- A one-phase commit *accepts* a branch a bean marked for rollback. A branch with no other participant is therefore prepared before it is committed.
- A prepare issued from a thread carrying another transaction fails with `XAER_PROTO`. The terminator is only reached from a clean thread.

**Verification.** 16 unit tests here. End to end, against a real GlassFish 8.0.4 on every change, on both codecs: the same bean over IIOP and over HTTP; every transaction attribute from `REQUIRED` to `NEVER`; `UserTransaction` commit, rollback and rollback-only; a bean on one server calling a bean on a second server over HTTP, both when it commits and when a rollback decided on the second server reaches the client. Latest run: https://github.com/AngeloRubens/glassfish/actions/runs/34875291232. The harness (`ci/`, `.github/`) is not part of this PR.

*Development: see the note in #26201; the same applies here.*


AngeloRubens:

PR #26217, opened 2026-09-16
Jenkinsfile - disabled snapshots, tar failures escalate, -ntp, stash all we built

dmatej:
* `-ntp` - don't log downloads; note - maven dependency plugin hides downloads.
* Snapshots poisoned build when implementing backward incompatible changes.
* sh did not escalate tar failures, bash does
* In #26202 I tried to limit maven-repo.tar.gz file size, but then missing dependencies were downloaded form Maven Central Snapshots, even if incompatible.

* The removed common-util in scatteredarchive tests pointed to exactly this issue. However the dependency was not even required.


dmatej:
These changes were separated from #26204

OndroMih (approved):

PR #26225, opened 2026-09-21
Remove ant from modules

OndroMih:
It's over 2MB big JAR and doesn't seem to be imported and used by any other module


dmatej:
It is written in the comment, why it is required. So first check webservices-osgi, maybe I already removed the dependency, and maybe not. I think some tests failed somewhere without it, but not sure now.  

Manifest of webservices-osgi 4.0.7-SNAPSHOT:
```
                                       org.apache.coyote;version="[11.0,12)";resolution:=optional
                                        org.apache.jcp.xml.dsig.internal.dom;version="[4.0,5)"
                                        org.apache.tools.ant.taskdefs;resolution:=optional
                                        org.apache.tools.ant.types;resolution:=optional
                                        org.apache.tools.ant;resolution:=optional
```

I think some tests in Metro (TCK?) using GlassFish failed without it.

dmatej:
Ok, let's try it, I updated the TCK job in Metro: https://ci.eclipse.org/metro/job/JAXWS_jakarta_TCK_10_run_GF
However the TCK is not ready for 8+ without SecurityManager (and policy file).

dmatej:
We could take it also from the opposite side - remove ant from webservices-osgi dependencies, and then go deeper, if it should stay somewhere, it should be an isolated standalone module.

PR #26230, opened 2026-09-22
Don't declare build as Unstable, just append warnings to its description

dmatej:
Resolves problem with mapping of Jenkins statuses to GitHub statuses - GitHub takes Unstable as a Failure, while my intention was just to report a warning.

However for GitHub users it is not important that Jenkins had to repeat initialization of nodes, it is important just for Jenkins admins.


OndroMih (approved):

dmatej:
Funny, we pretty overloaded Jenkins this waited nearly two hours ... and not a single warning. These issues are really unpredictable.

